PR #31, opened 2015-05-27
Core/Player: Skip item SPELLTRIGGER_ON_USE at form change

robinsch:
This will fix auras casted from items with spell charges (e.g.: http://www.wowhead.com/spell=25211/heavy-golden-necklace-of-battle) disappearing on form change

### DIFF
--- a/src/game/Group.cpp
+++ b/src/game/Group.cpp
@@ -418,6 +418,7 @@ void Group::Disband(bool hideDestroy)
         }
 
         _homebindIfInstance(player);
+        SendObjectUpdateToMembers(player);
     }
     RollId.clear();
     m_memberSlots.clear();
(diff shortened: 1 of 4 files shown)


stfx:
Thanks. Since we are working on wotlk and backporting from there do you think you could make a PR against wotlk if this is needed there as well or is it only needed for TBC/Classic?

robinsch:
The health pct thing is only for TBC and vanilla.

stfx:
Alright, what about the SPELLTRIGGER_ON_USE?

robinsch:
Can be used for all branches.

evil-at-wow:
@stfx I can't reopen this PR - because the repository that submitted the PR has been deleted - but I'll have a good look at the TBC/Classic specific part now.

stfx:
That part was regarding the health percent which is a separate PR: #32

evil-at-wow:
@stfx I know, I just didn't notice it got its own PR.

PR #42, opened 2015-10-07
Fix for triggered spells costing mana/rage/energy

killerwife:
Check whether spell is triggered before cost computation

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -2697,7 +2697,9 @@ void Spell::prepare(SpellCastTargets const* targets, Aura* triggeredByAura)
     }
 
     // Fill cost data
-    m_powerCost = CalculatePowerCost(m_spellInfo, m_caster, this, m_CastItem);
+    if (!m_IsTriggeredSpell){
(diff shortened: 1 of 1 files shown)


cala:
I only spot that one now, sorry.
Are we speaking of spells triggered by another spell or triggered by script(or other bypassing mechanics like GM command or whatever)? Because in the first case, I wonder if the power cost should still applied (no example here, just wondering).

killerwife:
Well this is just backport from wotlk :D

Example is GM triggered, triggered using boolean in castSpell

robinsch:
I would rather put that into CalculatePowerCost - so we can use this in other places

killerwife:
Propably move it in all cores tho for consistency.

killerwife:
Apparently even wotlk is missing it. It has to be in place for spells like Devastate (sunder armor is triggered with no rage cost). I am gonna amend the commit in a moment.

killerwife:
Hmmm i deleted that fork a long time ago. I will make a new PR.

PR #44, opened 2016-02-08
Add Attacking Target Farthest Away

kartoffelsup:
Watchkeeper Gargolmar's Surge should be cast
on the target which is the farthest away.

This change adds this functionality to the
Creature class.

### DIFF
--- a/src/scriptdev2/scripts/outland/hellfire_citadel/hellfire_ramparts/boss_watchkeeper_gargolmar.cpp
+++ b/src/scriptdev2/scripts/outland/hellfire_citadel/hellfire_ramparts/boss_watchkeeper_gargolmar.cpp
@@ -16,8 +16,8 @@
 
 /* ScriptData
 SDName: Boss_Watchkeeper_Gargolmar
-SD%Complete: 80
-SDComment: Missing adds to heal him. Surge should be used on pTarget furthest away, not random.
+SD%Complete: 90
+SDComment: Missing adds to heal him. Surge spell implementation needs some improvements (No charge animation)
 SDCategory: Hellfire Citadel, Hellfire Ramparts
 EndScriptData */
 
@@ -117,14 +117,12 @@ struct boss_watchkeeper_gargolmarAI : public ScriptedAI
 
         if (m_uiSurgeTimer < uiDiff)
         {
-            if (Unit* pTarget = m_creature->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0))
-            {
+            if (Unit* pTarget = m_creature->SelectAttackingTarget(ATTACKING_TARGET_FARTHEST_AWAY, 0, SPELL_SURGE, SELECT_FLAG_PLAYER))
                 if (DoCastSpellIfCan(pTarget, SPELL_SURGE) == CAST_OK)
                 {
                     DoScriptText(SAY_SURGE, m_creature);
                     m_uiSurgeTimer = urand(12100, 21700);
                 }
-            }
         }
         else
             m_uiSurgeTimer -= uiDiff;
--- a/src/game/Creature.cpp
+++ b/src/game/Creature.cpp
@@ -2175,6 +2175,11 @@ Unit* Creature::SelectAttackingTarget(AttackingTarget target, uint32 position, S
     if (position >= threatlist.size() || !threatlist.size())
         return nullptr;
 
+    if (threatlist.size() == 1)
+        if (Unit* pTarget = GetMap()->GetUnit((*itr)->getUnitGuid()))
+            if (_canSelect(pTarget, pSpellInfo, selectFlags))
+                return pTarget;
+
     switch (target)
     {
         case ATTACKING_TARGET_RANDOM:
@@ -2184,7 +2189,7 @@ Unit* Creature::SelectAttackingTarget(AttackingTarget target, uint32 position, S
             advance(itr, position);
             for (; itr != threatlist.end(); ++itr)
                 if (Unit* pTarget = GetMap()->GetUnit((*itr)->getUnitGuid()))
-                    if (!selectFlags || MeetsSelectAttackingRequirement(pTarget, pSpellInfo, selectFlags))
+                    if (_canSelect(pTarget, pSpellInfo, selectFlags))
                         suitableUnits.push_back(pTarget);
 
             if (!suitableUnits.empty())
@@ -2197,7 +2202,7 @@ Unit* Creature::SelectAttackingTarget(AttackingTarget target, uint32 position, S
             advance(itr, position);
             for (; itr != threatlist.end(); ++itr)
                 if (Unit* pTarget = GetMap()->GetUnit((*itr)->getUnitGuid()))
-                    if (!selectFlags || MeetsSelectAttackingRequirement(pTarget, pSpellInfo, selectFlags))
+                    if (_canSelect(pTarget, pSpellInfo, selectFlags))
                         return pTarget;
 
             break;
@@ -2207,14 +2212,37 @@ Unit* Creature::SelectAttackingTarget(AttackingTarget target, uint32 position, S
             advance(ritr, position);
             for (; ritr != threatlist.rend(); ++ritr)
                 if (Unit* pTarget = GetMap()->GetUnit((*itr)->getUnitGuid()))
-                    if (!selectFlags || MeetsSelectAttackingRequirement(pTarget, pSpellInfo, selectFlags))
+                    if (_canSelect(pTarget, pSpellInfo, selectFlags))
                         return pTarget;
 
             break;
         }
+        case ATTACKING_TARGET_FARTHEST_AWAY:
+        {
+            float highestDistance = -1;
+            Unit* farthestAway = nullptr;
+            for (; itr != threatlist.end(); ++itr)
+                if (Unit* pTarget = GetMap()->GetUnit((*itr)->getUnitGuid()))
+                {
+                    float dist = Creature::GetCombatDistance(pTarget, false);
+                    const bool isFartherAwayThanCurrent = dist > highestDistance;
+                    if (isFartherAwayThanCurrent && _canSelect(pTarget, pSpellInfo, selectFlags))
+                    {
+                        highestDistance = dist;
+                        farthestAway = pTarget;
+                    }
+                }
+
+            return farthestAway;
+        }
     }
 
     return nullptr;
+}
+
+bool Creature::_canSelect(Unit* pTarget, SpellEntry const* pSpellInfo /*= nullptr*/, uint32 selectFlags/*= 0*/) const
+{
+    return !selectFlags || MeetsSelectAttackingRequirement(pTarget, pSpellInfo, selectFlags);
 }
 
 void Creature::_AddCreatureSpellCooldown(uint32 spell_id, time_t end_time)
--- a/src/game/Creature.h
+++ b/src/game/Creature.h
@@ -313,6 +313,7 @@ enum AttackingTarget
     ATTACKING_TARGET_RANDOM = 0,                            // Just selects a random target
     ATTACKING_TARGET_TOPAGGRO,                              // Selects targes from top aggro to bottom
     ATTACKING_TARGET_BOTTOMAGGRO,                           // Selects targets from bottom aggro to top
+    ATTACKING_TARGET_FARTHEST_AWAY,                         // Selects the farthest away target
 };
 
 enum SelectFlags
@@ -791,6 +792,7 @@ class MANGOS_DLL_SPEC Creature : public Unit
     private:
         GridReference<Creature> m_gridRef;
         CreatureInfo const* m_creatureInfo;                 // in heroic mode can different from sObjectMgr::GetCreatureTemplate(GetEntry())
+        bool _canSelect(Unit* pTarget, SpellEntry const* pSpellInfo /*= nullptr*/, uint32 selectFlags/*= 0*/) const;
 };
 
 class ForcedDespawnDelayEvent : public BasicEvent
--- a/src/scriptdev2/scripts/outland/tempest_keep/arcatraz/boss_dalliah.cpp
+++ b/src/scriptdev2/scripts/outland/tempest_keep/arcatraz/boss_dalliah.cpp
@@ -165,18 +165,14 @@ struct boss_dalliahAI : public ScriptedAI
         }
 
         if (!m_bIsRegularMode)
-        {
             if (m_uiShadowWaveTimer < uiDiff)
             {
-                if (Unit* pTarget = m_creature->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0))
-                {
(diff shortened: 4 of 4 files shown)


Phatcat:
You can remove brackets from one line statements, making all brackets except the ones used in the _canSelect redundant.

Is Watchkeeper Gargolmar the only one with this kind of functionality? because then adding it as generic functionality may be a bit over the top in my opinion, especially if he is already a scripted mob. But if just 2 non-related mobs have this kind of functionality then I see no problem with this, beside the overuse of brackets.

kartoffelsup:
According to WowWiki, [Dalliah the Doomsayer](http://wowwiki.wikia.com/wiki/Dalliah_the_Doomsayer)'s Shadow Wave ability might also benefit from this functionality.

If you say that it is okay to keep it where it is, I will remove the unnecessary brackets. (Sorry, I'm way too used to do it that way from work).

Phatcat:
It would seem indeed that Dalliah would benefit from this change as well, I am behind the generic change then :+1:

Phatcat:
We would probably want someone like @killerwife or @xfurry to look at this, though.

you also of course need to squash it all into 1 commit but there's no reason to do that before you're done modifying it anyway; and one of the 2 people I mentioned would most likely have you modify *something* within it, if they are to look at this, that is.

Phatcat:
And unless these kind of cases can be found for vanilla (which I doubt) then this will also only need an upport to wrath and not a backport to vanilla.

Phatcat:
If you squash/fixup this a bit I won't mind testing it

kartoffelsup:
Like this?

Phatcat:
*Exactly* like that.

Phatcat:
Well, actually, you may want to seperate the script changes from the core changes, and do another commit where you add the functionality to Dalliah as well? If you don't mind of course.

kartoffelsup:
Sure will do! Might just take a bit though.

Phatcat:
It's what you're doing in the scripts that fails in VS.

xfurry:
I don't think we should do it like this.
There are very few cases where we actually really need to select the farthest away target.
I think it would be best to do this in script, similar to the TK boss - https://github.com/cmangos/mangos-tbc/blob/master/src/scriptdev2/scripts/outland/tempest_keep/the_eye/boss_void_reaver.cpp#L129

xfurry:
> Watchkeeper Gargolmar's Surge should be cast
> on the target which is the farthest away.

Also, I'm not 100% sure if this should be really on the farthest away target.
As per my knowledge the charge / surge spells should be cast on targets which are out of melee range, between certain distances from the caster, usually defined in the spell info.

kartoffelsup:
> Also, I'm not 100% sure if this should be really on the farthest away target.
http://www.wowhead.com/npc=17306/watchkeeper-gargolmar#comments:id=58561
http://www.wowhead.com/npc=17306/watchkeeper-gargolmar#comments:id=23221
http://www.wowhead.com/npc=17306/watchkeeper-gargolmar#comments:id=27150

src/scriptdev2/scripts/outland/hellfire_citadel/hellfire_ramparts/boss_watchkeeper_gargolmar.cpp
```diff
@@ -116,16 +116,13 @@ struct boss_watchkeeper_gargolmarAI : public ScriptedAI
...
-            {
-                if (DoCastSpellIfCan(pTarget, SPELL_SURGE) == CAST_OK)
-                {
-                    DoScriptText(SAY_SURGE, m_creature);
-                    m_uiSurgeTimer = urand(12100, 21700);
-                }
-            }
-        }
+            if (const SpellEntry* pSpell = sSpellStore.LookupEntry(SPELL_SURGE))
+                if (Unit* pTarget = m_creature->SelectAttackingTarget(ATTACKING_TARGET_FARTHEST_AWAY, 0, pSpell, SELECT_FLAG_PLAYER))
```
xfurry: instead of pSpell you can use the SPELL_SURGE directly. The core will get the SpellEntry for you.
Also, don't remove the {} from the timers, as it makes the code look bad.

kartoffelsup: I Wasn't aware that this was possible! Thanks for the hint. :)

xfurry:
I don't want to argue with the comments on wowhead, but I think for our case it would be more useful to have a function that will say SelectTargetWithinSpellRange() that will select a target within the min and max range of the spell.
Also, we need to consider that the logic behind this, isn't really to charge the target that is the farthest away from the caster, but rather to charge the highest aggro target that is out of melee range (usually a caster, and usually the farthest away).

Phatcat:
Actually, it is not only the comments on wowhead, the wiki says directly [on dalliah;](http://wowwiki.wikia.com/wiki/Dalliah_the_Doomsayer)
>Shadow Wave: A 2.5k Shadowbolt which also gives a 50% increased vulnerability to shadow damage as debuff. This will always target the furthest player from her. 

"This will **always** target the furthest player from her" - no mention of range or anything.
It very much seems like blizzard added this kind of functionality in TBC and used with a few bosses.

src/scriptdev2/scripts/outland/hellfire_citadel/hellfire_ramparts/boss_watchkeeper_gargolmar.cpp
```diff
@@ -16,8 +16,8 @@
 
 /* ScriptData
 SDName: Boss_Watchkeeper_Gargolmar
-SD%Complete: 80
-SDComment: Missing adds to heal him. Surge should be used on pTarget furthest away, not random.
+SD%Complete: 90
+SDComment: Missing adds to heal him. Surge spell implementation needs some improvements (No charge animation)
```
Appled: This charge animation is bigger problem, see here https://github.com/cmangos/issues/issues/697

kartoffelsup:
So how should I proceed? Implement it within the scripts and remove the reusable method within Creature.cpp? Leave it as is?

cyberium:
Well if we give that to scripter we will have same code in more than one place and probably implemented in different way if its done by different devs. So i prefer to let the core handle this case like you did.
However like @xfurry said i dont see any way to specify the maximum distance in your code. There must be one maximum, think about charging the farthest player in aggro list but realy far from the combat... This will broke the event.

xfurry:
@cyberium I have an idea on how to fix this better.
Let me come back on this in a few days. I'm very busy with other things at the moment, but I have this on my ToDo list.

kartoffelsup:
So I will just move on? :p

xfurry:
Quick one:
Does this piece of code do the trick?

```diff
diff --git a/src/scriptdev2/scripts/outland/hellfire_citadel/hellfire_ramparts/boss_watchkeeper_gargolmar.cpp b/src/scriptdev2/scripts/outland/hellfire_citadel/hellfire_ramparts/boss_watchkeeper_gargolmar.cpp
index 41533c4..eed898e 100644
--- a/src/scriptdev2/scripts/outland/hellfire_citadel/hellfire_ramparts/boss_watchkeeper_gargolmar.cpp
+++ b/src/scriptdev2/scripts/outland/hellfire_citadel/hellfire_ramparts/boss_watchkeeper_gargolmar.cpp
@@ -117,7 +117,7 @@ struct boss_watchkeeper_gargolmarAI : public ScriptedAI
 
         if (m_uiSurgeTimer < uiDiff)
         {
-            if (Unit* pTarget = m_creature->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0))
+            if (Unit* pTarget = m_creature->SelectAttackingTarget(ATTACKING_TARGET_TOPAGGRO, 1, SPELL_SURGE, SELECT_FLAG_PLAYER | SELECT_FLAG_NOT_IN_MELEE_RANGE))
             {
                 if (DoCastSpellIfCan(pTarget, SPELL_SURGE) == CAST_OK)
                 {
diff --git a/src/scriptdev2/scripts/outland/tempest_keep/arcatraz/boss_dalliah.cpp b/src/scriptdev2/scripts/outland/tempest_keep/arcatraz/boss_dalliah.cpp
index 16910ad..8537fa0 100644
--- a/src/scriptdev2/scripts/outland/tempest_keep/arcatraz/boss_dalliah.cpp
+++ b/src/scriptdev2/scripts/outland/tempest_keep/arcatraz/boss_dalliah.cpp
@@ -168,7 +168,7 @@ struct boss_dalliahAI : public ScriptedAI
         {
             if (m_uiShadowWaveTimer < uiDiff)
             {
-                if (Unit* pTarget = m_creature->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0))
+                if (Unit* pTarget = m_creature->SelectAttackingTarget(ATTACKING_TARGET_TOPAGGRO, 1, SPELL_SHADOW_WAVE, SELECT_FLAG_PLAYER | SELECT_FLAG_NOT_IN_MELEE_RANGE))
                 {
                     if (DoCastSpellIfCan(pTarget, SPELL_SHADOW_WAVE) == CAST_OK)
                         m_uiShadowWaveTimer = urand(13000, 17000);

```

robinsch:
When nobody is in melee range?

xfurry:
> When nobody is in melee range?

This is a melee boss, so there will be someone in melee range.
The requirement of @kartoffelsup was to have a way to select the farthest away target from the boss.

Technically we could have that, but I think it's more correct to select the highest aggro player outside of melee range (usually caster).
So, that's why I asked if my proposal will do the trick.

robinsch:
I edited my comment before you posted yours.

xfurry:
That's also a valid point.
In this case, I think that the select argument makes sense.
Let me review this one more time, and as soon I have the chance I will push the changed proposed by @kartoffelsup

kartoffelsup:
Btw. I am fairly certain that I tested being out of range of the boss and did not get charged by Gargolmar and Shadow Bolted by Dalliah. Perhaps the Spell Range is taken into account at `MeetsSelectAttackingRequirement(pTarget, pSpellInfo, selectFlags);`. Or did you mean that we would have a problem for spells which have no range at all?

cyberium:
@xfurry i clearly remember some strategy that would require someone to be the most far from the boss. That had nothing to do with highest threat or second highest threat. so i think this should be implemented.
Well all of this make me think that we need a simple way for script to call target with
- highest aggro,
- lowest aggro.
- more near
- more far
- moving
- random
- specific class
- any other that i forget.
Not sure how it is working currently but it should be easy for a script to do such kind of target selection.

Phatcat:
>think about charging the farthest player in aggro list but realy far from the combat... This will broke the event.

Could add an additional check in the target selection process to see if the selected target is in line of sight/in reach of;

`target->isInAccessablePlaceFor(caster)`
`caster->IsWithinLOSInMap(target)`

Even though I don't like it, a check could even be done directly for range;

`caster->IsWithinDistInMap(target, spellMaxRange)`

> I have an idea on how to fix this better.

Not by choosing the target with the highest aggro, how is that better? that is neither correct nor logical in any way considering the desidered outcome, furthermore, I already told you this, but you simply chose to ignore it.

> but I think it's more correct to select the highest aggro player outside of melee range

I'll glady repeat;

> Actually, it is not only the comments on wowhead, the wiki says directly [on dalliah;](http://wowwiki.wikia.com/wiki/Dalliah_the_Doomsayer)
> >Shadow Wave: A 2.5k Shadowbolt which also gives a 50% increased vulnerability to shadow damage as debuff. This will always target the furthest player from her. 

> "This will **always** target the furthest player from her" - no mention of spell range, aggro or anything.

Phatcat:
@xfurry 
You are very aware that you are disregarding blizzlike behaviour, right? you are disregarding me backing a generic solution, you are disregarding kartoffels work and research and you are even disregarding old wowhead comments and cached wiki as well, all in order to try to achieve some custom behaviour which you personally like better which you've failed to provide *any* backing for what-so-ever. It shouldn't be nessescary to have cyberium tell you, and you still didn't get it, still proposing your, sorry for my language, shoddy excuse of a solution, trampling on kartoffels work as well.

I'm sorry, but that kind of behaviour just rubs me the wrong way. You're supposed to be a bastion of quality scripting, the watchkeeper of SD2. You're supposed to ensure quality, not turn it away at the door and propose lesser solutions in turn.

cyberium:
No need to be rude... @xfurry know more how little change in core may impact the large amount of script than me or you @Phatcat, so well in term of stuborn, all devs are at least a bit. And in this matter you already show more skill than @xfurry in 10 years :)
What we need now is to see your energy on our code more than on the comment like that and achieve at least 10% of what @xfurry done here.

That said now i hope everyone go in their own place. We are here to talk about the code not about the people and their way of managing their stuff.

i will delete all comment like that from now that help on nothing exept flaming.

Phatcat:
@cyberium
> However like @xfurry said i dont see any way to specify the maximum distance in your code. There must be one maximum, think about charging the farthest player in aggro list but realy far from the combat...

Look here: https://github.com/cmangos/mangos-tbc/pull/44/commits/0e96d2579c49e184d596b3f1b2ce7618f704edd0#diff-07d7427618f296efb8683a3b3cc8a31cR2229

When selecting its target it actually does check if it can select them for that spell.

I see no reason to alter this any further at this moment, just as I see no reason not to merge this.

kartoffelsup:
```c++
bool Creature::MeetsSelectAttackingRequirement(Unit* pTarget, SpellEntry const* pSpellInfo, uint32 selectFlags) const
{
    if (selectFlags & SELECT_FLAG_PLAYER && pTarget->GetTypeId() != TYPEID_PLAYER)
        return false;

    if (selectFlags & SELECT_FLAG_POWER_MANA && pTarget->GetPowerType() != POWER_MANA)
        return false;
    else if (selectFlags & SELECT_FLAG_POWER_RAGE && pTarget->GetPowerType() != POWER_RAGE)
        return false;
    else if (selectFlags & SELECT_FLAG_POWER_ENERGY && pTarget->GetPowerType() != POWER_ENERGY)
        return false;

    if (selectFlags & SELECT_FLAG_IN_MELEE_RANGE && !CanReachWithMeleeAttack(pTarget))
        return false;
    if (selectFlags & SELECT_FLAG_NOT_IN_MELEE_RANGE && CanReachWithMeleeAttack(pTarget))
        return false;

    if (selectFlags & SELECT_FLAG_IN_LOS && !IsWithinLOSInMap(pTarget))
        return false;

    if (pSpellInfo)
    {
        switch (pSpellInfo->rangeIndex)
        {
            case SPELL_RANGE_IDX_SELF_ONLY: return false;
            case SPELL_RANGE_IDX_ANYWHERE:  return true;
            case SPELL_RANGE_IDX_COMBAT:    return CanReachWithMeleeAttack(pTarget);
        }

        SpellRangeEntry const* srange = sSpellRangeStore.LookupEntry(pSpellInfo->rangeIndex);
        float max_range = GetSpellMaxRange(srange);
        float min_range = GetSpellMinRange(srange);
        float dist = GetCombatDistance(pTarget, false);

        return dist < max_range && dist >= min_range;
    }

    return true;
}

```

Should have the range checks covered.

PR #48, opened 2016-04-22
backport wotlk 0390d28 Clean up some redundant GetSpellDuration calls

hlarsen:
there were a few changes for items/quests that weren't in tbc (and consequently not in the mangoc-tbc source), so i left those out.


Phatcat:
Yeah, if this is simply a backport why havn't we merged this?
Unless someone tells me otherwise I'm gonna merge and backport to vanilla.
why? Because these are generic changes; and we want our cores in sync, and since wrath is the most developed core, we're using that as the guideline, no?

robinsch:
:+1:

Phatcat:
Merged

PR #50, opened 2016-04-22
backport nost-tbc a8a5d84 allow mounting in stealth

hlarsen:


cala:
Thank you for this PR.
Mounting in stealth? Are you sure?
http://eu.battle.net/wow/en/forum/topic/6493629662
http://us.battle.net/wow/en/forum/topic/2456909888
Also http://wowwiki.wikia.com/wiki/Useful_macros_for_rogues seems to indicates that rogues cannot cast stealth while mounted. I assume the reverse being also not possible.

Schaka:
It's not mounting in stealth. It allows the player to mount WHILE stealthed, meaning it will break stealth instead of just giving an error message that you can't mount (it's the wrong message anyway).

The macros are outdated, from before a using a spell would automatically dismount you (stealth).

killerwife:
@cyberium merge pls :)

killerwife:
wotlk and classic too

cyberium:
Thank, merged

PR #52, opened 2016-04-23
fix lightwell gameobject not appearing after spell is cast

hlarsen:
fixed the effect that summoned the lightwell game object not triggering and adding the go type to the world


killerwife:
@cyberium all versions

cyberium:
Thank, merged

PR #55, opened 2016-04-25
Some bugs related to quest 10353 (Arconus the Insatiable)

Ozerev:
When you accept quest [10353 ](http://www.wowhead.com/quest=10353/arconus-the-insatiable)from npc [20552 ](http://www.wowhead.com/npc=20552/agent-ya-six) a [guardian ](http://www.wowhead.com/npc=20603/reflection-of-ya-six)is summoned and you are given a 10 minute debuff to prevent the guardian being summoned again within 10 minutes. This patch fixes some bugs related to this quest:

- The debuff is no longer a positive effect so players can no longer cancel it by rightclicking the icon.
- The guardian is no longer summoned if a player accepts the quest while the debuff is active on the player. 

Note: I do not know if the way I prevented the summon when the debuff is active is the best one, but I did not find another way to do it. Although I'm a bit new to this so it's very possible that I missed something.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -6295,6 +6295,9 @@ bool Spell::CheckTarget(Unit* target, SpellEffectIndex eff)
 
     switch (m_spellInfo->Id)
     {
+        case 34877:                                         // Custodian of Time summon (quest 10277)
+        case 35381:                                         // Reflection of Ya-six summon (quest 10353)
+            return !target->HasAura(m_spellInfo->Id);
         case 37433:                                         // Spout (The Lurker Below), only players affected if its not in water
             if (target->GetTypeId() != TYPEID_PLAYER || target->IsInWater())
                 return false;
--- a/src/game/SpellMgr.cpp
+++ b/src/game/SpellMgr.cpp
@@ -720,6 +720,7 @@ bool IsPositiveEffect(SpellEntry const* spellproto, SpellEffectIndex effIndex)
                     {
                         case 13139:                         // net-o-matic special effect
                         case 23445:                         // evil twin
+                        case 35381:                         // Reflection of Ya-six
                         case 35679:                         // Protectorate Demolitionist
                         case 37695:                         // Stanky
                         case 38637:                         // Nether Exhaustion (red)


killerwife:
You might wanna squash it into one commit

xfurry:
Well, the code already handles ```summonResult = DoSummonGuardian(summonPositions, summon_prop, eff_idx, level);```
in the ```else``` part.
Why do we need another special case then?
Creature 20603 is type 10, so it shouldn't have any problem.

Ozerev:
I didn't find another way to prevent a guardian being summoned if the player has a specific aura, but it's very possible that I missed it. But right now on the TBC core you can just abandon and retake the quest and summon as many copies of the guardian as you want, which is what I was trying to fix. Same thing happens on quest [10277](http://db.hellground.net/?quest=10277).

Maybe this isn't a problem on the wotlk version, I noticed that the DoSummonGuardian function was a little different.

xfurry:
Here is a better solution.
You add this code, and before casting the aura that summons the guardian, you remove it first.

Also, something is very weird on TBC...
On 3.x you can't have more than one guardian. If you try to add it twice, you will lose it.

```diff
diff --git a/src/game/SpellAuras.cpp b/src/game/SpellAuras.cpp
index d7bca26..2dc45aa 100644
--- a/src/game/SpellAuras.cpp
+++ b/src/game/SpellAuras.cpp
@@ -2684,6 +2684,13 @@ void Aura::HandleAuraDummy(bool apply, bool Real)
                     pCaster->getHostileRefManager().ResetThreatRedirection();
                 return;
             }
+            case 34877:                                     // Custodian of Time
+            case 35381:                                     // Reflection of Ya-six
+            {
+                if (Unit* pCaster = GetCaster())
+                    pCaster->RemoveGuardians();
+                return;
+            }
             case 36730:                                     // Flame Strike
             {
                 target->CastSpell(target, 36731, true, nullptr, this);
```

Ozerev:
I think the current commit is a reasonable solution. The debuff will not refresh and the NPC will not be summoned if you accept any of the quests while already having the debuff, which is the implied intention of the debuff description ("Cannot resummon a Custodian of Time for the duration of this debuff").

By the way, in the wotlk core you can also abandon and retake these quests to [summon as many NPCs as you like](http://i.imgur.com/SJ9v4rc.png). Custodian of Time is not a guardian (it is UNITNAME_SUMMON_TITLE_NONE) so there are no checks for duplicates of it. Reflection of Ya-six is a guardian but it fails the check `if (!m_IsTriggeredSpell && m_caster->GetTypeId() == TYPEID_PLAYER)`, probably because the spell is triggered (it comes from quest_template.SrcSpell in the db).

cyberium:
Should be fixed in last rev.
Sorry no proposal here was enough correct to merge it.

PR #56, opened 2016-05-27
#944 - Healing items for specific spells give no +healing effect

Leadballoon2000:
The item's bonus was being applied to each tick. Note that coefficient isn't correct for Rejuvantion - should be 0.2

https://github.com/cmangos/issues/issues/944

### DIFF
--- a/src/game/Unit.cpp
+++ b/src/game/Unit.cpp
@@ -6213,6 +6213,7 @@ uint32 Unit::SpellHealingBonusDone(Unit* pVictim, SpellEntry const* spellProto, 
     // Done total percent damage auras
     float  DoneTotalMod = 1.0f;
     int32  DoneTotal = 0;
+	int32  DoneAdvertisedBenefit = 0;
 
     // Healing done percent
     AuraList const& mHealingDonePct = GetAurasByType(SPELL_AURA_MOD_HEALING_DONE_PERCENT);
@@ -6230,9 +6231,9 @@ uint32 Unit::SpellHealingBonusDone(Unit* pVictim, SpellEntry const* spellProto, 
         switch ((*i)->GetModifier()->m_miscvalue)
         {
             case 4415: // Increased Rejuvenation Healing
-            case 4953:
+            case 4953: // Harold's Rejuvenating Broach
             case 3736: // Hateful Totem of the Third Wind / Increased Lesser Healing Wave / LK Arena (4/5/6) Totem of the Third Wind / Savage Totem of the Third Wind
-                DoneTotal += (*i)->GetModifier()->m_amount;
+				DoneAdvertisedBenefit += (*i)->GetModifier()->m_amount;
                 break;
             default:
                 break;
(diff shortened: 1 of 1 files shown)


killerwife:
Still need to fix those tabs ^^

src/game/Unit.cpp
```diff
@@ -6229,18 +6230,17 @@ uint32 Unit::SpellHealingBonusDone(Unit* pVictim, SpellEntry const* spellProto,
             continue;
         switch ((*i)->GetModifier()->m_miscvalue)
         {
-            case 4415: // Increased Rejuvenation Healing
-            case 4953:
-            case 3736: // Hateful Totem of the Third Wind / Increased Lesser Healing Wave / LK Arena (4/5/6) Totem of the Third Wind / Savage Totem of the Third Wind
```
cala: Why remove that spell?

killerwife: He said its wotlk only.

robinsch: It's not. Check for SPELL_AURA_OVERRIDE_CLASS_SCRIPTS with miscvalue 3736

killerwife: Turns out you are right. Will tell him to change it back.

killerwife:
@Leadballoon2000 remove the removal of the alleged wotlk thing, since its present in TBC too

Leadballoon2000:
Sorry, I was wrong to say that it's a WotLK item. It went in during the last TBC season.
The item doesn't   have any +healing bonus so it doesn't need to have an entry in this case statement. I've tested that the +resilience effect still occurs with this change applied.

I meant to add a commit comment but I'm having trouble getting Git to do what I want so far.

killerwife:
No worries, thats the whole point of reviewing stuff.

Leadballoon2000:
I've added the case statement back since it's needed for item id 25645 Totem of the Plains
I don't see how the comment regarding the PvP relic relates to the statement.
There will need to be a database entry to spell_affect to get the Totem of the Plains to work, I'll raise an issue if there isn't one already.

Leadballoon2000:
I've added a number of other items to this PR but haven't been able to handle them the same way since the items add a dummy aura which doesn't have a miscvalue set.

Leadballoon2000:
I'm having trouble with Libram of Souls Redeemed itemid 28592
Blessing of Healing has two effect with the +healing values for Holy Light and Flash of Light. The aura for the libram only has one effect for Holy Light.
It doesn't seem very satisfactory to hard code the bonues based on the spell cast.

robinsch:
https://github.com/robinsch/TBCPvP/commit/fb0427f31f2e25f24ef635f2d979043f8dc29ee4

Leadballoon2000:
OK, I'll back out the Libram of Souls Redeemed change :)

Shouldn't the effect of the Libram and of Blessing of Light add to TakenAdvertisedBenefit otherwise downranked spells would get the full bonus?

killerwife:
@Leadballoon2000 feel free to test out @robinsch patch, he isnt contributing to cmangos atm anyway, but I dont think he will mind a patch of his making it here.

killerwife:
@Leadballoon2000 one more thing, do it properly so that robinshes authorship is kept for the second commit. Either do cherry-pick or commit --author

Leadballoon2000:
OK, I'm (pretty sure) I've updated the author.

src/game/Unit.cpp
```diff
@@ -1197,7 +1197,7 @@ void Unit::CastCustomSpell(Unit* Victim, SpellEntry const* spellInfo, int32 cons
 
     if (bp1)
         spell->m_currentBasePoints[EFFECT_INDEX_1] = *bp1;
-
+
```
Phatcat: this tab is a bit needless, wouldn't you say? =)

cyberium:
@killerwife, lot of hardcoded stuff here. Is it still needed?

killerwife:
@cyberium it is needed, but not in this form. Closing and making a github issue is the way to go propably.

PR #57, opened 2016-05-30
Fix los chase

killerwife:
@cyberium if you have any questions ask @Laizerox 
Else, please push it :D

### DIFF
--- a/src/game/CreatureEventAI.cpp
+++ b/src/game/CreatureEventAI.cpp
@@ -1370,11 +1370,18 @@ void CreatureEventAI::UpdateAI(const uint32 diff)
         // Update creature dynamic movement position before doing anything else
         if (m_DynamicMovement)
         {
-            if (!m_creature->hasUnitState(UNIT_STAT_CAN_NOT_REACT) && !m_creature->IsNonMeleeSpellCasted(false))
+            if (!m_creature->hasUnitState(UNIT_STAT_CAN_NOT_REACT))
             {
-                if (m_LastSpellMaxRange && m_creature->IsInRange(m_creature->getVictim(), 0, (m_LastSpellMaxRange / 1.5f)))
-                    SetCombatMovement(false, true);
-                else
+                bool uiInLineOfSight = m_creature->IsWithinLOSInMap(m_creature->getVictim());
+                
+                if (uiInLineOfSight && !m_creature->IsNonMeleeSpellCasted(false))
+                {
+                    if (m_LastSpellMaxRange && m_creature->IsInRange(m_creature->getVictim(), 0, (m_LastSpellMaxRange / 1.5f)))
+                        SetCombatMovement(false, true);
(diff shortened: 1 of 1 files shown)


src/game/CreatureEventAI.cpp
```diff
@@ -1366,11 +1366,18 @@ void CreatureEventAI::UpdateAI(const uint32 diff)
         // Update creature dynamic movement position before doing anything else
         if (m_DynamicMovement)
         {
-            if (!m_creature->hasUnitState(UNIT_STAT_CAN_NOT_REACT) && !m_creature->IsNonMeleeSpellCasted(false))
+            if (!m_creature->hasUnitState(UNIT_STAT_CAN_NOT_REACT))
             {
-                if (m_LastSpellMaxRange && m_creature->IsInRange(m_creature->getVictim(), 0, (m_LastSpellMaxRange / 1.5f)))
-                    SetCombatMovement(false, true);
-                else
+                bool uiInLineOfSight = m_creature->IsWithinLOSInMap(m_creature->getVictim());
```
Phatcat: This variable can be moved inside the if statement on line 1373 - I would also move the bool check to be the first instead of the spellcast check as that would seem heavier to me.

robinsch:
While you are at it, remove the entire SetCombatMovement and all DB related stuff because it should be all handled right here.

killerwife:
Feel free to PR it, I do not have time to work on AI as well. I exported this for @Laizerox

src/game/CreatureEventAI.cpp
```diff
@@ -1370,11 +1370,18 @@ void CreatureEventAI::UpdateAI(const uint32 diff)
...
+                
+                if (uiInLineOfSight && !m_creature->IsNonMeleeSpellCasted(false))
```
Phatcat: `if (bool uiInLineOfSight = m_creature->IsWithinLOSInMap(m_creature->getVictim())`

If you change this line to that, and wrap line 1379 through 1382 in a new if statement like
`if  (!m_creature->IsNonMeleeSpellCasted(false))`

then you can change the else if into a simple else, and you'll be rid of line 1375 altogether.

cala:
I'd be happy to test this and merge it but what is this supposed to fix, please? @killerwife @Laizerox 
So I can know how to test it properly because the dynamic movement seem quite good to me as it is, so if I don't know what to expect/look at, I won't be able to review it properly. :smile:

killerwife:
When the creature runs out of LOS, it wont follow the player when dynamic movement is engaged and combatMovement is disabled. (with this change it will) Besides, dynamic movement still has a hard to track bug related to max range, but currently unable to reproduce.

cala:
In https://github.com/cmangos/mangos-tbc/commit/79de6bf20ebc6af123f2e2c9b8c742fbcabf082d
Thank you.

PR #58, opened 2016-05-30
Remove powerCost from triggered spells

killerwife:
@robinsch this should be more to your liking then.
@cala afaik we are missing it from all cores

Example: devastate triggers sunder armor for no cost. (possibly many more)

### DIFF
--- a/src/game/CreatureAI.cpp
+++ b/src/game/CreatureAI.cpp
@@ -52,7 +52,7 @@ CanCastResult CreatureAI::CanCastSpell(Unit* pTarget, const SpellEntry* pSpell, 
             return CAST_FAIL_STATE;
 
         // Check for power (also done by Spell::CheckCast())
-        if (m_creature->GetPower((Powers)pSpell->powerType) < Spell::CalculatePowerCost(pSpell, m_creature))
+        if (m_creature->GetPower((Powers)pSpell->powerType) < Spell::CalculatePowerCost(pSpell, m_creature, isTriggered))
             return CAST_FAIL_POWER;
 
         if (!m_creature->IsWithinLOSInMap(pTarget) && m_creature != pTarget)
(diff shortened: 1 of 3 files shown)


robinsch:
No need to pass it as a parameter

killerwife:
Its a static function

cyberium:
Like @robinsch i think its better to remove that new argument.
CalculatePowerCost have then to be used only when we realy want to know that cost.
So everywere we can encounter triggered spell we have to add a check for it.

´´´cpp
uint32 powerCost = isTriggered ? 0 : CalculatePowerCost();
´´´
Or we can implement a more general method like GetPowerCost() and GetStaticPowerCost(...)?

robinsch:
Remove the static and let it return 0 if spell is triggered

killerwife:
@cyberium the function is used in 3 places, you require 100% exact code each time and we calculate it in AIs if the source needs to know if it has enough mana. I.e. you dont have spell*. Removing argument would mean the need for 2 equally same functions.

cyberium:
@robinsch You probably think about something like this inside the method
```cpp
if (spell && spell->IsTriggered()
    return 0;
```
wich is smart too :)

@killerwife And? Just call it GetStaticPowerCost() within GetPowerCost() and handle the special case for instanciated Spell vs not instanciated only in second case. Because currently i dont see the point to send own spell pointer when calling CalculatePowerCost(...).

```cpp
m_powerCost = CalculatePowerCost(m_spellInfo, m_caster, this, m_CastItem); // << bad use of 'this' in this case because we are already in spell class.
```

killerwife:
Then why dont we just send spellinfo in the first place?

killerwife:
Nvm, scratch that, its because of applyspellmod, which is something i am working on reworking. Yuck.

PR #61, opened 2016-06-10
Spell scripting implementation (Make spellEffects.cpp (not)great again)

killerwife:
Implementation of spell scripting outside of those ugly switch cases. Its a system that can be easily extended.

Disclaimer: It only works for VS 140/VS 120 right now and as an example there is a port of my 
http://www.wowhead.com/quest=10857/teleport-this script implementation.

This could possibly solve a lot of headaches :)

@cala @Laizerox @cyberium @xfurry and whomever else this may concern.

PS: To make it link I had to use some sorcery so possibly might not be "standard" ^^

### DIFF
--- a/src/game/SpellScriptsTBC.cpp
+++ b/src/game/SpellScriptsTBC.cpp
@@ -8,8 +8,7 @@ enum
 
 struct SoulMirror : public Spell
 {
-    SoulMirror(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr)
-        : Spell(caster, info, triggered, originalCasterGUID, triggeredBy) {}
+    using Spell::Spell;
 
     void EffectDummy(SpellEffectIndex eff_idx) override
     {
@@ -34,11 +33,7 @@ enum
 /*Spell 38915*/
 struct MentalInterference : public Spell
 {
-    MentalInterference(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr)
-        : Spell(caster, info, triggered, originalCasterGUID, triggeredBy)
-    {
-
-    }
+    using Spell::Spell;
 
     SpellCastResult CheckCast(bool strict) override
     {
@@ -70,11 +65,7 @@ enum
 /*Spell 38920*/
 struct DetonateTeleporter : public Spell
 {
-    DetonateTeleporter(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr)
-        : Spell(caster, info, triggered, originalCasterGUID, triggeredBy)
-    {
-
-    }
+    using Spell::Spell;
 
     void EffectScriptEffect(SpellEffectIndex eff_idx) override
     {
--- a/src/game/SpellScriptsTBC.cpp
+++ b/src/game/SpellScriptsTBC.cpp
@@ -1,6 +1,30 @@
 #include "SpellScriptFactory.h"
 
-extern ScriptedSpellFactory* spellFactories[MAX_TBC_SPELL_ID];
+/*Spell 34063*/
+enum
+{
+    SPELL_SOUL_MIRROR = 34063
+};
+
+struct SoulMirror : public Spell
+{
+    SoulMirror(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr)
+        : Spell(caster, info, triggered, originalCasterGUID, triggeredBy) {}
+
+    void EffectDummy(SpellEffectIndex eff_idx) override
+    {
+        if (!unitTarget)
+            return;
+
+        m_caster->SummonCreature(19480, unitTarget->GetPositionX(), unitTarget->GetPositionY(), unitTarget->GetPositionZ(), unitTarget->GetOrientation(), TEMPSUMMON_TIMED_OOC_OR_DEAD_DESPAWN, 30000);
+        unitTarget->DealDamage(unitTarget, unitTarget->GetHealth(), nullptr, DIRECT_DAMAGE, SPELL_SCHOOL_MASK_NORMAL, nullptr, false);
+    }
+};
+
+Spell* SoulMirrorFactory(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr)
+{
+    return new SoulMirror(caster, info, triggered, originalCasterGUID, triggeredBy);
+}
 
 enum
 {
@@ -16,8 +40,12 @@ struct MentalInterference : public Spell
 
     }
 
-    SpellCastResult CustomCheckCast(bool strict) override
+    SpellCastResult CheckCast(bool strict) override
     {
+        SpellCastResult result = Spell::CheckCast(strict);
+        if (result != SPELL_CAST_OK)
+            return result;
+
         if (ObjectGuid target = m_caster->GetTargetGuid())
         {
             if (!(target.GetEntry() == 16943 || target.GetEntry() == 20928))  // Mental Interference can be cast only on these two targets
@@ -28,13 +56,11 @@ struct MentalInterference : public Spell
     }
 };
 
-struct MentalInterferenceFactory : public ScriptedSpellFactory
+
+Spell* MentalInterferenceFactory(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr)
 {
-    virtual Spell* operator ()(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr) override
-    {
-        return new MentalInterference(caster, info, triggered, originalCasterGUID, triggeredBy);
-    }
-};
+    return new MentalInterference(caster, info, triggered, originalCasterGUID, triggeredBy);
+}
 
 enum
 {
@@ -59,17 +85,15 @@ struct DetonateTeleporter : public Spell
     }
 };
 
-struct DetonateTeleporterFactory : public ScriptedSpellFactory
+Spell* DetonateTeleporterFactory(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr)
 {
-    virtual Spell* operator ()(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr) override
-    {
-        return new DetonateTeleporter(caster, info, triggered, originalCasterGUID, triggeredBy);
-    }
-};
+    return new DetonateTeleporter(caster, info, triggered, originalCasterGUID, triggeredBy);
+}
 
 void AddSpellFactoriesTBC()
 {
-    spellFactories[SPELL_MENTAL_INTERFERENCE] = new MentalInterferenceFactory();
-    spellFactories[SPELL_DETONATE_TELEPORTER] = new DetonateTeleporterFactory();
+    spellFactories[SPELL_SOUL_MIRROR] = SoulMirrorFactory;
+    spellFactories[SPELL_MENTAL_INTERFERENCE] = MentalInterferenceFactory;
+    spellFactories[SPELL_DETONATE_TELEPORTER] = DetonateTeleporterFactory;
 }
 
--- a/src/game/SpellScriptsTBC.cpp
+++ b/src/game/SpellScriptsTBC.cpp
@@ -1,0 +1,75 @@
+#include "SpellScriptFactory.h"
+
+extern ScriptedSpellFactory* spellFactories[MAX_TBC_SPELL_ID];
+
+enum
+{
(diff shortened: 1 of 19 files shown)


killerwife:
Theoretical progress: Add IsSkipLOSSpell function which we can override in script. Also IsPositiveEffect helper will be moved to Spell class so that we can override it in special cases. 

Also since @Laizerox is lazy and I have to do his work, he proposes to remove static table array and replace it with std::map with std::function to constructor functions.

On the note of using a singleton, we have to consider that spells are being used all the time and dereferencing a singleton each time might be very slow (since access to constructor functions doesnt need to by synchronized ?!)

robinsch:
This looks even more ugly then having it in switch cases.

killerwife:
1. Thanks for the constructive criticism (kappa)
2. Its not nearly finished
3. The main point is to concentrate spell code in one place. I had to script 3 effects and do some custom targeting for a single spell in TBC and it quickly became a mess for orientation.

Feel free to provide feedback tho.

Blumfield:
You could do it like trinitycore, it's more "clean" imo

killerwife:
Registering will be changed, but going with the "string" sd2 approach is redundant for spells tbh the way TC is doing it. I wanna avoid any sort of DB dependancy on this.

Phatcat:
@robinsch What a nice surprise to see you here. How about you pr some of your nost tbc work for cmangos so I wouldn't have to?

robinsch:
@Phatcat Take whatever you want from my commits or TBCPvP, I'm currently way too busy to do PRs but if you have any questions regarding my changes feel free to ask

Phatcat:
@robinsch Oh, you better believe that I'll take what I want. Yeah, you're way too busy working closed source yet again. Good for you.

Blumfield:
@Phatcat What's your problem exactly ? I'm giving feedback on Killerwife's change, this is what PR are made for. who I follow and what I contribute to is irrelevant, you're being hostile to everyone for no reason and you expect me to share what i've done ?

Phatcat:
Seriously? You giving feedback? Just what exacrtly is killerwife supposed to do with that comment?
Let me ask you; what have you exactly done and shared with cmangos now we're at it?
Get real, mate.

Blumfield:
He could get inspired and change few things or completely rewrite it using another approach. You can see I edited the wiki recently but once again what I do is irrelevant, i'm ending this conversation I don't have time for a child

Phatcat:
@Blumfield You're right, though. There's no reason to be this aggressive towards you. Sorry for that. Any contribution is appreciated, he just doesn't have much to go by with you guys comments...

xfurry:
The problem with this PR is that it changes all effects, instead of focusing on the cases which require our attention. 
As far as I know, most of spell scripting issues are with auras and targets, not with the effects themselves. 
I recommend that we start there.

killerwife:
It does if you want to. effects are an incredible mess at the moment. We want a design that can be used to script anything, not like TC, only what hooks enable. Also spells first auras later, as their behaviour is less intuitive and is heavily influenced by spells. 

Also, the reason I am designing this is that I wanna export my fixes without you hacking them SD2 script side.

Also I looked at TC implementation, and its heavily limited by hooks. we want a nice polymorphic spell design, instead of hook functions everywhere. feel free to hit me up with questions and ill explain my reasoning.

cyberium:
I think this is a good implementation of how spell script should work.
We may override almost what we want with minimal cpu/memory change and that is very cool.
Maybe it should be extended to be able to override specificboost too.

I think its not so hard to understaind for a dev and only need some practice to use it.
This is a good exemple of what we can do with polymorphism.
To make this more understandable i suggest you to give some more exemples of an currently hardcoded scripted spell translated to this system.

Thank you @killerwife for this PR.
We only have to identify all downside and limitation before implementing this in cmangos.

killerwife:
What do you mean by SpecificBoost?

I agree with the rest and will provide more samples.

VladimirMangos:
Just worry that like scripting will be easy way hide our limited info about proper spell attributes usage instead current spell ids/etc checks. Easy way that will hard reverted later when/if proper attribute descovered.

killerwife:
@VladimirMangos It will be harder to spot for those who are not involved in scripting, but I will easilly see it when I overload the same function too many times. Also we implemented some AuraInterruptFlags that TC had correctly guessed. There are many more like it but a lot of time has to be dedicated to it. I usually look at UNKs anytime im working with a new spell :) 

We are actually very close to having stealth work properly, only 2 more UNKs afaik.(and to my knowledge I have these things 1:1 on vengeance)

cyberium:
@VladimirMangos you are right, however, after 10 years of devs, i think we can start to implement a way to do properly some script/workaround that we will probably never know the blizzway.

xfurry:
Let's see some examples of this.
I would really love to implement the scripts described here: http://cmangos.net/thread-6700.html with spell scripts, one way or another.

http://paste2.org/VF7eW15F
http://paste2.org/vaHbkB2Z
http://paste2.org/xbC66sLA

@killerwife can you replicate the logic that I developed, with your concept?

killerwife:
Uhm, your concept relies on the SD2 principle of registering spells that I am trying to avoid (I want 0 DB dependency on this system).

You do these SpellAuraPeriodic and SpellTargetCheck and EffectAuraChange_spell_giddyup

Instead what you would do in a proper polymorphic scenario is overload Aura::PeriodickTick, Spell::CheckCast (i will explain this one a bit more) and Aura::WhateverAuraHandlerYouChoose


In regards to the checkcast, all you need to do is something like
SpellCastResult CheckCast(bool strict)
{
SpellCastResult result = Spell::CheckCast(strict);
if(result == SPELL_CAST_OK)
result = // do custom checks
return result;
}
And you dont have to do any of that ugly SD2 database registering and whatnot. All you do is register the factory function to the spell factories map (right now its a table, it will be changed) and thats it. Most importantly, almost 0 runtime computation overhead.

cyberium:
Yes @killerwife proposal is more flexible and less cpu consuming. It may require some learning to use it efficiently but its not so hard to understaind.
Iam still waiting a good exemple :)

killerwife:
Sorry, I am busy with launching a new server and last exam tomorrow, on friday I will expand this PR :)

killerwife:
@cyberium I removed some unnecessary fluff, still needs to be changed from table to std::unordered_map. Also still didnt unite on registering with @Laizerox due to server launch. Added another example, now how to implement a simple dummy (its much nicer than what is possible in SD2 now), Lastly changed checkcast to a less clunky, more powerful version.

src/game/SpellScriptsTBC.cpp
```diff
@@ -0,0 +1,75 @@
...
+enum
+{
+    SPELL_MENTAL_INTERFERENCE = 38915
+};
+
+/*Spell 38915*/
+struct MentalInterference : public Spell
+{
+    MentalInterference(Unit* caster, SpellEntry const* info, bool triggered, ObjectGuid originalCasterGUID = ObjectGuid(), SpellEntry const* triggeredBy = nullptr)
+        : Spell(caster, info, triggered, originalCasterGUID, triggeredBy)
```
Lillecarl: http://en.cppreference.com/w/cpp/language/using_declaration

For making this a bit easier we could just be typing using Spell::Spell;

It should work since C++11, which is kinda nice... If we wanna use the spell constructor, as we're doing here atm! :)

killerwife:
@Lillecarl i added your change

Lillecarl:
@killerwife nice! :-)

boxa:
oh, NO. very BAD design (code).
look at TC for example - sometimes very good...

cyberium:
@boxa better proposal?

robinsch:
^ same opinion

cyberium:
@robinsch same answer :)

robinsch:
TC's way

cyberium:
Well i dont realy want to study TC way, dont have time for it. Maybe @killerwife or someone esle could give us some comparaison and pro/cons?

boxa:
@cyberium as your known @boxa works on wow servers more six ears. from me a lot good changed in core (and in MaNGOS). but anyway im R2 man =) where all works. maybe a small hacks, but my players on servers is it not notice...
why im want to CMaNGOS it was the best...
a little of my critics, i think not hurt =)

boxa:
@cyberium 
explain please what is "MARTIAN way" =)

Phatcat:
@boxa pretty sure it's used to indicate a placeholder.

Also I'm pretty sure we already been over why @killerwife chose not to go with TC's method. you know this, robin, also, we already know your opinion, you made sure on that... You want killerwife to keep repeating himself?

I'm not saying you're wrong in suggesting to look at TC, but everything indicated that killerwife already did look at TC. He even directly said so, and voiced his concerns about their solution, on this very PR even.

cyberium:
@boxa yes dont mind about "Martian way".

About your remark we cannot be everywere and @cmangos/developer are busy or in holiday these days.

killerwife:
Alright, so. TrinityCore system is composed of programmable hooks which can be utilized for behaviour as SMARTSCRIPTS. Whilst I see the benefits of such a contraption, (enabling us to script everything from DB) I regard their system as the biggest set of redundancies I have ever seen. (keep in mind im not calling it stupid, if it works, its not stupid)

My idea behind spell scripting is that it should be a domain of c++, due to the high grain control required. I do not see the point of implementing a hook for that one spell out of 50k that is only in wotlk. Trinitycores core is riddled with hooks everywhere, which is not exactly the OOP way I am trying to achieve.

Yesterday @cyberium has brought up a point that this system doesnt enable bigger possibilities for DB scripter for spells. The thing is it does not diminish their capabilities either.

Also keep in mind this: you can override base spell, and implement hooks inside the child spell, whereas TC implemented these hooks in parent spell, which results in a lot more calls being done everywhere (and I do not like that).

So far, I only heard words like "bad" "do it like TC", do not view my standpoint as an ignorance to your comments. Give me hard evidence why my approach is bad, some serious constructive arguments, how my system can be improved or could be made more sleep, and then we can talk.

Also, the reason why I am registering in code, is because I want this system to be 100% independent from the DB. (registering wise) I am unsure how many of you have been trying to unify public TBC databases, as much as me and @X-Savior and devildog are doing, but its a real mess when all C++ changes need to be added to DB (looking at you SD2), not even considering it wastes a lot of time, trying everything to be in sync. Not to mention when I am exporting it to the public, for cmangos.

I understand I have a unique viewpoint, but such is life in wow-emu. Also, this turned kinda into a rant, so please bear with me.

Laizerox:
@boxa TC's approach are hooks that you call before or after the spell. But there are spells that require whole effects being overriden because of what spell does. Instead of implement "Do this before spell", then special cases in spell effect and then calling hooks "Do this afer spell is done" gives more overhead in terms of performance and whatnot is more confusing.

In killerwife's approach we're using "polymophism". Spell class is virtual and has set of default behavior. If some spell is really super special we can create a script for it just overriding the effect, checkcast or whatever we need instead of calling hooks before and after.

Also we want  cmangos to move forward not to copy others. It works for them but using these kind of system we could drop hardcoded bonus calculation also and move into their own scripts. Since spell class is virtual (and what not Aura) it gives us unmilimited amount of possibilities what we can do, while as using hook we're bound only to that.

xfurry:
Here are a few comments from my side regarding on this discussion:


* I don't fully agree with the following:

> My idea behind spell scripting is that it should be a domain of c++

Usually it's easier and makes more sense to script combat spells in C++ because you have more control over the events, and also the spell code is quite complex but also small.
However, many spells are used in cinematic type of events, like quests or dialogues. For this kind of things, we should really go for the DB approach. The reason is that these kind of events are usually quite simple from the scripting complexity, but also quite big in terms of data volume.
Your examples for spells ```/*Spell 38920*/``` or ```/*Spell 34063*/``` are pretty bad, as technically these spells **should** go into dbscripts.
(one SQL line VS thirty C++ lines... you decide which one makes more sense).

* Adding spell scripts for spells with effects 3 or 77, or for specific auras, doesn't help very much.
Why? Because these effects are already handled pretty well in SD2 when required and I think it's recommended to keep them like that. It makes way more sense to have a boss script and the related spell effects scripts in the same place.

The reason why I see spell scripts as a good solution is for examples like this: http://paste2.org/xbC66sLA where we need to script specific auras and spells which are specific for players and have no particular logic for scripting.

* So, overall I think that the idea is good.
But, as per the comments above, I suggest that we should limit this only to specific effects that are indeed required for scripting.

For example ```EffectSchoolDMG``` or  ```EffectWeaponDmg``` make perfect sense, but effects like ```EffectQuestComplete```or ```EffectSendTaxi``` or ```EffectPlaySound``` or ```HandleAuraControlVehicle``` shouldn't be script-able at all.

killerwife:
I agree with stuff @xfurry said to an extent. Dbscripting for stuff like talking etc must be possible, bcs there is a shitton of examples where its just stuff talking. Also I agree that changing SD2 implementations of dummies to my implementation is pointless if its used only on one encounter etc. (unless its a limiting factor). That overriding thing, uhm, every spell implementation will have to be reviewed anyway, and people can add the "virtual" keyword on their end anyway, so i dont see this as a big deal.

Lastly, I want to avoid SD2 like registering like u have in your auras example. Also we are talking about spells atm, not auras. Thats a completely different beast to cover.

Lillecarl:
I agree with the fact that sd2 stuff shouldn't be replaced with this, but
this should be seen as a replacement for the ugly special cases we have
around the spell system.

Cheers

xfurry:
> Lastly, I want to avoid SD2 like registering like u have in your auras example. Also we are talking about spells atm, not auras. Thats a completely different beast to cover.

Well, I think that this was the biggest flaw of my proposal.
And scripting auras is probably more important than spell effects, due to the fact that most of the issues currently are with auras.

> but
> this should be seen as a replacement for the ugly special cases we have
> around the spell system.

Yes, indeed. But replacing almost 10 years of dummy and script effect spells is a lot of work.
So probably we should start adding new spells first.
Also simple cases might still look better in the big list, than in a separate function on its own.

killerwife:
Yup, I can get behind that.

cyberium:
We can close this for now and getting back to this later.

PR #65, opened 2016-06-21
Implement TARGET_FLAG_GAMEOBJECT_ITEM

killerwife:
@cyberium


Phatcat:
Unless @cyberium (or anyone else for that matter) sees a problem with this, and reports it within the 14 day mark, then this is getting merged. It's a small one and it looks like we are already 3 core devs behind the change so there's no need to drag this one out.

Phatcat:
Merged.

PR #67, opened 2016-06-25
Fix looting of chests which require a key

killerwife:
@cyberium 
https://github.com/cmangos/issues/issues/777


killerwife:
Causing crashes. Will update when I have solutions.

PR #68, opened 2016-06-26
Redirect Chain Lightning to Grounding Totem as well

kartoffelsup:


killerwife:
Interesting

kartoffelsup:
Not sure if this has any negative impact on other chained damage spells, since I do not know of any other than Chain Lightning. If you can tell me how to get a list of all chained damage spells, I will thoroughly test those as well.

robinsch:
This will break all chain spells.

kartoffelsup:
How will it break all of them?

robinsch:
Because it will always search of chain targets if magnet returns nullptr, even if EffectChainTarget <= 1

kartoffelsup:
So what you're saying is, I should chain those two conditions with an or? :< Or would that not solve it?

robinsch:
No, it should simply just set the magnet target if we found one. There is no else condition.

kartoffelsup:
It appears that I misunderstood the magnet target.

kartoffelsup:
I'm sorry, I'll try to find a better solution :/

killerwife:
Also, should you wanna dynamically search spell dbc, i suggest you load it up into SQL. Works like a charm.

robinsch:
I prefer using QSW (https://github.com/anathemareborn/QSpellWork/releases) or download the source if you want to change something.

killerwife:
Switch case might be a way to go until we find out a more general behaviour. And QSW is fine, until you are doing SpellAttribute research and you need to be quick about it.

kartoffelsup:
Does this look better now?
>you need to change other calls where the target is set by this function
I let VisualStudio find all references (I hope this is reliable) and the only other place where it wasn't used with 
`if (Unit* pUnitTarget = m_caster->SelectMagnetTarget(target, this, effIndex))`
was Unit.cpp#2261 wich I adjusted within my previous commit.

robinsch:
All you have to do is set the target to magnet if we found a magnet, otherwise the target is stored in m_targets.getUnitTarget() and execute code https://github.com/cmangos/mangos-tbc/blob/master/src/game/Spell.cpp#L1670 if EffectChainTarget > 1

kartoffelsup:
Not sure if I understand you correctly, but it sounds like I did exactly what you describe here. I just removed the giant if/else block and replaced it with if/breaks.

robinsch:
Then it should be fine, but please check the second paragraph. I'm pretty sure this breaks all spells with TARGET_SINGLE_ENEMY and TARGET_DUELVSPLAYER

kartoffelsup:
Alright, I will check all calls to select magnet target again to make sure! :)

kartoffelsup:
It this what you had in mind?

robinsch:
Yes, this should be fine. (Except for coding style)

kartoffelsup:
You mean the curly braces for if/else with only a single statement?

robinsch:
Yes

kartoffelsup:
Darn, I always forget them :( 
I will apply the code style asap.

src/game/Spell.cpp
```diff
@@ -1659,70 +1659,76 @@ void Spell::SetTargetMap(SpellEffectIndex effIndex, uint32 targetMode, UnitList&
...
+            float max_range;
+            if (m_spellInfo->DmgClass == SPELL_DAMAGE_CLASS_MELEE)
+                max_range = radius;
+            else
+                // FIXME: This very like horrible hack and wrong for most spells
+                max_range = radius + unMaxTargets * CHAIN_SPELL_JUMP_RADIUS;
 
-                tempTargetUnitMap.sort(TargetDistanceOrderNear(pUnitTarget));
+            UnitList tempTargetUnitMap;
+            {
```
robinsch: ?

kartoffelsup:
How should I proceed here?

PR #69, opened 2016-06-29
Pet Patch Pistachio (TBC)

Phatcat:
Save DoSummon player pets (mage and priest summons) in DB to save and
load pet reaction state amongst other settings

Handle statues properly as guardians now that the core supports passive pets.

Ref Issue: https://github.com/cmangos/issues/issues/882

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-tbc/69)

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -3577,28 +3577,22 @@ void Spell::EffectSummonType(SpellEffectIndex eff_idx)
                     break;
                 case UNITNAME_SUMMON_TITLE_GUARDIAN:
                 {
-                    if (prop_id == 61)                      // mixed guardians, totems, statues
+                    if (prop_id == 61)
                     {
-                        // * Stone Statue, etc  -- fits much better totem AI
-                        if (m_spellInfo->SpellIconID == 2056)
-                            DoSummonTotem(eff_idx);
-                        else
+                        // Totem cases
+                        if (CreatureInfo const* cInfo = ObjectMgr::GetCreatureTemplate(m_spellInfo->EffectMiscValue[eff_idx]))
                         {
-                            // possible sort totems/guardians only by summon creature type
-                            CreatureInfo const* cInfo = ObjectMgr::GetCreatureTemplate(m_spellInfo->EffectMiscValue[eff_idx]);
-
-                            if (!cInfo)
-                                return;
-
-                            // FIXME: not all totems and similar cases seelcted by this check...
                             if (cInfo->CreatureType == CREATURE_TYPE_TOTEM)
+                            {
                                 DoSummonTotem(eff_idx);
-                            else
-                                DoSummonGuardian(eff_idx, summon_prop->FactionId);
+                                break;
+                            }
                         }
+                        else
+                            return;
                     }
-                    else
-                        DoSummonGuardian(eff_idx, summon_prop->FactionId);
+
+                    DoSummonGuardian(eff_idx, summon_prop->FactionId);
                     break;
                 }
                 case UNITNAME_SUMMON_TITLE_TOTEM:
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -3639,14 +3639,37 @@ void Spell::DoSummon(SpellEffectIndex eff_idx)
     if (!pet_entry)
         return;
 
+    Pet* spawnCreature = new Pet();
+
+    if (m_caster->GetTypeId() == TYPEID_PLAYER)
+    {
+        // Load pet from db; if any to load
+        if (m_caster->getClass() != CLASS_PRIEST
+            && spawnCreature->LoadPetFromDB((Player*)m_caster, pet_entry))
+        {
+            spawnCreature->SetHealth(spawnCreature->GetMaxHealth());
+            spawnCreature->SetPower(POWER_MANA, spawnCreature->GetMaxPower(POWER_MANA));
+
+            // set timer for unsummon
+            if (m_duration > 0)
+                spawnCreature->SetDuration(m_duration);
+
+            spawnCreature->SavePetToDB(PET_SAVE_AS_CURRENT);
+            return;
+        }
+
+        spawnCreature->setPetType(SUMMON_PET);
+    }
+    else
+        spawnCreature->setPetType(GUARDIAN_PET);
+
     CreatureInfo const* cInfo = ObjectMgr::GetCreatureTemplate(pet_entry);
     if (!cInfo)
     {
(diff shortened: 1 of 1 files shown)


robinsch:
I'm pretty sure this breaks temp summon locations (e.g. Shadowfiend)

Phatcat:
How? - Exactly the shadowfiend and the water elemental were the ones I tested

Phatcat:
Ah, the summon location, got you. I think you're right on that one. Maybe only save and load mage pet then (water elemental) as that one is summoned besides the owner always anyways, right?

robinsch:
Since Shadowfiend should spawn aggressively no matter what you could also do this - but well that would be just another workaround. Maybe just safe reaction state.

Phatcat:
Maybe just do new summon on priest pet, force reaction state to aggressive in the end and that's it.
For water elemental save and load to and from db I still think is the best option since we also need to save and load spell settings, etc., (just not being visually updated to the player atm so to the player it's always displayed as default while in fact it has the settings it was last saved with-)

Phatcat:
@robinsch So what do you think now?

src/game/SpellEffects.cpp
```diff
@@ -3700,15 +3714,19 @@ void Spell::DoSummon(SpellEffectIndex eff_idx)
 
     if (m_caster->GetTypeId() == TYPEID_PLAYER)
     {
-        // this enables popup window (pet dismiss, cancel)
         spawnCreature->SetUInt32Value(UNIT_FIELD_FLAGS, UNIT_FLAG_PVP_ATTACKABLE);
 
-        spawnCreature->GetCharmInfo()->SetPetNumber(pet_number, true);
+        CharmInfo* charmInfo = spawnCreature->GetCharmInfo();
+        if (m_caster->getClass() == CLASS_PRIEST)
```
robinsch: Rather check for Effect SPELL_EFFECT_SUMMON and TARGET_CURRENT_ENEMY_COORDINATES.

src/game/SpellEffects.cpp
```diff
@@ -3638,23 +3638,41 @@ void Spell::DoSummon(SpellEffectIndex eff_idx)
...
+
+    if (m_caster->GetTypeId() == TYPEID_PLAYER)
+    {
+        // Load pet from db; if any to load
+        if (m_caster->getClass() == CLASS_MAGE && spawnCreature->LoadPetFromDB((Player*)m_caster, pet_entry)) 
+        {
+            spawnCreature->SetHealth(spawnCreature->GetMaxHealth());
+            spawnCreature->SetPower(POWER_MANA, spawnCreature->GetMaxPower(POWER_MANA));
+
+            spawnCreature->SavePetToDB(PET_SAVE_AS_CURRENT);
```
robinsch: Why call save here?

src/game/SpellEffects.cpp
```diff
@@ -3638,23 +3638,41 @@ void Spell::DoSummon(SpellEffectIndex eff_idx)
...
+
```
robinsch: remove line

Phatcat: what, the empty line?

src/game/SpellEffects.cpp
Phatcat: Given that we are already within a check that has confirmed us to be a player, wouldn't this check be much faster? do other players have summons like these?

src/game/SpellEffects.cpp
```diff
@@ -3638,23 +3638,41 @@ void Spell::DoSummon(SpellEffectIndex eff_idx)
...
+    Pet* spawnCreature = new Pet();
+
+    // set timer for unsummon
+    if (m_duration > 0)
+        spawnCreature->SetDuration(m_duration);
+
+    if (m_caster->GetTypeId() == TYPEID_PLAYER)
+    {
+        // Load pet from db; if any to load
+        if (m_caster->getClass() == CLASS_MAGE && spawnCreature->LoadPetFromDB((Player*)m_caster, pet_entry))
```
robinsch: Aren't Shaman Elemental also controllable?

src/game/SpellEffects.cpp
robinsch: Yes

src/game/SpellEffects.cpp
robinsch: Atleast not in TBC from what I know. But if you want to keep it sync with others, DK Gargoyle for example should also start attacking on Summon. (Which is currently handled by a hack in TrinityCore)

Phatcat: Yes, the more in sync the better; I'll go for your check. All I know is that these types of pets weren't introduced before tbc so they do not exist at all in vanilla.

src/game/SpellEffects.cpp
Phatcat: Sure, why not.

src/game/SpellEffects.cpp
Phatcat: I have no idea honestly - At least I'm pretty sure the druid treants are NOT controllable, but I am not even 100% sure on that one, I switched from druid to rogue from retail vanilla to tbc..

src/game/SpellEffects.cpp
robinsch: My bad, they should not be aggressive. Just start attacking there target no matter what reaction stat

Phatcat: The thing is that it is passive since it is not given a react state, and while defensive it would attack its target, on passive it just runs back to its owner, so we have to overwrite it to be either defensive or aggressive and aggressive makes more sense in my opinion. Are you sure the same does not go for DK gargoyles?

Phatcat: Oh yeah, wrath doesn't have this function at all...

robinsch: Shouldn't the default react state be defensive?

src/game/SpellEffects.cpp
Phatcat: Okay so Shaman pets come from totems.. Are you sure they are controllable?

src/game/SpellEffects.cpp
Phatcat: Alright.

src/game/SpellEffects.cpp
Phatcat: Just gonna change it to getClass() != CLASS_PRIEST instead, done deal.

Phatcat:
@robinsch How about now?
@killerwife Would love to know what you think of this as well.

@cala dunno if you're on TBC as well, if not then just overlook the namedrop :tongue:

PR #71, opened 2016-07-05
Game_event linkedTo addition

killerwife:
PRing this as per @X-Savior request

Adds the ability to link game_events

@Phatcat if you can, try to expedite this one, TBC-DB is being exported

### DIFF
--- a/src/game/GameEventMgr.cpp
+++ b/src/game/GameEventMgr.cpp
@@ -108,7 +108,7 @@ void GameEventMgr::LoadFromDB()
         mGameEvent.resize(max_event_id + 1);
     }
 
-    QueryResult* result = WorldDatabase.Query("SELECT entry,UNIX_TIMESTAMP(start_time),UNIX_TIMESTAMP(end_time),occurence,length,holiday,description FROM game_event");
+    QueryResult* result = WorldDatabase.Query("SELECT entry,UNIX_TIMESTAMP(start_time),UNIX_TIMESTAMP(end_time),occurence,length,holiday,linkedTo,description FROM game_event");
     if (!result)
     {
         mGameEvent.clear();
@@ -123,7 +123,6 @@ void GameEventMgr::LoadFromDB()
         BarGoLink bar(result->GetRowCount());
         do
         {
-            ++count;
             Field* fields = result->Fetch();
 
             bar.step();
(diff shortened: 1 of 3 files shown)


Phatcat:
Okay, I'll look at this asap.

robinsch:
What is this for? I mean I get what it does.

src/game/GameEventMgr.cpp
```diff
@@ -656,12 +680,17 @@ void GameEventMgr::UnApplyEvent(uint16 event_id)
 void GameEventMgr::ApplyNewEvent(uint16 event_id, bool resume)
 {
     m_ActiveEvents.insert(event_id);
+    sLog.outString("GameEvent %u \"%s\" started.", event_id, mGameEvent[event_id].description.c_str());
+    if (event_id==987)
```
robinsch: code format and magic numbers

killerwife: Daily restart, nuff said.

killerwife:
Have an event like brewfest, or darkmoon faire, that has several smaller components, like, lets say, hourly events with some smaller shenanigans. You NEVER want this smaller event to be started when the bigger event is not started, lets say, because of spawns or maybe something else. This adds the ability to link events to each other, so that some cannot be started if others are not running.

src/game/GameEventMgr.cpp
robinsch: Comment was missing before.

robinsch:
Are you sure it's the right way to handle those smaller events with game_event? I mean even if you could simply add the start time, end time and change the occurence that would have the same effect.

killerwife:
Well, I dont design game-events, and that wouldnt provide this failsafe. + I am not the author of this change, as it was done way before my time.

killerwife:
@cala @xfurry @Grz3s Please review it :) since it is a DB change.

Phatcat:
I see no problem with this, but then again, I am not big on game events, so I can't vouch for anything beyond the codestyle.

dagochen:
I would do this with a different approach. We allready have conditions, so why not give an event a condition?

Phatcat:
Wrath https://github.com/cmangos/mangos-wotlk/commit/7e23f3cc02cc4f8c339f973e1169d41c7eb30792
TBC https://github.com/cmangos/mangos-tbc/commit/e8771822dbd0b919001d2a574a0c72f35ce2c917
classic https://github.com/cmangos/mangos-classic/commit/b084f64d44342e75fd49f477bd677003cecebc20

PR #73, opened 2016-07-06
[ITEM] Expedition flare not working

Ccaz:
Item Expedition Flare summons either an Expedition Preserver or Expedition Scout to assist the Player.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -1149,6 +1149,13 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
 
                     return;
                 }
+                case 32027:                                 // Expedition Flare
+                {
+                    // 32029 = Expedition Preserver | 32030 = Expedition Scout
+                    m_caster->CastSpell(m_caster, (urand(0, 1) ? 32029 : 32030), true);
+
+                    return;
+                }
                 case 32146:                                 // Liquid Fire
                 {
                     if (!unitTarget || unitTarget->GetTypeId() != TYPEID_UNIT || m_caster->GetTypeId() != TYPEID_PLAYER)


Ccaz:
DB update for Expedition Preserver and Expedition Scout is also necessary. They are both level 1 NPC's and have 8hp. If desired I can deliver a sql update.

Phatcat:
Hey man, thank you for you PR;

Right off the bat; we don't like tabs here. We love whitespaces, however, so we use 4 whitespaces to represent a tab.

I'm gonna ask stupid here because I'm stupid; you're sure that the spell cast by the item indeed does have a dummy effect?

If the NPCs summoned by the spell are bugged and should not be level 1, and indeed have more hp, then yes, you should very much consider doing a DB update as well =)

Ccaz:
Hi,

I used the Tool QSW to look it up. The item has both EffectScriptEffect and EffectDummy.
 
I am fairly new to cmangos, just started like 2 days ago so I used EffectScriptEffect instead of EffectDummy. Don't know the difference yet and what would be better.

It's hard to find some information about this item on the wikis. The CC-Buglist provides some information though. https://github.com/ccshiro/cc-buglist/issues/1786

I will think about the whitespaces and provide a sql update.

Phatcat:
EffectScriptEffect

It sounds like this item spell needs to be scripted; I am fairly unfamiliar with the spell system so I will not be able to help you there; but the dummy effect itself looks fine (besides codestyle)

Regarding codestyle; there is no reason to set a float variable if you are only going to use it in one spot; just write it directly into where it's needed.

You should NEVER manually summon creatures if it can be avoided; I am 99% sure there is a dedicated spell for those summons.

killerwife:
@Phatcat He is from me, I told him to do that :D The dummy summons the mob because there is a random chance which mod can be summoned, the template spell system doesnt support that, thats why its scripted in a SPELL_EFFECT_DUMMY

src/game/SpellEffects.cpp
```diff
@@ -5457,6 +5457,33 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
...
                 }
+                
+                
+                case 32027:				// Expedition Flare
+			{
+				if (!unitTarget)
+					return;
+
+				uint32 creature_id;
+				uint32 rand = urand(0, 99);
```
robinsch: urand (0,1)

src/game/SpellEffects.cpp
```diff
@@ -5457,6 +5457,33 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
...
+
+				//summon Expedition Preserver or Expedition Scout 
+				if (rand < 50)
```
robinsch: if (rand)

Phatcat:
Not exactly what I meant. I'm gonna comment diretly in the code ^^

src/game/SpellEffects.cpp
```diff
@@ -5457,6 +5457,33 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
...
+					creature_id = 18194;
+				else 
+					creature_id = 18126;
+		
+				float angle = unitTarget->GetOrientation();
+			
+				Creature* pCreature = m_caster->SummonCreature(creature_id, 0.0f, 0.0f, 0.0f, angle, TEMPSUMMON_TIMED_OOC_OR_DEAD_DESPAWN, 60000, true);
```
Phatcat: This call seems wrong to me;

I am 99% sure there is a spell which takes care of what is being done with this call right here.

src/game/SpellEffects.cpp
```diff
@@ -5457,6 +5457,33 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
...
+
+				uint32 creature_id;
+				uint32 rand = urand(0, 99);
+
+				//summon Expedition Preserver or Expedition Scout 
+				if (rand < 50)
+					creature_id = 18194;
+				else 
+					creature_id = 18126;
+
```
Phatcat: You can replace the whole if/else shabang as well as the 2 variables if you copy the method used for this dummy effect: https://github.com/cmangos/mangos-tbc/blob/master/src/game/SpellEffects.cpp#L799 (which by the way is superb code style as well)

src/game/SpellEffects.cpp
```diff
@@ -5457,6 +5457,33 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
...
+			
+				Creature* pCreature = m_caster->SummonCreature(creature_id, 0.0f, 0.0f, 0.0f, angle, TEMPSUMMON_TIMED_OOC_OR_DEAD_DESPAWN, 60000, true);
+
+				if (pCreature->AI())
+					pCreature->AI()->AttackStart(m_caster->getVictim());
+				return;
+
+			}
+                
+
```
xfurry: Please use the code style guidelines https://github.com/cmangos/issues/wiki/Coding-Standards

Also, don't summon creature as ActiveObject (remove the last summon parameter). and maybe also check for ```m_caster->getVictim())``` in case there is no victim.

Finally no [Hungarian notation](https://msdn.microsoft.com/en-us/library/windows/desktop/aa378932%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) for variables in SpellEffects (don't ask me why, but this is how everything was agreed before)

src/game/SpellEffects.cpp
killerwife: There is a ton of serverside spells that summon creatures (i have a suspicion blizz summons all like that), problem is, a shitton of these serverside spells are not in client spell.dbc, hence we have to use summonCreature function, i.e. the usage is correct here

killerwife:
I understood what you meant, but you didnt what I did :D Blizz used dummies in strange ways, some are even used with resilience aided damage calculations ^^

src/game/SpellEffects.cpp
```diff
@@ -5457,33 +5457,21 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
...
+                case 32027:                                 // Expedition Flare
+                {
+                    if (!unitTarget)
+                        return;
+
+                    uint32 creature_id = urand(0, 1)
+                                         ? 18194    //summon Expedition Preserver
+                                         : 18126;   //summon Expedition Scout
+
+                    Creature* creature = m_caster->SummonCreature(creature_id, 0.0f, 0.0f, 0.0f, unitTarget->GetOrientation(), TEMPSUMMON_TIMED_OOC_OR_DEAD_DESPAWN, 60000);
```
killerwife: Have you tried using the dummy effect and look at what m_Targets points to? its dynamic_object_front if i remember correctly, which under some circumstances could give proper location for creature summon, rather than in the caster
if it doesnt work, try to change it so that the creature is summoned next to caster

killerwife:
Also one thing to not, instead of closing PR you can do git push -f and it will update PR after uve edited your commits :)

src/game/SpellEffects.cpp
Phatcat: I can note that blizzard does not summon all like that; an example would be taxi summons; we are handling many like the method above; but this creates issues, and actually quite a few of them do exist in the client spell .dbc, cala found one in particular.

Phatcat:
Id: 32029
Name: Summon Expedition Preserver
General info
Category = 0, SpellIconId = 1, ActiveIconId = 0, SpellVisual = (0, 0)
SpellFamilyName = SPELLFAMILY_GENERAL, SpellFamilyFlags = 0x0000000000000000
SpellSchoolMask = 1 (SCHOOL_MASK_PHYSICAL, SCHOOL_MASK_ALL)
DamageClass = 0 (SPELL_DAMAGE_CLASS_NONE)
PreventionType = 0 (SPELL_PREVENTION_TYPE_NONE)
Advanced info
Spell Level = 0, BaseLevel 0, MaxLevel 0, MaxTargetLevel 0
Category = 0
DispelType = 0 (DISPEL_NONE)
Mechanic = 0 (MECHANIC_NONE)
SpellRange (Id 1) "Self Only"
MinRange = 0
MaxRange = 0
CastingTime (Id 1) = 0.00
Duration: ID (3) 60000, 0, 60000
Interrupt Flags: 0x00000000, AuraIF 0x00000000, ChannelIF 0x00000000
Chance = 101, charges - 0
Effects
Effect 0:
Id: 28 (SPELL_EFFECT_SUMMON)
BasePoints = 1
Targets (32, 0) (TARGET_MINION, NO_TARGET)
EffectMiscValueA = 18194
EffectMiscValueB = 61
Radius (Id 9) 20.00
Effect 1:
NO EFFECT
Effect 2:
NO EFFECT

Phatcat:
Id: 32030
Name: Summon Expedition Scout
General info
Category = 0, SpellIconId = 1, ActiveIconId = 0, SpellVisual = (0, 0)
SpellFamilyName = SPELLFAMILY_GENERAL, SpellFamilyFlags = 0x0000000000000000
SpellSchoolMask = 1 (SCHOOL_MASK_PHYSICAL, SCHOOL_MASK_ALL)
DamageClass = 0 (SPELL_DAMAGE_CLASS_NONE)
PreventionType = 0 (SPELL_PREVENTION_TYPE_NONE)
Advanced info
Spell Level = 0, BaseLevel 0, MaxLevel 0, MaxTargetLevel 0
Category = 0
DispelType = 0 (DISPEL_NONE)
Mechanic = 0 (MECHANIC_NONE)
SpellRange (Id 1) "Self Only"
MinRange = 0
MaxRange = 0
CastingTime (Id 1) = 0.00
Duration: ID (3) 60000, 0, 60000
Interrupt Flags: 0x00000000, AuraIF 0x00000000, ChannelIF 0x00000000
Chance = 101, charges - 0
Effects
Effect 0:
Id: 28 (SPELL_EFFECT_SUMMON)
BasePoints = 1
Targets (32, 0) (TARGET_MINION, NO_TARGET)
EffectMiscValueA = 18126
EffectMiscValueB = 61
Radius (Id 9) 20.00
Effect 1:
NO EFFECT
Effect 2:
NO EFFECT

Phatcat:
And now I'm 100% sure.

src/game/SpellEffects.cpp
Phatcat: Please refer to my latest comments.

src/game/SpellEffects.cpp
Phatcat: There are actual spells to take care of the summons, those spells should hopefully make the spells behave properly =)

killerwife:
Fair enough.

Phatcat:
Hold off on the DB update as well until you have tested it using the dedicated spells; it may solve some of the further issues you mentioned.

Ccaz:
Should be something like this then, I am @work right now, I will test it properly at the weekend, thanks guys.

`
                               
	case 32027:
        {
			if (!unitTarget)
				return;

			uint32 spell_id = urand(0, 1)
				? 32029    //summon Expedition Preserver
				: 32030;   //summon Expedition Scout

			m_caster->CastSpell(m_caster, spell_id, true, m_CastItem);
			return;

		}



`

Phatcat:
No empty lines before a 'bracket' line, though.

Phatcat:
It can actually be simplified even more into just

	case 32027:
        {
		if (!unitTarget)
			return;

		// 32029 = summon Expedition Preserver | 32030 = summon Expedition Scout
		m_caster->CastSpell((urand(0, 1) ? 32029 : 32030), spell_id, true, m_CastItem);
		return;
	}

which will spare the system the variable.

Phatcat:
Oh yeah, you are using m_CastItem, then you actually want to make sure that it actually exists, so you should do a check for it exactly like you do for unitTarget.

Phatcat:
but you are however not using unitTarget anymore, so that check is redundant, so you should simply replace !unitTarget with !m_CastItem

Ccaz:
Shouldnt it be like that? ;)
`
m_caster->CastSpell(m_caster,(urand(0, 1) ? 32029 : 32030), true, m_CastItem);
`

Phatcat:
I noticed the error; I think I edited that comment 3 or 4 times before I got it right... I am really stoned right now xD

Phatcat:
.. It can be simplified even more xD

	case 32027: // summon Expedition helper
	{
		if (m_CastItem)
			m_caster->CastSpell(m_caster, (urand(0, 1) ? 32029 : 32030), true, m_CastItem);
			// 32029 = Expedition Preserver | 32030 = Expedition Scout

		return;
	}

Ccaz:
hmm did a git push -f and now he closed the PR?

Phatcat:
You forgot to add the new commit to this branch.

Phatcat:
So you pushed an empty branch (one that is current with master), hence it has no changes so it closes itself.

src/game/SpellEffects.cpp
```diff
@@ -1149,6 +1149,13 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
 
                     return;
                 }
+                case 32027:                                 // Expedition Flare
+                {
+                    if (!m_CastItem)                        // 32029 = Expedition Preserver | 32030 = Expedition Scout
```
Phatcat: remove ! as we now use the if condition for the reverse functionality =)
Also I would move the comment on this line back a few spaces; in my opinion they shouldn't be alligned with the case comments but it's not a biggie.

Ccaz: Well If i remove the ! the spell will not work anymore :D any ideas?

robinsch: Is it even required to check for nullptr here? Shouldn't this be checked in CastSpell?

Phatcat: That sounds very very weird, because (!m_CastItem) means NOT CastItem, so that means it only works if spell is not cast from an item? That sounds weird. I believe you, but it sounds like something is broken somehow.

Phatcat: The immediate fix would be to remove the check altogether and not send the m_CastItem ref along with the spell cast; this just seems wrong to me...

Phatcat: @robinsch Not according to just about every other dummy effect; if nothing else it's a failsafe making sure we don't send off a nullptr to a spellcast that might rely on an actual reference should we for some odd reason not have one when we reach this point.

Phatcat: Wait a second... This Flare gun, is it perhaps one that is consumed on use? because then we of course cannot get the ref to the item as it doesn't exist anymore.

robinsch: Items should ne destroyed after cast, wasn't there a recent commit that addresses this issue?

Ccaz: The Items disappears from your inventory if you use it.

Phatcat: Well, both yes and no; there was a recent commit related to this; but that was related to channeled spell casts which would crash the server if they came from an item and the item was deleted before the channel was finished.

robinsch: I mean as @Ccaz stated this code works, which means the check can be removed and m_CastItem reference should not be passed

Phatcat: You're right, it's fine to send a nullptr, nothing bad can happen, well, the spell may not function properly, but in this case it propably will since blizz must have run into the same kind of thing;
This was also my suggestion, I just wasn't aware that the item was consumed on use, now that I know it is, then removing the check doesn't seem all that wrong to me anymore.

Phatcat: @robinsch I would very much agree with you there; I have no idea why our system is designed so that items are removed before we actually done with them.

killerwife: Its not, but what if the player suddenly DCed, or other shenanigans were afoot

Phatcat: Then why won't the spell work if you do a check for m_CastItem with this dummy case then, since this is cast from an item, if our system is not designed so that items are removed before we're done using them?

killerwife: Because you add it to CastSpell, which I usually avoid, at least on voa end, because when I am designing stuff I test it without items, unless it provides some special, needed functionality. Maybe the m_CastItem parameter in CastSpell isnt even needed in this instance.

Phatcat: That does not explain why m_CastItem is a nullptr when the spell is very clearly cast from an item.

Phatcat: also, if you would care to note, every other dummy effect with a CastItem has a check for the CastItem, and all dummy effects that come from items send along their CastItem; so it is not something we want to avoid doing, we would much rather want to find out why our CastItem is getting nullified. Again, my guess would be because it's consumed on use.

killerwife: Well, then you wouldnt wanna send the m_castItem along at all, because it will expire every time. It depends on how you look at cast items, either A) spell is tied to castItem or B) castItem is tied to spell, when its a key unlock spell i assume the former, while when its some script spell, I assume the latter. This cas is in my opinion the latter.

xfurry: you don't need to check m_castItem in the effect function. You can remove that completely.
You should only care about the target.

Phatcat: Does that go for *every* spell effect dummy? Cause then I'm gonna update my simplify PR.

src/game/SpellEffects.cpp
```diff
@@ -1151,7 +1151,7 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
                 }
                 case 32027:                                 // Expedition Flare
                 {
-                    if (!m_CastItem)                        // 32029 = Expedition Preserver | 32030 = Expedition Scout
+                        // 32029 = Expedition Preserver | 32030 = Expedition Scout
                         m_caster->CastSpell(m_caster, (urand(0, 1) ? 32029 : 32030), true, m_CastItem);
```
Phatcat: m_caster->CastSpell(m_caster, (urand(0, 1) ? 32029 : 32030), true)

- Drop CastItem altogether from this one, since CastItem is nullptr in this case and nullptr is the default it can be considered redundancy.

src/game/SpellEffects.cpp
xfurry: It depends on the actual logic of the effect.

If your target is an item, then we need to check the itemTarget.
If the target is a gameobject, we need to check for the gameobject.
Also, many effects have logic that will check for the TypeId() of caster or target.

But usually we don't need to check for m_caster, because that can't be NULL.
The caster was already used many times in the cast function before the spell effects are triggered.

Phatcat: Actually I was refering to the other spell effect dummies which have the (m_CastItem) check which use m_CastItem exactly like it was used here, not as the target.

I am just wondering if some spells may rely on actually having the m_CastItem passed along; why else would we be doing it at all if not?

src/game/SpellEffects.cpp
Phatcat: Also since you removed the 'if' statement both the comment and the spellcast itself is over-indented.

killerwife: It wont be always nullptr, consuming is done at the end of spell::cast if I remember correctly. Problem is that during next spell execution, lets say when its with a hidden cast time, the effect of the next spell might not necessarily happen immediately, hence the m_castItem will be nullptr inevitably. (due to the nature of how spellEvent scheduling works)

Phatcat: However in this case it will always be nullptr since the item will never come back, wouldn't you agree?

killerwife: During the dummy, the item will still exist. TakeCastItem is called after handlers

Phatcat: Didn't you also agree just about an hour ago?

> Well, then you wouldnt wanna send the m_castItem along at all, because it will expire every time.

Phatcat: If the item still exists during the dummy, why does if(m_CastItem) fail?
because it's a nullptr...

killerwife: The previous mention was related to spellEvent queuing which will cause the pointer to be nullptr in the future spell. I didnt know if(m_CastItem) is failing.

Phatcat: > Then why won't the spell work if you do a check for m_CastItem with this dummy case then

-

> That does not explain why m_CastItem is a nullptr when the spell is very clearly cast from an item.

-

>  every other dummy effect with a CastItem has a check for the CastItem, and all dummy effects that come from items send along their CastItem; 

Kilerwife.... I simply don't believe that.

killerwife: Well I was criticizing that the spells wont work when cast alone, i.e. the CastItem checks at the start of dummies. Once again to get you a definitive answer I would have to launch it and confirm it myself. Last few items i dealt with, always had m_CastItem just fine in dummy effect, but then again, I am not saying items arent getting corrupted somewhere ^^

Phatcat: Actually I very much think that the spell will indeed work if cast alone; it just won't cast if the cast is conditioned on m_CastItem not being a nullptr, because it is in this case (according to the PR'er himself) - therefor we have to remove the check, and if the CastItem is anyways a nullptr, there is no reason to send it along.

CastItem works fine in most dummies because the item casting the spell usually doesn't get consumed on use? maybe?

src/game/SpellEffects.cpp
```diff
@@ -1149,6 +1149,13 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
 
                     return;
                 }
+                case 32027:                                 // Expedition Flare
+                {
+                        // 32029 = Expedition Preserver | 32030 = Expedition Scout
```
robinsch: coding style

Phatcat: What's wrong with the code style there? It's just a comment..

Phatcat: Oh, I see it now... Still over-indented.

Phatcat:
Alright, now all you gotta do is squash the commits into one
do a
`git rebase -i head~4`

and in the document that opens, change the 3 lower lines from `pick` into `f` or `s` - leave the top one as `pick`

robinsch:
You no longer need to squash commits, you can do it while you merge it into master.

Phatcat:
Yes, I very well know that I can; I could have done it several days ago; I want @Ccaz to get better with git, for his own sake.

killerwife:
This, ^^.

Phatcat:
@robinsch Also, don't encourge people to be lazy so that I end up with more work; I'm very grateful for your reviewing and that you take your time to do it, but forget your grudge against me; it's over.

Ccaz:
okay done, do I have to do another push after?

Phatcat:
yes, and force it.

Ccaz:
Just to be sure, I did a
git push --force https://github.com/Ccaz/mangos-tbc.git
But it isnt displayed here, so did I miss something? I can see that I squashed the commits into one but thats it.

Phatcat:
A Pull Request works in the way that it takes changes from a branch on your remote; https://github.com/Ccaz/mangos-tbc.git
You do not have the rights to push to cmangos at all, but a pull request updates itself automatically when you update a branch. the reason you do a push is to push your local changes to remote; you do not need to add the link, you can simply do a
`git push -f`
the reason you do a -f is because if you change a commit which you have already pushed to remote, or you remove a commit which exists on remote, when you try to push those changes, it will tell you that you have merge errors, for good reason; doing a -f tells git that you know what you're doing and that it should ignore any such concerns.

killerwife:
Also one thing to note, NEVER DO -f on main branch (i.e. master).

Phatcat:
You kind of had to here since you've PR'd from master, also something you want to avoid; by having your master free you always have a 'clean' branch; usually I just branch off of cmangos/classic, you can do that if you add cmangos to remote (it may already be, in which case you should be able to choose it from where you switch branches), but still I would never do a PR from any master. It just makes stuff complicated for yourself, and it's very easy to branch off actually.

Phatcat:
Now, there's actually one last thing to do; this is not needed at all, but if you do a
`git rebase cmangos/master`
and then force push it, then when I pull in your code it's going to show as a purple PR on your profile, as merged; otherwise I am going to have to pull in the commit in another way which results in the PR closing with a red state (not merged) - It's a very small thing; if you don't care about such petty formalities (personally, I don't), then it's ready to merge.

Ccaz:
I dont care either but If I try a git rebase cmangos/master, it says
fatal: Needed a single revision
invalid upstream cmangos/master

git remote -v
origin  http://github.com/cmangos/mangos-tbc.git (fetch)
origin  http://github.com/cmangos/mangos-tbc.git (push)

Phatcat:
going by that info there you should be able to do a `git rebase origin/master`

Phatcat:
The reason it's important to learn this either way is because this is how you update your repository with changes from the cmangos repository.

Phatcat:
Alternatively, you add it:
`git remote add cmangos https://github.com/cmangos/mangos-tbc.git`
`git fetch cmangos`

`git rebase cmangos/master`

That one cannot fail.

Ccaz:
Thanks that worked so far.
`git rebase cmangos/master
First, rewinding head to replay your work on top of it...
Applying: [ITEM] Expedition flare not working `update`

But If I try to do a git push -f now I get 

`fatal: The current branch cmangos/mangos-tbc has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin cmangos/mangos-tbc
`
I made a new brach coz of what Kilelrwife told me to never do a git push -f on the master

Phatcat:
> The current branch cmangos/mangos-tbc

you were not supposed to change branch.

killerwife:
@Phatcat he wasnt supposed to commit these things on top of Ccaz:master either ^^

@Ccaz always make a feature branch for a new thing, and one branch per pull request

Phatcat:
Okay, I'm gonna take it from here; first time is always the most confusing; all of this will make sense soon.

Ccaz:
That would be the best :D but thanks nevertheless

Phatcat:
Merged.

PR #76, opened 2016-07-12
Remove TBB

Lillecarl:
First PR in long time, don't know how PR's really work.

Anyways, this removes a dependency that we're in no use of.


Phatcat:
This is... this is huge...
If we can remove almost 20k lines of code and not introduce any regressions, that would be insanely awesome. I vote yes, yes, a thousand times yes on this one! :+1:

Phatcat:
Seroiusly? No-one? Well then I guess no-one minds me merging this in... let's say... 72 hours? That's like half a week, making it more than 10 days this PR has been up.

Am I familiar with TBB? Nope. and now I never will be since Carl has done such a fine job of removing it.

robinsch:
Seems like it's just a unused dependency...

Phatcat:
All the more reason to just get rid of it.

cala:
> Seroiusly? No-one? Well then I guess no-one minds me merging this in... let's say... 72 hours? That's like half a week, making it more than 10 days this PR has been up.

Remember that the summer period is either:
- holidays for many
- rush at work because many are in holidays and yet the company still has to deliver

So no replies for days are to be expected. :sunny: 

Though it indeed seems that TBB can be removed, I think that for this one, it would be better to have @cyberium or @namreeb opinion before merging. @namreeb did not remove it when he dropped ACE recently, maybe he did not have the time or there maybe also be a good reason for TBB to still be in.

So let's wait for their opinion before, there is no hurry there. And by the way, glad to see a new contribution from @Lillecarl :+1:

cyberium:
We are ok to remove it, @namreeb already proposed that https://github.com/cmangos/mangos-wotlk/pull/194
TBB did not worth the 20k lines of code needed for it.
This may have some performence impact on linux binaries but there is not any proof.

Lillecarl:
@cyberium i highly doubt that tbb malloc from many years ago is better then
last standard implementation?

killerwife:
I would ask @Laizerox since we experimented with it on linux ^^

Phatcat:
I am sorry for seeming a bit impatient, but considering all things, if noone reacts, the chance for something like this to go stale is risiculously high.

The chances for anything going stale here is a bit too high in my opinion. We have prs that have been open for years... not with several thousand lines of changes; some being rather small and obvious.
I wish to put an end to that.

Also notice how the commenting started after I threatened with merging blindly. And now we're actually talking about what the removal could impact, which is what I truly wished to accomplish with my comment.

@cala The formulation was chosen carefully to provoke, sorry if it offended anyone. Nothing but love for the project and my fellow devs <3

Ambal:
Initially TBB was added purely in order to use Intel's memory allocator which was much superior to default ones in Windows' and some Linux builds. Servers running on Windows XP saw ~30% reduction in CPU usage while some Linux boxes saw ~10% reduction. So only Windows builds benefited the most from TBB allocator. So feel free to remove TBB code base but consider leaving possibility to compile Mangos with TBB if someone would like to do it.

namreeb:
Are we sure that tbb is unused on UNIX?  I haven't checked that myself.  My PR for the wotlk repository removes it from the Visual Studio build files because it is unused there.

I think that we do need to consider using something like jemalloc, however, to address the problem of memory fragmentation for realms with high uptime and high population.

Lillecarl:
It's highly unused on Linux and isn't compatible with freebsd.
I've never seen or used it on Linux other then for testing.

I'll get back when I've read more about other allocators.

Lillecarl:
http://www.nedprod.com/programs/portable/nedmalloc/

According to this allocator it's of no use to use a third party memory allocator, as in... This is not where we're gonna earn lots, better to not depend on more code..?

"(Windows 7, Linux 3.x, FreeBSD 8, Mac OS X 10.6 all contain state-of-the-art allocators and no third party allocator is likely to significantly improve on them in real world results)."

I don't know any distro running Linux 2.x still IIRC.

I value not depending on things that become obsoleted (our TBB is old AF now).

Maintaining as little as possible is valuable too! :)

cyberium:
merged for wotlk, need backport. -> https://github.com/cmangos/mangos-wotlk/commit/5b0f690dcf91a61a967acf772578619ccff9b749

Lillecarl:
Wooooo

cala:
Backported to TBC by https://github.com/cmangos/mangos-tbc/commit/b08f3bdba95610293dd94ea51f5cdecc748a3759 and https://github.com/cmangos/mangos-tbc/commit/a957111b1fb85520b639c55d4c8080b12b015497
Pending backport to Classic.

cyberium:
fully backported

PR #77, opened 2016-07-15
Fixed warning for string to char* conversion

dvdvideo1234:
ObjectMgr.cpp:6929:39: warning: deprecated conversion from string constant to ‘char*’ [-Wwrite-strings]


cyberium:
implemented

PR #64, opened 2016-06-20
Plagued lands

killerwife:
Quest 2118 fixup and some new core changes.

Required SQL for changes
DELETE FROM spell_script_target WHERE entry=9439;
INSERT INTO spell_script_target VALUES(9439, 1, 2164, 0);

UPDATE quest_template SET ReqSpellCast1=0 WHERE entry=2118;
UPDATE gameobject_template SET data2=10, data4=0 WHERE entry=109515;

DELETE FROM spell_dbc WHERE Id=9439; -- change to spell_template in open source for now
INSERT INTO spell_dbc(Id, SpellName, Effect1, EffectImplicitTargetA1, EffectBasePoints1, rangeIndex, SpellIconID, EquippedItemClass) VALUES
(9439, 'Bear Captured in Trap', 3, 38, 1, 12, 1, -1);

UPDATE creature_template SET FactionAlliance=35,FactionHorde=35 WHERE entry=11836;

EDIT: @cyberium @xfurry @cala

PR #84, opened 2016-07-22
Fix Execute rage draining on miss

Warlockbugs:
Execute should consume all remaining rage only on successful hits.

### DIFF
--- a/src/game/Pet.cpp
+++ b/src/game/Pet.cpp
@@ -201,6 +201,9 @@ bool Pet::LoadPetFromDB(Player* owner, uint32 petentry, uint32 petnumber, bool c
     SetUInt32Value(UNIT_NPC_FLAGS, UNIT_NPC_FLAG_NONE);
     SetName(fields[11].GetString());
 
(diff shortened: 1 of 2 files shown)


robinsch:
I would rather change m_powerCost instead of calling ModifyPower because then it should be handled correctly by TakePower

cyberium:
Well after some reflexion maybe this will be better with -> https://github.com/cmangos/mangos-tbc/pull/64 to avoid the need of adding another hardcoded value.
Then we'll just have to use returned value of CastCustomSpell to apply or not the rage consumming.

Warlockbugs:
Sure, it can wait.

Warlockbugs:
I actually don't think it's needed anymore with:
https://github.com/cmangos/mangos-tbc/commit/52fe9f6e7904e32c08b7bc833ccf2b6ec284aaa3
https://github.com/cmangos/mangos-tbc/commit/9de117d3707358f20e1d3c0918c4ca67af9f3102
We roll hit for Execute once (in the parent spell) and refund parent spell's cost if needed. When Execute child spell is processed it's already guaranteed to land.

PR #85, opened 2016-07-22
Fix Health Funnel's spectacular "fake death" bug

Warlockbugs:
Interrupt channeled spells with periodic power cost if caster has insufficient power to keep channeling

### DIFF
--- a/src/game/SpellAuras.cpp
+++ b/src/game/SpellAuras.cpp
@@ -7440,10 +7440,16 @@ void SpellAuraHolder::Update(uint32 diff)
 
                 if (manaPerSecond)
                 {
-                    if (powertype == POWER_HEALTH)
-                        caster->ModifyHealth(-manaPerSecond);
+                    bool insufficient = (powertype == POWER_HEALTH) ? (caster->GetHealth() < (manaPerSecond + 1)) : (caster->GetPower(powertype) < manaPerSecond);
+                    if (insufficient && IsChanneledSpell(m_spellProto))
+                        caster->InterruptSpell(CURRENT_CHANNELED_SPELL);
                     else
-                        caster->ModifyPower(powertype, -manaPerSecond);
+                    {
+                        if (powertype == POWER_HEALTH)
+                            caster->ModifyHealth(-manaPerSecond);
+                        else
+                            caster->ModifyPower(powertype, -manaPerSecond);
+                    }
                 }
             }
         }


robinsch:
I would rather use some general check from CheckCast (CheckPower?), it should also return a error when there is not enough power (or health) to cast.

cyberium:
This will not 'Fix' that issue it will just improve it. It will reduce drasticaly the frequence of that issue.
There is still the possibility to get 'GetPower()' exactly equal to 'manapersecond' resulting to set health to zero resulting of death animation folowed by resurecting due to health regen.
So is this spell is supposed to realy kill its caster? If yes we have to set death state too. If not we could use manapersecond + 1 to let at least 1 HP.

killerwife:
Afaik stops at 1

Warlockbugs:
It's supposed to stop when there is not enough health to transfer. Updated condition to cost+1.

robinsch:
All of those things should be checked by CheckPower. You just need to call those checks or CheckCast at aura update. This should also fix other behaviours such as channeleing spells while out of range and so on.

cyberium:
@robinsch CheckCast is not adapted for that, i mean too much uneeded check for this case. CheckPower well maybe but its a spell method and not static.
So proposed solution is acceptable. Maybe we'll have to add distence check later but even that iam not sure. I remeber such extra far cast from spell like arcane missiles
https://www.youtube.com/watch?v=iNYswLPteqo

cyberium:
Well after more reviewing i conclude this is not the place to fix that issue,
Could you please test this patch -> https://github.com/cyberium/mangos-tbc/commit/8b7614a00d48f10e6d7a20564fb3598dffc9f632 and make your remark?

Warlockbugs:
[8b7614a] (https://github.com/cyberium/mangos-tbc/commit/8b7614a00d48f10e6d7a20564fb3598dffc9f632) solves this (without the periodic cost mod though).

cyberium:
ok remaining one thing, i encoutered cast of this spell that doing nothing due to pet have full life.
Is it supposed to be : 
- canceled
- do anyway the damage part
- let it cast and do nothing

Warlockbugs:
I'd go with canceling. According to [WotLK era video] (https://www.youtube.com/watch?v=JYyHTlb2Lpw) they can't just keep it channeling more than 1 seconds (aka cancel before tick if the pet health is full).

evil-at-wow:
I'd go with cancel as well. I can't say for sure what happened, but nowadays you can't even cast Health Funnel if your demon is full health (you get an error saying so). I don't remember what happened when your demon got back to full health because of the cast though, so I can't say for sure if it stopped channeling or just continued without much effect.

PR #86, opened 2016-07-24
Fix MSG_INSPECT_HONOR_STATS response

Warlockbugs:
The single byte parameter is actually the player's highest vanilla PvP rank index.
The range of valid values is 0-18.

I am not a mangos developer or a reverse engineer myself, so this discovery was made by debugging [Examiner] (http://legacy.curseforge.com/media/files/78/477/Examiner08.05.03.zip) addon, which is malfunctioning on all existing TBC servers.
It used to throw Lua errors on seemingly random inspects when calling UnitPVPRank function with the 6th returned value of GetInspectHonorData function.
A quick test confirmed that contents of the byte parameter are getting passed exactly as the 6th returned value.
The small size of the data (1 byte) confirms the claim by implication.

### DIFF
--- a/src/game/MiscHandler.cpp
+++ b/src/game/MiscHandler.cpp
@@ -1093,7 +1093,7 @@ void WorldSession::HandleInspectHonorStatsOpcode(WorldPacket& recv_data)
 
     WorldPacket data(MSG_INSPECT_HONOR_STATS, 8 + 1 + 4 * 4);
     data << player->GetObjectGuid();
-    data << uint8(player->GetUInt32Value(PLAYER_FIELD_HONOR_CURRENCY));
+    data << uint8(player->GetHighestPvPRankIndex());
     data << uint32(player->GetUInt32Value(PLAYER_FIELD_KILLS));
     data << uint32(player->GetUInt32Value(PLAYER_FIELD_TODAY_CONTRIBUTION));
     data << uint32(player->GetUInt32Value(PLAYER_FIELD_YESTERDAY_CONTRIBUTION));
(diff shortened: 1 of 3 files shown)


xfurry:
Poke @killerwife 
The update looks reasonable to me. Not sure if this is only for TBC or for newer versions as well.

Warlockbugs:
Yes, this also applies to WotLK and onwards, since the packet structure still features a byte.

[3.3.5 version](http://legacy.curseforge.com/media/files/445/178/Examiner10.08.14.zip) of the Examiner still contains the same code.

killerwife:
Actually @Warlockbugs is a vengeance player, so that was the original reason for him to make this patch. I guess that answers your question @xfurry

xfurry:
Then I think we can go ahead and push this.
I noticed that @Warlockbugs code usually looks pretty good, so if the community confirm that the changes are correct, I see no reason for waiting.

Right now I don't have the possibility to accept this, but I think so other developer can take over the task and accept the PR.

cyberium:
i dont understaind what you mean by "fixing the packet structure" here.
I dont see any change to the structure itself.
```diff
-    data << uint8(player->GetUInt32Value(PLAYER_FIELD_HONOR_CURRENCY));
+    data << uint8(OldLifetimeHighestPvPRankIndex);
```
You basicaly just fix the server response by providing the correct highest rank for players.
It should be reasonable to add a function tp player class for that.
uint8 Player::GetHighestPvpRank(); or such.

Warlockbugs:
My bad, sorry for using incorrect terminology here. 

Amended as the method of player class.

cyberium:
Merged thanks

PR #89, opened 2016-07-25
Fix Arcane Torrent mana replenishing

Warlockbugs:
Someone got wrong spell ids. >.<

### DIFF
--- a/src/game/Pet.cpp
+++ b/src/game/Pet.cpp
@@ -201,6 +201,9 @@ bool Pet::LoadPetFromDB(Player* owner, uint32 petentry, uint32 petnumber, bool c
     SetUInt32Value(UNIT_NPC_FLAGS, UNIT_NPC_FLAG_NONE);
     SetName(fields[11].GetString());
 
+    SetByteValue(UNIT_FIELD_BYTES_2, 1, UNIT_BYTE2_FLAG_UNK3 | UNIT_BYTE2_FLAG_AURAS | UNIT_BYTE2_FLAG_UNK5);
+    SetUInt32Value(UNIT_FIELD_FLAGS, UNIT_FLAG_PVP_ATTACKABLE);
+
     if (getPetType() == HUNTER_PET)
     {
         // loyalty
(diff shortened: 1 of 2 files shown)


cyberium:
Are you sure this is correct?
Two same spells except effect1 SPELL_EFFECT_ENERGIZE vs SPELL_EFFECT_ENERGIZE_PCT

Warlockbugs:
[SpellEffects] (https://github.com/cmangos/mangos-tbc/blob/master/src/game/SpellEffects.cpp#L1076)  contains script for 28730 and 28733.

cyberium:
Ok, any clue of from where those value come then? Wotlk? Classic?

Warlockbugs:
I am not sure exactly, my guess, one pair is meant for rogues and the other one is for mana using classes (all other blood elves). Its all got mixed and messed up. If it's true, additional fix is needed to make sure both versions correctly use dummy stacks and not replenishing mana "for free" without it (like it is atm).

Warlockbugs:
Re-checked on fresh branch: nevermind the pr, seems to be working as intended atm. Apparently, something was broken on my side.

killerwife:
@Warlockbugs its not :D

cyberium:
Why its not would be a better answer ...

killerwife:
Arcane torrent is implemented twice by mistake. Once in aura mod silence, Once in effect dummy. The correct way is for it to be implemented in Effect Dummy. Also, 28734 shouldnt restore mana, only the other one should,
 if (m_spellInfo->Id == 28734)
        return;
is missing from EffectPowerDrain

killerwife:
@cyberium I immediately started writing that message, ^^ didnt know u both were here at the same moment.

cyberium:
iam everywere at any moment looking over your shoulder.

Warlockbugs:
Ok, retested the whole combination:
Arcane Torrent does not reward mana if there is no dummy stacks from Mana Tap.
Mana Tap does reward mana on use instead of just draining it from enemy and applying dummy stack (this shouldn't happen).

killerwife:
line 1076 in SpellEffects.cpp

Warlockbugs:
Closing this with intention to open a new one with a solution for Mana Tap. Arcane Torrent somehow works at the moment in fact, no matter a proper way or not.

killerwife:
https://gist.github.com/killerwife/a08f1d1256f96c4f8d4d4dbdb70cd054

PR this - profit

robinsch:
meh

PR #90, opened 2016-07-26
Fix Improved Healthstone talent for Soulwells

Warlockbugs:
Summons improved Soulwells instead of regular one if caster has the talent.

This may be a bit hacky, but i couldn't find a prettier way to achieve the same result.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -7220,6 +7220,18 @@ void Spell::EffectTransmitted(SpellEffectIndex eff_idx)
 {
     uint32 name_id = m_spellInfo->EffectMiscValue[eff_idx];
 
+    switch (m_spellInfo->Id)
+    {
+        case 29886: // Create Soulwell
+            if (m_caster->HasAura(18692))
+                name_id = 183510;
+            else if (m_caster->HasAura(18693))
+                name_id = 183511;
+            break;
+        default:
+            break;
+    }
+
     GameObjectInfo const* goinfo = ObjectMgr::GetGameObjectInfo(name_id);
 
     if (!goinfo)


src/game/SpellEffects.cpp
```diff
@@ -7220,6 +7220,14 @@ void Spell::EffectTransmitted(SpellEffectIndex eff_idx)
 {
     uint32 name_id = m_spellInfo->EffectMiscValue[eff_idx];
 
+    // Improved Healthstone Talent Support for Soulwell
+    if (m_caster && name_id == 181621) {
```
robinsch: Codestyle {

Warlockbugs:
Aye, an old habit for Qt style blocks. Fixed.

xfurry:
Hmm... quite hack.
Let's see if we can find a way around this and use only spell ids, preferably.

robinsch:
I'm not really sure, I mean when player logs off you can no longer get improved Stones.

robinsch:
@xfurry What other way? The spell should summon a diffrent gameobject when having a certain aura (talent). I have no idea how to handle this otherwise.

killerwife:
the only other way is to edit SpellInfo during runtime for specific player, so that the GO summoned in effect is the one of choice, but since we store spellinfo globally thats just not possible
it is very likely we "overload" stuff in a wrong way, kinda made apparent with the whole "overload class scripts" effects,

robinsch:
Which will basicly result in the same check?

killerwife:
Unless you propose a runtime solution, which on apply of talent aura, you would change a value in the memory for given player which would edit GO summoned for this particular spell, you will end up with some flavor of the same check anyway. I will try to pull an idea out of my arse in the future for this, but for now, this PR is legit.

xfurry:
There is no "normal" way. But something like this looks better.
Instead of checking for the GO entry, we check for the spell entry. Makes more sense.

```diff
@@ -11416,10 +11416,23 @@ void Spell::EffectModifyThreatPercent(SpellEffectIndex /*eff_idx*/)
 
 void Spell::EffectTransmitted(SpellEffectIndex eff_idx)
 {
     uint32 name_id = m_spellInfo->EffectMiscValue[eff_idx];
 
+    switch (m_spellInfo->Id)
+    {
+        case 29886:
+            if (m_caster->HasAura(18692))
+                name_id = 183510;
+            else if (m_caster->HasAura(18693))
+                name_id = 183511;
+            break;
+        case 58889:
+            // ToDo: something with GOs 193170 and 193171
+            break;
+    }
+
     GameObjectInfo const* goinfo = ObjectMgr::GetGameObjectInfo(name_id);
 
     if (!goinfo)
     {
         sLog.outErrorDb("Gameobject (Entry: %u) not exist and not created at spell (ID: %u) cast", name_id, m_spellInfo->Id);

```

Warlockbugs:
Updated to switch.

xfurry:
Poke @killerwife about the ToDo: note from my patch. Any idea how would that work?

robinsch:
You could also move this to gameobject scripts.

killerwife:
@xfurry same rules for that TODO, the GOs are ordered by numbers for the talents

Also, this will be easily made more nice once we add my spell_scripting implementation. It also gave me an idea how to generically pass parameters from my overloading ^^

PR #91, opened 2016-07-26
Update client control state when on taxi

Warlockbugs:
A server-client communication enhancement.

As evident from the source of [QuestHelper 0.50] (http://legacy.curseforge.com/media/files/88/684/QuestHelper-0.50.zip) for TBC, client expects a control update when taking/leaving a taxi flight path.

```lua
  if event == "PLAYER_CONTROL_GAINED" then
    self:flightEnded()
  end

  if event == "PLAYER_CONTROL_LOST" then
    self:flightBegan()
  end
```
This patch fixes flight path timer display for QuestHelper and possibly other addons.


Dramacydal:
:+1:

PR #92, opened 2016-07-26
Fix Pyroclasm talent

Warlockbugs:
Adjust procFlags for all affected spells, include Soul Fire in core conditions.


robinsch:
Just remove the else return line. All cases are should be covered by spell_proc_entry

Warlockbugs:
Updated.

cyberium:
is this needed for wotlk?

Warlockbugs:
No, the talent was entirely changed in WotLK to provide bonus damage on crit.

Phatcat:
> -                else
-                    return SPELL_AURA_PROC_FAILED;

Those 2 lines should stay in wrath then? (if they currently reside there at all) @robinsch ?

Also what is the issue with Pyroclasm currently?

Warlockbugs:
If you read description of the talent in wrath, it has nothing to do with stuns anymore. It was revamped. If any stun related code exists in wrath branch, it should be removed.

Phatcat:
Actually not looked into this in detail yet, just gathering as much information as I feel I maybe need before I do, so I won't have to sit around waiting for answers which I could have gotten beforehand had I just prepared myself a little. Sorry if some of my questions are very basic at times.

PR #93, opened 2016-07-26
Fix Aftermath talent

Warlockbugs:
Prevents warlocks from dazing themselves with proc on self-damage when using Hellfire.

Is there a better way to do this? I found this one requiring the least amount of extra code involved.


killerwife:
you might as well fix self stun from warlock talent as well ^^

robinsch:
There are tons of other spells that should not proc on self - iirc you can create a general check because !PositiveSpells should never proc on self.

killerwife:
Yes but aftermath and blackout are not positive spells afaik

robinsch:
That's what I've said?

killerwife:
Jesus, there is and exclamation mark, my bad. Oh and also, the SPELL_ATTR_EX_NEGATIVE in current code is wrong, it actually means "cannot be reflected". Negative flag is a different one.

Warlockbugs:
Second Attempt. Now made with a generic check, quick test is successful, but i quickly became "dangerous". I'll look for cases where it fails.

robinsch:
Have you tried to debug it to see why target is caster?

Warlockbugs:
It procs because of damage done to caster by Hellfire and we do not have extra flags to prevent procs on self-damage.

robinsch:
Then you should check it there - I doub there is a spell that should proc on self when caster == victim.

killerwife:
@robinsch might be onto something, same like triggered spells dont proc unless specified (not yet implemented tho)

Warlockbugs:
Updated, now targeting specific case. Still feels like something is missing.

cyberium:
This sound reasonable however iam not sure if what it can broke.
Could you please add an explicit  detail_log before the return and an explicit TODO to remove it later?

Warlockbugs:
Updated with output and rebased.

Confirmed to cure:
* Warlocks dazing themselves with Aftermath proc or stunning with Pyroclasm on self-damage from Hellfire
* Shadow Priests applying Shadow Vulnerability to themselves with Shadow Word: Death

Warlockbugs:
Currently i am doing an experimental partial positive/negative effect handling rewrite with intention to "unhack" it, and i was able to fix the issue in the epicenter of the spell system. Just noting that this PR as partially deprecated, just in case if someone decides to merge it. I'll close it when the better solution will be fully ready and tested.

PR #94, opened 2016-07-27
Fix periodic leech effects threat generation

Warlockbugs:
Periodic Health leech should not generate healing aggro (only damage).
Sources: [DTM](http://legacy.curseforge.com/media/files/201/513/DiamondThreatMeter_1.7.0_hybrid.zip), [KTM](http://legacy.curseforge.com/media/files/87/88/KLHThreatMeter.22.1.zip).


Warlockbugs:
After more extensive research, i came to conclusion that mana leech part should be decided on per spell basis, because some spells with that effect still supposed to do threat, so the second part of the change is reverted.

PR #95, opened 2016-07-27
Fix Demonic Sacrifice

Warlockbugs:
Demonic Sacrifice should not put caster in short combat with his own (dead) pet.
This may slow down buffing in groups significantly, because buffers also receive that pointless short combat state in a chain reaction.

I admit, this is a rather hackish one, but the context is funny: we make pet suicide here instead of killing it ourselves. :)
Also, from performance standpoint we don't introduce additional code or branching anywhere and reusing existing one.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -286,7 +286,7 @@ void Spell::EffectInstaKill(SpellEffectIndex /*eff_idx*/)
     data << uint32(m_spellInfo->Id);
     m_caster->SendMessageToSet(&data, true);
 
-    m_caster->DealDamage(unitTarget, unitTarget->GetHealth(), nullptr, DIRECT_DAMAGE, SPELL_SCHOOL_MASK_NORMAL, nullptr, false);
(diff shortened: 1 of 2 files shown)


robinsch:
Spell has SPELL_ATTR_EX_NO_INITIAL_AGGRO in 3.0.x - is there a way to add attr to a spell in mangos? I would rather fix it this way.

killerwife:
There is on vengeance ^^ I just need to export it.

robinsch:
I don't really know why you guys are mentioning vengance all the time, does it even matter? Pretty annoying.

killerwife:
Izi, we actually contribute. And for me there is no point in writing "I have something on our repo for this" instead of "I have something on vengeance for this". Nonetheless, I will try to export it since @devmaximus  is too busy

Warlockbugs:
Another attempt. This one a bit less hacky and makes use of the said flag. None of the other major spells were affected by the change (most instakill effects are some NPC/testing ones).

PR #97, opened 2016-07-28
Blizzlike NPC crit adjustments

Warlockbugs:
NPCs can no longer crit with spells and abilities.
This part is often overlooked in many forks, resulting in NPCs being able to crit with abilities (and spells), which were modeled after ones available to players (and have similar damage parameters).

Source:
[WoWWiki] (https://web.archive.org/web/20160728101909/http://wowwiki.wikia.com/wiki/Spell_critical_strike?oldid=2104859)
>  Mobs are unable to get critical strikes with spells. 

[WoWHead] (https://web.archive.org/web/20160728090939/http://www.wowhead.com/forums&topic=17116/do-npcs-crit)

> NPC specials do NOT crit/crush now. The only thing that can crit/crush is their auto-attack.

This applies to all game versions.

### DIFF
--- a/src/game/Unit.cpp
+++ b/src/game/Unit.cpp
@@ -6089,6 +6089,14 @@ bool Unit::IsSpellCrit(Unit* pVictim, SpellEntry const* spellProto, SpellSchoolM
     if (spellProto->HasAttribute(SPELL_ATTR_EX2_CANT_CRIT))
         return false;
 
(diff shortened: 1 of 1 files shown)


robinsch:
iirc Totems can't crit.

Warlockbugs:
AFAIK they can, if the spell being casted allows it. Fire Nova, Magma if i am not mistaken.
The other way (without getting rid of that base 5% for consistency reasons) to solve it is to manually sort out which NPC spells should not crit and localize the problem.

PR #101, opened 2016-07-30
Fix Feral talent tree corruption

Warlockbugs:
Do not replace spells in spellbook when their Rank 1 is listed in the talent tree. The spell in question is Faerie Fire (Feral).

### DIFF
--- a/src/game/SpellMgr.cpp
+++ b/src/game/SpellMgr.cpp
@@ -1764,6 +1764,12 @@ bool SpellMgr::canStackSpellRanksInSpellBook(SpellEntry const* spellInfo) const
 {
     if (IsPassiveSpell(spellInfo))                          // ranked passive spell
         return false;
+    if (const SpellChainNode* node = GetSpellChainNode(spellInfo->Id))
+    {
+        // do not corrupt talent tree display by removing a rank from there, e.g. Faerie Fire (feral)
+        if (GetTalentSpellPos(node->first))
+            return true;
+    }
     if (spellInfo->powerType != POWER_MANA && spellInfo->powerType != POWER_HEALTH)
         return false;
     if (IsProfessionOrRidingSpell(spellInfo->Id))


evil-at-wow:
I remember the problem with Faerie Fire (Feral). I would say this is not the perfect fix, but probably acceptable. There are several other spells that will be affected by this patch (basically any spell not using mana or health that is part of a talent spell chain) and will now return true instead of false as before, but that seems to have no real consequences as far as I know.

Some info that might be useful in this area and that people might not remember or not know about yet was once shared by me on the forum - see the old threads [talent reset](http://cmangos.net/thread-334.html) and [Faerie Fire (Feral) Talent](http://cmangos.net/thread-342.html).

Warlockbugs:
The patch you linked was the first thing i thought of, but i came to conclusion it is a pure hack and quickly dismissed, given how people rated similar patches i created for other spells. So i tried to target the underlying issue instead to keep it with the philosophy of the project.

evil-at-wow:
Yes, I had the impression you know about the underlying issue, and I think you made an acceptable patch for that. But I tried to provide some extra context for people who might wonder why your change fixes that talent :smile:

A better alternative would probably be to not automatically return false for spells that aren't using mana or health (as I described on the forum), so basically removing the lines below your change, but who knows what consequences that would bring... so that would certainly need a lot more thought and testing.

cyberium:
is this have to be applied in other core too?
I consider it also as an acceptable solution i mean it make sense.

Warlockbugs:
No idea if the same issue exists in other cores. @Phatcat

robinsch:
I don't think this is still a thing on wotlk.

cala:
There is an issue in Classic where druids can get the feral faerie fire talent from the feral tree and keep it even after reseting their talents, having the spell for ever and for free. So, if this fix/issue is the same here, it is needed in Classic too.

Phatcat:
Sorry for not replying to this one even though I was tagged in it. Basically what @cala said, this is needed for classic. No idea about wrath. Can test it.

Phatcat:
Next time I say I can do something remind me to actually go do it -.-

I found no issues with primal fury on the wrath branch, neither learning it or resetting all talents (unlearning it)

Phatcat:
This is not an issue on the classic branch, you unlearn feral faerie fire when you unlearn your talents from the class trainer.

Phatcat:
Ehh... This is not even an issue on current TBC master.... is there even any corruption to fix anymore?

Phatcat:
oh wait, now I see it. sorry.

Warlockbugs:
Its still an issue in the current TBC master, when you relog your FF ranks are still removed from spellbook. Thing like that don't magically fix themselves :smile: 
So, what do we say then? Is it needed for all cores?

Phatcat:
I'm just gonna recheck once more for wrath since I obviously didn't check it right the first time for classic and tbc (while rechecking revealed the corruption on both branches)

Phatcat:
For the sake of having synced cores sure, but there actually is no need for it in wrath as there are no talents which are affected by this, at least not for druids. for classic and for tbc we need this yes.

PR #102, opened 2016-07-30
Fix Feral Swiftness talent

Warlockbugs:
Check for stances when re-applying passive outdoor-only spells.

Without this patch this talent would be active in any form or without forms.


kartoffelsup:
https://github.com/cmangos/issues/issues/434
Not sure which is better. Your approach or what @robinsch suggested in the issue:
` if (GetErrorAtShapeshiftedCast(spellInfo, GetShapeshiftForm()) != SPELL_CAST_OK)`

PR #105, opened 2016-08-04
Fix Health Leech-based spells being able to crit

Warlockbugs:
This fixes Death Coil being able to crit.

Sources: [DrDamage] (http://www.wowace.com/media/files/126/551/DrDamage-r81778.1.zip), [WoWWiki] (http://wowwiki.wikia.com/wiki/Death_Coil_(warlock_ability)?oldid=1597846)


robinsch:
I still would rather go for loading spell attributes from DB (not from DBCs)

Warlockbugs:
Unfortunately, we don't have spell.dbc DB or code double as of yet. :(

Lillecarl:
Looks good (Though as @robinsch said, this should be somewhere else in future..)

PR #81, opened 2016-07-18
Adjust threat generation for class abilities

Warlockbugs:
Nearly complete revamp for TBC. Values for abilities present in vanilla can be freely ported to classic as is, i think, because major re-balance happened only in WotLK.

PR #80, opened 2016-07-17
Adjust bonus power coefficients for class abilities and spells

Warlockbugs:
More realistic TBC coefficients for casters, fix for obvious mistakes (such as Rain of Fire), fix for hidden bug (Ignite talent should not receive bonus spell power) and cleanup (Living Bomb is WotLK spell).

PR #107, opened 2016-08-06
Cleanup spell chains, coefficients and threat

Warlockbugs:
Finalizing (hopefully) what was started in [80](https://github.com/cmangos/mangos-tbc/pull/80) & [81](https://github.com/cmangos/mangos-tbc/pull/81):
* Fixed lots of wrong spellids or separate effects, typos, rounding, etc
* Redundant rank spam cleaned up and inserted into spell_chain
* In addition, a subset of data was changed after a very close inspection and comparision to the values provided by sources:

DTM: [DiamondThreatMeter for TBC] (http://legacy.curseforge.com/media/files/201/513/DiamondThreatMeter_1.7.0_hybrid.zip)
EJ: [1] (https://web.archive.org/web/20080813085840/http://elitistjerks.com/f47/t19038-spell_coefficients/) and [2] (https://web.archive.org/web/20080813193306/http://elitistjerks.com/f15/t9363-damage_coefficients/)

New power coefficients
-
Chastise (EJ)
Judgement of Command (EJ)

Fine-tuned power coefficients (minor changes)
-
Insect Swarm to EJ
Blizzard to EJ
Flamestrike to EJ
Frost Nova to EJ
Blast Wave to EJ
Cone of Cold to EJ
Avenger's Shield to EJ
Seal of Righteousness to EJ
Holy Fire to EJ
Starshards to EJ
Devouring Plague to EJ
Shadowguard to EJ
Earth Shock to EJ
Frost Shock to EJ
New threat:
-
These spells will not cause threat:
Lifebloom Final Tick (DTM)
Drain Mana (DTM)


Lillecarl:
Looks good to me, though where did you get the data about the coefficeients from?

Warlockbugs:
I've quoted every source i used in the post.

Lillecarl:
@Warlockbugs Sorry i interpreted it wrong :)

Phatcat:
This looks very much like a db update, am I wrong? - Usually those are PR'd for the db in question, and in your case, I'm guessing it's TBC-DB: https://github.com/TBC-DB/Database

Also you've made the same mistake that I did here: https://github.com/classicdb/database/pull/828#issuecomment-210384280

Lillecarl:
@Phatcat some data are provided by mangos, for example spell data. Then all "db makers" take this mangos data and puts it into their own databases.

Warlockbugs:
These tables are marked as "supported by the core team", so, i assume, TBCDB syncs them with the core dump, not in reverse.

killerwife:
In a nutshell.

Phatcat:
Alright, sounds reasonable enough I guess.

Warlockbugs:
Fixed a couple of typos.

Warlockbugs:
Additional cleanup according to [warnings] (http://pastebin.com/qF0ZC9D2).

killerwife:
@Warlockbugs 
-(33763, 0.3429, 0.0742,  0,     0,     'Druid - Lifebloom'),
+(33763, 0,      0.0742,  0,     0,     'Druid - Lifebloom HoT'),
+(33778, 0.3429, 0,       0,     0,     'Druid - Lifebloom Final Tick'),

this broke Lifebloom dmg computation. I do not know why, but the last tick if you do this doesnt scale with SP. I think it has something to do with being a custom spell.

killerwife:
https://github.com/cmangos/mangos-tbc/blob/master/src/game/SpellAuras.cpp#L2508 @Warlockbugs 
For the future, some spells utilize empty dmg spells, shadow word:death, lifebloom last tick, which skip bonus computation. Hence it has to be done by parent spell, and then its passed down. Since you removed coeff from lifebloom dot, it doesn't compute correctly anymore.

Warlockbugs:
Crit chance does work separately though, so i (incorrectly) assumed this is true for SP too, which is why i separated them. It's getting more and more confusing, no consistency at all...

killerwife:
Imagine following scenario with the spell bonuses. You have 5 spells, all using same dummy spell for a special effect. All have different cast times/different parameters and each needs to be able to use that spell_bonus_data thingy. This way, you would be able to only set one, and not do anything else. Yes, it is indeed confusing. What about crit chance? Afaik crit isnt set anywhere, apart from the attribute CANT_CRIT.

Phatcat:
Merged.

PR #108, opened 2016-08-11
Apply Sunder Armor at Devastate.

Phatcat:
https://github.com/cmangos/issues/issues/1018#issuecomment-239075039

Signed-off-by: lovelol <tifxoxo@web.de>

---
This change is [Reviewable](https://reviewable.io/reviews/cmangos/mangos-tbc/108)

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -2982,9 +2982,12 @@ void Spell::cast(bool skipCheck)
     // CAST SPELL
     SendSpellCooldown();
 
-    TakePower();
-    TakeReagents();                                         // we must remove reagents before HandleEffects to allow place crafted item in same slot
-    TakeAmmo();
+    if (!m_IsTriggeredSpell)
+    {
+        TakePower();
+        TakeReagents();                                         // we must remove reagents before HandleEffects to allow place crafted item in same slot
+        TakeAmmo();
+    }
 
     SendCastResult(castResult);
     SendSpellGo();                                          // we must send smsg_spell_go packet before m_castItem delete in TakeCastItem()...
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -4800,22 +4800,38 @@ void Spell::EffectWeaponDmg(SpellEffectIndex eff_idx)
             // Devastate
             if (m_spellInfo->SpellVisual == 671 && m_spellInfo->SpellIconID == 1508)
             {
-                customBonusDamagePercentMod = true;
-                bonusDamagePercentMod = 0.0f;               // only applied if auras found
-
                 // Sunder Armor
                 Aura* sunder = unitTarget->GetAura(SPELL_AURA_MOD_RESISTANCE, SPELLFAMILY_WARRIOR, uint64(0x0000000000004000), m_caster->GetObjectGuid());
 
-                // Devastate bonus and sunder armor refresh, additional threat
+                // Devastate bonus and sunder armor refresh
                 if (sunder)
                 {
                     sunder->GetHolder()->RefreshHolder();
+                    spell_bonus += sunder->GetStackAmount() * CalculateDamage(EFFECT_INDEX_2, unitTarget);
+                }
 
-                    // 100% * stack
-                    bonusDamagePercentMod += 1.0f * sunder->GetStackAmount();
+                // Devastate causing Sunder Armor Effect
+                if (!sunder || sunder->GetStackAmount() < sunder->GetSpellProto()->StackAmount)
+                {
+                    // get highest rank of the Sunder Armor spell
+                    const PlayerSpellMap& sp_list = ((Player*)m_caster)->GetSpellMap();
+                    for (PlayerSpellMap::const_iterator itr = sp_list.begin(); itr != sp_list.end(); ++itr)
+                    {
+                        // only highest rank is shown in spell book, so simply check if shown in spell book
+                        if (!itr->second.active || itr->second.disabled || itr->second.state == PLAYERSPELL_REMOVED)
+                            continue;
 
-                    // 14 * stack
-                    unitTarget->AddThreat(m_caster, 14.0f * sunder->GetStackAmount(), false, GetSpellSchoolMask(m_spellInfo), m_spellInfo);
+                        SpellEntry const *spellInfo = sSpellStore.LookupEntry(itr->first);
+                        if (!spellInfo)
+                            continue;
+
+                        if (spellInfo->IsFitToFamily(SPELLFAMILY_WARRIOR, uint64(0x0000000000004000))
+                            && spellInfo->Id != m_spellInfo->Id)
+                        {
+                             m_caster->CastSpell(unitTarget, spellInfo->Id, true);
+                             break;
+                        }
+                    }
                 }
             }
             break;


robinsch:
wow, didn't know this account still exists from me

src/game/Spell.cpp
```diff
@@ -2982,9 +2982,12 @@ void Spell::cast(bool skipCheck)
     // CAST SPELL
     SendSpellCooldown();
 
-    TakePower();
-    TakeReagents();                                         // we must remove reagents before HandleEffects to allow place crafted item in same slot
-    TakeAmmo();
+    if (!m_IsTriggeredSpell)
```
killerwife: This has nothing to do with this patch. if (!m_IsTriggeredSpell) was added before CalculatePowerCost hence sunder will not cost mana.

Phatcat: Tell that to robin, then: https://github.com/lovelol/server/commit/c80a61854def9c0d08b59bf92b4c2f219755492c

robinsch: This change was done like 5 years ago.

Phatcat: Hey, it wasn't a diss in any way, but since you created this you may still have some idea how it fits together and how reliant the other changes are on that wrongful check there, or how valid the other changes are at all.

robinsch: I don't really rememeber but as @killerwife said triggered spells should cost no power hence is should belong to CalculatePowerCost -> will have the same result but in a better place.

Phatcat: But would that mean that the rest of the changes are still good and it's only the check that's invalid, or the whole patch that's invalid? (I havn't looked into this in detail myself, I just saw the open issue and was going through a repo, yours as it seems, and saw a commit which was supposed to fix the issue and it seemed somewhat legit so I thought I would PR it)

Phatcat: And now gone is that repo and those commits. Now why would you do that. Doesn't matter, I fetched it locally. Unless you're saying that none of that work was worth anything. Should I go ahead and revert the other commit I pulled from that repo?

killerwife:
Both of you, look at this https://gist.github.com/killerwife/9155003762ef260fc9f6de867274f2a1

robinsch:
Oh sorry, I merged those two account. I don't think there was anything worth it on there because those fixes are acient and most of them not done correctly. I think there is a commit about Devastate on NostTBC which is way cleaner.

src/game/Spell.cpp
robinsch: Oh sorry, I merged those two account. I don't think there was anything worth it on there because those fixes are acient and most of them not done correctly. I think there is a commit about Devastate on NostTBC which is way cleaner.

Phatcat: Actually, I was talking about this: https://github.com/cmangos/mangos-tbc/commit/47ee89b33291d37b6dee9fab7f7a51f07821a3b1

robinsch:
@killerwife No need to refresh aura holder at 5 stacks because that's what casting a spell with stacks on max stack does.

src/game/Spell.cpp
robinsch: Yes, this is legit for PR. In TBC Avenger's Shield was the only spell that should not hit controlled units, this attr has been confirmed on WOTLK branch where many spells got added to this behaviour.

Phatcat:
> unitTarget->AddThreat(m_caster, 14.0f * sunder->GetStackAmount(), false, GetSpellSchoolMask(m_spellInfo), m_spellInfo);

Is that manual threat adding?

I liked this solution because it doesn't seem to have manual threat adding.

robinsch:
Threat should scale with Stacks, so this version doesn't work.

Phatcat:
the kind of stuff that is being done in current master and in killerwife's solution is the kind of stuff that belongs in a dummy effect if anywhere - not here. It just seems so wrong to be adding threat here.

killerwife:
@robinsch look at conditions, we dont cast spell on max stacks

robinsch:
@killerwife Oh my bad, what's the point of it?

Phatcat:
Spell::EffectWeaponDmg

.... Why on earth would we choose to handle that spell here at all? Is this really our only option?

killerwife:
Well sunder armor doesnt deal any damage, only applies stacks. Hence it is a fairly heavyweight way to go about things. If we just call refreshholder we can skip both looking for highest rank of sunder AND casting sunder, which is quite nice. Especially when mostly during boss combat, ur like all the time on max stacks.

And @Phatcat look at spellproto of devastate and then ask.

robinsch:
@killerwife seems fine, what happens when somebody with a lower ranked Sunder Armor tries to apply it to the target. (I have no idea if SPELL_AURA_BOUNCED is implemented on mangos) Refreshing the higher ranked aura seems not to be right, however I can't provide a proof for my claim.

killerwife:
Well, the idea is that there are special rules for overwriting stuff. Even though I know we havent got it implemented in cmangos ofc.

Sure, you have a point, but I have no clue how we would find out. And until we have a way, I usually go for the "lets not make our life complicated". We do not even know if this spellbook searching for sunder rank is good.

Warlockbugs:
Just don't forget about threat value (quoting DTM):
```
function(sourcePtr, targetPtr, abilityRank) 
    local value = 120;
    local there, rank, count, timeLeft = DTM_Unit_SearchEffect(targetPtr, "SUNDER_ARMOR");
    if ( there and count ) and ( count >= 1 ) then
		value = 120 + (count-1)*15;
		if ( sourcePtr and targetPtr ) then
			-- DTM_Trace("THREAT_EVENT", "[%s] used Devastate on [%s]. Sunder armor stacked %d time(s) => Base threat value: %d", 1, UnitName(sourcePtr), UnitName(targetPtr), count, value); 
        end
    end
    return value;
end
```

killerwife:
@Warlockbugs look at my gist

robinsch:
@killerwife seems stupid to me, but hey there are so many cases where we iterate through aura where it doesn't make sense at all, so your solution seems fine to me.

killerwife:
As of now, I do not have a better solution, the only other one is to hardcode sunder rank to devastate rank, and that one I like less. Also, notice how Devastate doesnt have EFFECT1 and sunder is EFFECT1 only. I suspect blizz did some neat overlap shenanigans.

robinsch:
Other way round.

killerwife:
Theoretically, I could alter spell_dbc and add the missing EFFECT1 to devastate, but thats completely meh tbh.

robinsch:
I have no clue what you want to add there?

robinsch:
I can't follow you, you mean add Sunder Armor Effect0 to Devastate Effect0? That would not even work out.

killerwife:
Now that I think about it yeah, would be wrong Icon.

robinsch:
I mean to skip that itr you have to add a value that changes at learning spells for example, that's what I did to no longer iter at MOD_STALKED auras.

killerwife:
Shoot a reference or smth. Doesnt sound bad, but depends how its done.

robinsch:
I mean for example in this case. In player class something like 
std::unordered_map<uint32, uint32> maxRankTriggerStore; where key 1 is Devastate and key 2 is max rank sunder, which will be changed at spell learn function.

killerwife:
Well then, I suggest that either we commit our version right now, and then when you have time to do your store thingy, we eliminate all of these in the core, OR we merge this after that trigger store is implemented. I myself am currently busy with some scheduling.

robinsch:
Your solution looks fine to me, also this would make zero performance impact.

killerwife:
Well then, I will try to make it into a PR today.

Phatcat:
Make a comment refering to this one when you do and I'll close this.

PR #109, opened 2016-08-11
Implement correct devastate

killerwife:
Adds correct sunder application and fixed bonus damage addition

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -4793,6 +4793,7 @@ void Spell::EffectWeaponDmg(SpellEffectIndex eff_idx)
     float totalDamagePercentMod  = 1.0f;                    // applied to final bonus+weapon damage
     bool normalized = false;
 
+    int32 fixed_bonus = 0;
     int32 spell_bonus = 0;                                  // bonus specific for spell
 
     switch (m_spellInfo->SpellFamilyName)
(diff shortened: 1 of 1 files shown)


killerwife:
https://github.com/cmangos/mangos-tbc/pull/108 improvement
@Phatcat @robinsch

src/game/SpellEffects.cpp
```diff
@@ -4800,19 +4801,50 @@ void Spell::EffectWeaponDmg(SpellEffectIndex eff_idx)
...
                 // Devastate bonus and sunder armor refresh, additional threat
                 if (sunder)
                 {
-                    sunder->GetHolder()->RefreshHolder();
+                    if (sunder->GetStackAmount() == 5)
+                        sunder->GetHolder()->RefreshHolder();
 
-                    // 100% * stack
-                    bonusDamagePercentMod += 1.0f * sunder->GetStackAmount();
+                    for (int j = 0; j < MAX_EFFECT_INDEX; j++)
```
robinsch: isn't ++j the correct syntax for it

killerwife: for primitive types it doesn't matter at all, but sure, I can change that after lunch

robinsch: I don't know how coding standarts are at cmangos, but basicly I would use a uint8 (even tho it's completly unrelated)

killerwife: later down we use int for this as well

Laizerox: http://stackoverflow.com/a/4706225

In this case it doesn't really matter it's pre or post :)

robinsch: Wasn't about pre- or postfix, if in most cases for iter SpellEffectIndex ++i is used, then I would prefer to use it for coding style.

Phatcat:
http://wowwiki.wikia.com/wiki/Devastate

So, it looks like this skill changed quite a bit between the expansion; what exact differences do we have to take into account server-side?

src/game/SpellEffects.cpp
```diff
@@ -4797,19 +4798,50 @@ void Spell::EffectWeaponDmg(SpellEffectIndex eff_idx)
...
+                    // Rescan sunder armor to apply all bonus effects if aura was not present.
+                    if (!sunder)
+                        sunder = unitTarget->GetAura(SPELL_AURA_MOD_RESISTANCE, SPELLFAMILY_WARRIOR, uint64(0x0000000000004000), m_caster->GetObjectGuid());
+                }
+
                 // Devastate bonus and sunder armor refresh, additional threat
                 if (sunder)
                 {
-                    sunder->GetHolder()->RefreshHolder();
+                    if (sunder->GetStackAmount() == 5)
```
Phatcat: since `sunder->GetStackAmount()` is called more than once maybe it should be stored in a variable at the beginning of this whole scope?

killerwife: no, because it needs to be called again in case sunder was called in previous if

killerwife:
This patch is TBC only, wotlk has already proper implementation since sunder armor is % in wotlk not flat +.

src/game/SpellEffects.cpp
robinsch: == maxstackamount

src/game/SpellEffects.cpp
```diff
@@ -4797,19 +4798,50 @@ void Spell::EffectWeaponDmg(SpellEffectIndex eff_idx)
...
-
                 // Sunder Armor
                 Aura* sunder = unitTarget->GetAura(SPELL_AURA_MOD_RESISTANCE, SPELLFAMILY_WARRIOR, uint64(0x0000000000004000), m_caster->GetObjectGuid());
 
+                // apply sunder armor first
+                if (!sunder || sunder && sunder->GetStackAmount() < 5)
+                {
+                    // get highest rank of the sunder armor spell
+                    const PlayerSpellMap& sp_list = ((Player*)m_caster)->GetSpellMap();
+                    for (PlayerSpellMap::const_iterator itr = sp_list.begin(); itr != sp_list.end(); ++itr)
```
robinsch: I would rather store highest Sunder Armor somewhere instead of itr on every cast.

Phatcat:
TBC only, so devastate is working as it should on classic?

Warlockbugs:
There is no Devastate in classic. This is a TBC+ spell.

src/game/SpellEffects.cpp
killerwife: Later down the line I might. For now it has 0 performance impact.

src/game/SpellEffects.cpp
killerwife: Yes, i will fix that immedaitely.

Phatcat:
I guess it's this:
> Bc icon Patch 2.0.3 (09-Jan-2007): The bonus threat of "Devastate" on all ranks has been increased. Bonus threat will no longer decrease as ranks increase.

throwing me off.

I am gonna get a move on this one soon unless anyone has a problem with the design or some improvements perhaps? Most likely I'm pushing this tonight so be quick.

src/game/SpellEffects.cpp
Phatcat: I am no good with auras, nor do I know the very specifics of this one skill. Barely heard of it, actually. I couldn't make out if sunder were cast at all during your code, you can maybe tell that I took a guess and it was wrong ^^

killerwife:
I think 2.0.3 is first TBC patch, 2.0.1 was pre-expansion patch. So no changes on our end.

Phatcat:
Yeah, no, you guys are right, I wasn't considering the pre-patch, I thought it was refering back to earlier, I do believe you guys, simply didn't know for sure.

Phatcat:
Merged in https://github.com/cmangos/mangos-tbc/commit/89fde25ddd480e49b99c9876acd65a56c9e64cd4

PR #112, opened 2016-08-13
Fixes water walking not being castable on players in various 'forms'

tbsp:
This applies to warriors (due to stances), priests in shadow form and stealthed rogues.

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -5308,7 +5308,7 @@ SpellCastResult Spell::CheckCast(bool strict)
                     Player const* player = static_cast<Player const*>(expectedTarget);
                     
                     // Player is not allowed to cast water walk on shapeshifted/mounted player 
-                    if (player->GetShapeshiftForm() != FORM_NONE || player->IsMounted())
+                    if (player->IsInDisallowedMountForm() || player->IsMounted())
                         return SPELL_FAILED_BAD_TARGETS;
                 }
 


killerwife:
Looks good to me tbh.

tbsp:
I just tested it; ghost wolf and travel form both report "invalid target".

killerwife:
Requesting merge on this please. I can vouch for its functionality. @Phatcat @cyberium

Phatcat:
and it applies to all versions?

killerwife:
To my knowledge, it does.

tbsp:
I checked and believe the same change applies to classic, but could not find the same checks being done in wotlk.

Phatcat:
> To my knowledge, it does. But wotlk can have waterwalking when mounted i think.

This would explain why the check is not done in Wrath. Looks like the new check also pretty much only looks for the player being mounted, no?

If that is the case I am gonna slap it on the TBC and the Classic branch and that'll be it... unless water walking was maybe not allowed in classic when in form? - I don't know, never did much water walking..

killerwife:
Its exactly what this patch is about. You cant have waterwalking in druid forms but u can have it in warrior stances (all are classified as shapeshift).

Phatcat:
So the first check would then actually also need to be applied to the wrath core, unless they changed that as well for wrath? (like mounted water-walking)

killerwife:
Afaik in wotlk u can have it in all forms too :)

Phatcat:
Just gonna roll with that for now, then.

PR #113, opened 2016-08-14
Remove SD2Config and use global sConfig

kartoffelsup:
Since SD2 no longer has its own config file but
instead, uses the mangosd config it should not
be necessary to load the same config file twice.

Also, this would not work when the config path
was set via the command line argument '-c'.

### DIFF
--- a/src/scriptdev2/ScriptMgr.cpp
+++ b/src/scriptdev2/ScriptMgr.cpp
@@ -17,13 +17,11 @@ typedef std::vector<Script*> SDScriptVec;
 int num_sc_scripts;
 SDScriptVec m_scripts;
 
-Config SD2Config;
-
 void FillSpellSummary();
 
 void LoadDatabase()
 {
-    std::string strSD2DBinfo = SD2Config.GetStringDefault("WorldDatabaseInfo");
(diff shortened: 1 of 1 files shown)


src/scriptdev2/ScriptMgr.cpp
```diff
@@ -89,56 +87,48 @@ void FreeScriptLibrary()
...
-       outstring_log("SD2: Using configuration file %s", _MANGOSD_CONFIG);
+   // ScriptDev2 startup
+   outstring_log("");
+   outstring_log(" MMM  MMM    MM");
+   outstring_log("M  MM M  M  M  M");
+   outstring_log("MM    M   M   M");
+   outstring_log(" MMM  M   M  M");
+   outstring_log("   MM M   M MMMM");
+   outstring_log("MM  M M  M ");
+   outstring_log(" MMM  MMM  http://www.scriptdev2.com");
```
Phatcat: Now you're at it we probably wanna update this to the cmangos SD2 subforum, (link is in the README)

xfurry:
I don't think we need this.
We already use the global config in SD2. Actually, this commit breaks the logic in the config, as we need to initialize it anyway.
I would say that we keep thing as they stand today.

PR #114, opened 2016-08-17
Crowd Control handling improvements

Warlockbugs:
I've noticed quite a few issues with CC effects. After examining the code closely, i've rewritten a part of it and introduced a new unified logical construct: Incapacitated State.
Incapacitated State is a state causing unit to lose ability to move, attack/cast and creatures to lose their current targets.
The following unit states are counted as incapacitated:
* Stunned
* Confused
* Fleeing

Previous handlers are wiped and transformed into shortcuts to a new function SetIncapacitatedState for compatibility. The function itself has the capability to set/unset up to all 3 states at once due to the state being a bitflag, but currently it is used in an individual manner only.

Incapacitated movements are now updated in a proper order:
1. Apply/remove stun (fully immobilized)
2. Confused (walking wandering) over -> Fleeing (running wandering)

The list of issues resolved:
* Creatures will no longer resume fighting on Fear or Confuse effect expiration if more Fear or Confuse effects are applied on it (Fear/Death Coil/Pplymorph/Blind/etc overlapping)
* Confused creatures will now correctly lose their current target
* Fleeing creatures will no longer be immobilized on confusion effect application and will wander around at walking speed instead
* Fixed a heavily exploitable case with confused creatures (including polymorphed) no longer being able to attack after effect overlapping by a fear (copy-pasted MovementExpired from Fear on confusion application)
* Fixed a case with wandering fleeing creatures continuing to auto-attack passers-by when fleeing preventing aura was removed and fleeing movement was re-initiated (Curse of Recklessnes)
* Fixed granting client control on expiration of Fear/Confusion effects to a player in a completed battleground (Score screen)

### DIFF
--- a/src/game/Player.cpp
+++ b/src/game/Player.cpp
@@ -19319,6 +19319,14 @@ void Player::ResurectUsingRequestData()
     SpawnCorpseBones();
 }
 
+bool Player::IsClientControl(Unit* target) const
+{
+    return (target && !target->IsFleeing() && !target->IsConfused() && !target->IsTaxiFlying() &&
+            (target->GetTypeId() != TYPEID_PLAYER ||
+            !((Player*)target)->InBattleGround() || ((Player*)target)->GetBattleGround()->GetStatus() != STATUS_WAIT_LEAVE) &&
+            target->GetCharmerOrOwnerOrOwnGuid() == GetObjectGuid());
+}
+
 void Player::SetClientControl(Unit* target, uint8 allowMove)
 {
     WorldPacket data(SMSG_CLIENT_CONTROL_UPDATE, target->GetPackGUID().size() + 1);
--- a/src/game/Player.h
+++ b/src/game/Player.h
@@ -2028,6 +2028,7 @@ class MANGOS_DLL_SPEC Player : public Unit
         bool IsFlying() const { return m_movementInfo.HasMovementFlag(MOVEFLAG_FLYING); }
         bool IsFreeFlying() const { return HasAuraType(SPELL_AURA_MOD_FLIGHT_SPEED_MOUNTED) || HasAuraType(SPELL_AURA_FLY); }
 
+        bool IsClientControl(Unit* target) const;
         void SetClientControl(Unit* target, uint8 allowMove);
         void SetMover(Unit* target) { m_mover = target ? target : this; }
         Unit* GetMover() const { return m_mover; }
--- a/src/game/SpellAuras.cpp
+++ b/src/game/SpellAuras.cpp
@@ -3499,12 +3499,20 @@ void Aura::HandleModConfuse(bool apply, bool Real)
     if (!Real)
         return;
 
+    // Do not remove it yet if more effects are up, do it for the last effect
+    if (!apply && GetTarget()->HasAuraType(SPELL_AURA_MOD_CONFUSE))
+        return;
+
     GetTarget()->SetConfused(apply, GetCasterGuid(), GetId());
 }
 
 void Aura::HandleModFear(bool apply, bool Real)
 {
     if (!Real)
+        return;
+
+    // Do not remove it yet if more effects are up, do it for the last effect
+    if (!apply && GetTarget()->HasAuraType(SPELL_AURA_MOD_FEAR))
         return;
 
     GetTarget()->SetFeared(apply, GetCasterGuid(), GetId());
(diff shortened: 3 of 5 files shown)


Blumfield:
That's way better than my current implementation, is there any side effect that you have noticed ?

Warlockbugs:
Since i mostly played with original methods' contents, i doubt it became much buggier. I spammed many CC effects at once via debug commands, and didn't get any occasion of creature locking up in passive state or spamming auto-attack at inappropriate times.
The areas i am less certain of are new ones:
* How creatures retreating ("flees away in fear" on low health) plays around with Fear/Confusion effects since movement reapplication is a new thing
* Client Control interaction with Mind Control, including slave unit being Feared/Confused. I've added the owner check, but Mind Control is a danger zone by itself, i am afraid to even touch it further.

Warlockbugs:
I've tested creature retreating and mind control and didn't find any side effects.
In addition i've refactored the code for better understanding and optimized it a bit (less calls and repeating operations).

Warlockbugs:
I found minor collisions with mind control after exhaustive tests, but they not directly related to this patch in any way, but to mind control itself (its handlers). It's present in the master. Hell, in the master it is many times worse, because CC effects collide by itself. I'll look into, probably in a separate PR.

Phatcat:
Mind control has been completely reworked by recent changes so you may want to retest that part of this patch.

Warlockbugs:
These MC changes are not in the TBC branch yet, so this will have to wait a bit it seems.

Phatcat:
Oh but they are. They got pushed to master last night and is pending downport to classic.
Look for the large commit by cyberium :)

Warlockbugs:
A few collisions were eliminated, but minor collisions with MC effect fading/application left:
- When MC fades from stunned/feared/confused creature, it acquires target and resumes fighting until one of CC effects will expire (and proper behavior will be re-initialized).
- When player gains control of already stunned/feared/confused, he is able to act freely with controlled creature as if none of these effects were applied (untill one of them expires and appropriate state is updated).

I guess i have to check this out for this PR then.

Phatcat:
> When MC fades from stunned/feared/confused creature, it acquires target and resumes fighting until one of CC effects will expire (and proper behavior will be re-initialized).

This sounds like an AI thing. I am working on pretty much all the AIs the moment and may already have this solved actually. If you can make sure a unit affected by such an aura effect has UNIT_STAT_CAN_NOT_REACT then I can handle this in the AI, where it should be handled (it's the AI forcing it to attack currently)

Warlockbugs:
Not just AI, but player can act while gaining MC over already Feared/Confused creature. That shouldn't happen, gaining MC in this case should just grant vision and action bar, ability to act should not be granted until all effects are gone. I am currently looking into it.

Phatcat:
I was only referencing your first point, your second point most likely lies close to @cyberium 's MC work so you probably want to poke him about it as well for some insight.

Phatcat:
> This sounds like an AI thing. I am working on pretty much all the AIs the moment and may already have this solved actually.

this has not been pushed to master, though ,it is still in testing phase and getting expanded on and will most likely not get pushed before GuardianAI has been merged to the TBC and classic as well since it also includes some work on that.

Warlockbugs:
Also, we have some sort of duality. We have UnitFlags for Stunned/Feared/Confused and UnitStates for Stunned/Feared/Confused. My thought would be using UnitFlags for spell effects only, and UnitStates for movement generators and all that internal stuff. That's what i am doing now.

Warlockbugs:
Here's up-to-date squash of the PR.

Changes: 

- I added UNIT_STAT_CONFUSED to UNIT_STAT_LOST_CONTROL
- Player will no longer be granted full control over MCed creature while confusion/fear effects present
- Added and refactored some getters for IncapacitatedState construct

Unresolved collisions with MC:

- Fleeing and Confused movement generators are removed on MC application/remove instead of being preserved if appropriate effects are still present
- AI is also resuming action instead of abiding IncapacitatedState.

I don't wanna mess up with Cyberium's mastermind on this subject.

cyberium:
Nice chat, nice work.
About the MC movement i knew that. I think i already added a TODO in the code about it. I forced the movement somewhat to avoid having to manage it as it was not fully related with MC.
Now you just force me to rework that part ;)

About UNIT_STAT_ vs UNIT_FLAG,_ Thay have not same purpose, But they have to be set accordinaly.
I think the best way to handle those is to keep that in mind :
- UNIT_FLAG is the char animation. 
- UNIT_STAT is the current state of the unit.
- As UNIT_STAT was introduced long time after UNIT_FLAG they are direct use of UNIT_FLAG in core. This is bad.
- We have to use UNIT_STATE to know the real current state of the UNIT and set the UNIT_FLAG accordinaly, the UNIT_FLAG depend of the UNIT_STAT not the inverse. I think this need a complete rewrite to add a more clear class that will manage that for us.

Warlockbugs:
Currently all these UNIT_FLAG are handled through SetIncapacitatedState, so if these flags are set, its guaranteed that some aura applied it. I have left the comment about this on getters. Other than that, i think its ready.
WTB more reviews. :smile:

Warlockbugs:
Updated. Now with additional unified handler for roots and stuns: Immobilized State.

Warlockbugs:
One of the commits unfolded a bug in root handler, which i discovered while rewriting it. With recent commits it made npcs teleporting after root effect was removed instead of moving. I intend to push it soon. Any last minute opinions/suggestions?

Warlockbugs:
Merged in https://github.com/cmangos/mangos-tbc/commit/bbe7bd7c44d48ed24e5e57bc64bc3e81690f51ac.

PR #115, opened 2016-08-18
Group improvements

Warlockbugs:
A new feature ~~and 2 backports~~.
Implemented a new logic similar to blizz's for choosing a member who is online (and check for assist status when in raids) as a new leader in case of group leader leaving the group. Now leadership wont be transferred to an offline group member, effectively screwing up remaining online group members.
~~Restored Main Assisst/Main Tank feature with 2 backports from WotLK.~~ (Merged separately)


killerwife:
@war

Warlockbugs:
Prospected improvements TODOs to combine with now closed [classic PR](https://github.com/cmangos/mangos-classic/pull/98):
* Add a 5min inactivity timer for an offline group leader

I got distracted by aura tasks atm, i've stopped at planning how to implement an efficient separate timer and loaded groups iterator.

Tagging @Phatcat

Warlockbugs:
@Phatcat 
A comment about the "transfer inactive leadership" feature itself: It was all tied to a slot reservation. If you remember, on a full server, you had a 5min grace period to log in back without getting into a queue again. If a player's slot expired, its group leadership was transfered (and he was kicked from the battleground group, if this was the case). It worked even for long offline times: for example, whole group logged off an hour ago or even several days ago, the first player to log in will receive the leadership, because the leader's slot in this case was expired long ago. The leadership transfer is only the case for groups with 1 or more players online, groups whose entire list are offline are excluded from this check.

Phatcat:
I didn't know those 2 things were tied together honestly but it doesn't sound too far out; I only raided as a raid member, never did any officer roles or anything so never paid much attention to how leader was handled in all honesty. I wasn't that old back then either, but I do remember not having to queue up again.

I guess the change the other PR wanted to introduce may have been Blizzlike after all.

Warlockbugs:
Merged WotLK backports to get it out of the way. This PR will be used for features only.

Phatcat:
Do the backports need to be ported back to classic as well or was it something only missing on TBC?

Warlockbugs:
These were TBC only changes. No main tanks or other fancy stuff for vanilla as far as i know.

Phatcat:
I don't remember any of that at all healing in vanilla, no.

Warlockbugs:
Implemented offline leader transfer feature with customizable delay setting.

Obligatory honorary mention: @DSkywalk for trying to do this first on classic.

PR #116, opened 2016-08-19
Fix multi-instance flight paths after client control patch

Warlockbugs:
Self-explanatory. Fixes Ghostlands<>EPL flight path. Serverside processing, because client no longer spams us with useless packets.

Also, control patch potentially fixed an old bug with players who are flying-by on taxi randomly visually falling to the ground level before disappearing (because we are not broadcasting movement garbage from client anymore).

Can be applied to vanilla for the sake of unification, though there arent any multi-map flight paths.

[Read the full story here] (https://github.com/cmangos/mangos-wotlk/commit/61fc35b3c83f15aa548bd68a111f8d21a921b18a)

### DIFF
--- a/src/game/TaxiHandler.cpp
+++ b/src/game/TaxiHandler.cpp
@@ -199,35 +199,10 @@ void WorldSession::HandleMoveSplineDoneOpcode(WorldPacket& recv_data)
     recv_data >> movementInfo;
     recv_data >> Unused<uint32>();                          // unk
 
-
-    // in taxi flight packet received in 2 case:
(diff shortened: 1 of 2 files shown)


Phatcat:
:+1: for it being compatible with the classic branch.

killerwife:
Later down the line, having small diff will be very important imo so +1

Phatcat:
This also looks a lot more clean than whatever we were doing before. I am gonna test and report on this soon (poke me if I have yet to report back on this within this week)

PR #118, opened 2016-08-31
EventAI reset fix

killerwife:
Changed from reset on enter combat, on reset during Reset function,
except for GENERIC, which is properly reset only on respawn

@cyberium this is what I was talking about, pls merge this :)


cyberium:
Thank you. Do you remember how to test it to see before/after result?

killerwife:
Dont merge it yet, found a bug XD

PR #120, opened 2016-09-02
Implement experience multiplier

killerwife:
Selfexplanatory

### DIFF
--- a/src/game/Player.cpp
+++ b/src/game/Player.cpp
@@ -2326,6 +2326,9 @@ void Player::GiveXP(uint32 xp, Unit* victim)
     if (level >= sWorld.getConfig(CONFIG_UINT32_MAX_PLAYER_LEVEL))
         return;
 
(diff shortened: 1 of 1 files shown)


Phatcat:
Thank you!

Phatcat:
this seems too simple, though... where is the actual exp calculation being done?

Fabi:
https://github.com/cmangos/mangos-tbc/blob/master/src/game/Formulas.h#L110

Phatcat:
That is where it should be applied imo

Phatcat:
that is exactly where it fits, together with `sWorld.getConfig(CONFIG_FLOAT_RATE_XP_KILL)`

https://github.com/cmangos/mangos-tbc/blob/master/src/game/Formulas.h#L124

Probably that whole function could do with a small lift-up =)

Phatcat:
Implemented.

PR #121, opened 2016-09-04
Linkagechanges

killerwife:
Fixes triggering of linkage events, moved it to respective AIs to provide additional control.

Also implemented evade linking flag.

Potentially breaks wrongly designed SD2 scripts.

@cyberium

### DIFF
--- a/src/game/CreatureLinkingMgr.h
+++ b/src/game/CreatureLinkingMgr.h
@@ -68,6 +68,7 @@ enum CreatureLinkingFlags
     FLAG_RESPAWN_ON_DEATH           = 0x0040,
     FLAG_RESPAWN_ON_RESPAWN         = 0x0080,
     FLAG_DESPAWN_ON_RESPAWN         = 0x0100,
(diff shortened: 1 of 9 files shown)


src/game/CreatureLinkingMgr.h
```diff
@@ -68,6 +68,7 @@ enum CreatureLinkingFlags
     FLAG_RESPAWN_ON_DEATH           = 0x0040,
     FLAG_RESPAWN_ON_RESPAWN         = 0x0080,
     FLAG_DESPAWN_ON_RESPAWN         = 0x0100,
+    FLAG_EVADE_ON_EVADE             = 0x4000,
```
Phatcat: should be EVENT_ON_EVADE , no?

killerwife: no, it means slave evades when master evades, so the description is on point.

robinsch: EVADE_ON_OWNER_EVADE

PR #123, opened 2016-09-10
Fix swiftmend ignoring owner of hots

Warlockbugs:
Forces Swiftmend ability to interact only with caster's HoTs instead of wiping all druid HoTs on the target.

### DIFF
--- a/src/game/SpellEffects.cpp
+++ b/src/game/SpellEffects.cpp
@@ -3050,7 +3050,7 @@ void Spell::EffectHeal(SpellEffectIndex /*eff_idx*/)
             int32 tickheal = targetAura->GetModifier()->m_amount;
             int32 tickcount = GetSpellDuration(targetAura->GetSpellProto()) / targetAura->GetSpellProto()->EffectAmplitude[idx] - 1;
 
-            unitTarget->RemoveAurasDueToSpell(targetAura->GetId());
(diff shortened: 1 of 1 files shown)


Phatcat:
About the change itself, yes, we should of course only remove hots from the caster and not hots from all druids. I havn't verified this issue myself but I'm thinking you're not proposing a fix for nothing.

Code wise I think this looks decent. Personally I prefer the style of having the boolean operators first on a line instead of last on the line above, on logical operations spanding several lines of code. very minor thing which I doubt anyone besides me even cares at all about.

About the method it seems as simple as it can get, the way you compare it, so I canot see how that could fail. I don't quite understand the other change, I am very unfamiliar with the aura system, but you most likely didn't change it just for the fun of it so... I wouldn't be able to say if this is the proper way to do this.

Warlockbugs:
The second part just removes the aura owned by appropriate caster instead of wiping all auras of the same spell id.

ItsmeAga:
http://wowwiki.wikia.com/wiki/Swiftmend?oldid=1562661

"A player can consume a Regrowth or Rejuvenation with Swiftmend casted by another druid. In this case it uses the base value of the HoT casters spell, but uses the +healing of the caster of Swiftmend."

wouldn't this fix prevent using other's hots?

Warlockbugs:
Reverted filtering part. AFAIK this applies to all cores.

PR #126, opened 2016-09-13
Fix SMSG_SPELLLOGEXECUTE structure and data

Laizerox:
Cleanup spell log execution
Simplify a lot of code
Fix almost all issues with combat log?

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -3645,100 +3645,29 @@ void Spell::WriteSpellGoTargets(WorldPacket* data)
 
 void Spell::SendLogExecute()
 {
-    Unit* target = m_targets.getUnitTarget() ? m_targets.getUnitTarget() : m_caster;
-
     WorldPacket data(SMSG_SPELLLOGEXECUTE, (8 + 4 + 4 + 4 + 4 + 8));
 
-    if (m_caster->GetTypeId() == TYPEID_PLAYER)
-        data << m_caster->GetPackGUID();
-    else
-        data << target->GetPackGUID();
-
+    data << m_caster->GetPackGUID();
     data << uint32(m_spellInfo->Id);
-    uint32 count1 = 1;
-    data << uint32(count1);                                 // count1 (effect count?)
-    for (uint32 i = 0; i < count1; ++i)
+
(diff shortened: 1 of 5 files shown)


Warlockbugs:
I can confirm that extra attacks and counterspell are displayed properly with this patch.

src/game/Spell.cpp
```diff
@@ -6910,3 +6839,53 @@ void Spell::GetSpellRangeAndRadius(SpellEffectIndex effIndex, float& radius, uin
...
+    logEffect << param4;
+
+    m_executeLog[effIndex].push_back(logEffect);
+}
+
+template void Spell::LogEffectExecute(uint8, uint32);
+template void Spell::LogEffectExecute(uint8, PackedGuid);
+template void Spell::LogEffectExecute(uint8, PackedGuid, uint32);
+template void Spell::LogEffectExecute(uint8, PackedGuid, int32, int32);
+template void Spell::LogEffectExecute(uint8, PackedGuid, uint32, uint32, float);
```
RAlexis: New line

Phatcat:
@Laizerox merge errors you may wanna look at after recent commits

Laizerox:
@Phatcat This is still kind of WIP and I will merge it in the next update likely.

src/game/Spell.cpp
mityada: Why not use variadic templates here?

```
void FillExecuteLog(ByteBuffer &buffer)
{
}

template <typename P, typename ... Params>
void FillExecuteLog(ByteBuffer &buffer, P param, Params ... params)
{
    buffer << param;
    FillExecuteLog(buffer, params...);
}

template <typename ... Params>
void Spell::LogEffectExecute(uint8 effIndex, Params ... params)
{
    ByteBuffer logEffect;
    FillExecuteLog(logEffect, params...);

    m_executeLog[effIndex].push_back(logEffect);
}
```

Laizerox:
@mityada it was the plan but after thinking of it only overload that would require it is the one with 1 param.

Also we would still need to declare possible templates and templates can be too ambigous. Which means my final version which I havent push yet will have 5 overloads instead.

PR #129, opened 2016-09-22
Eaicommands

killerwife:
Implement two EAI commands required for this kind of behaviour:

https://gist.github.com/killerwife/569e3517c36634d116c620c7a1c93c91

The pause waypoints is a mirror to DBSCRIPT, so that we can make scripts work when creature leaves combat again.

The Interrupt Spell is to provide interrupt command when creature casts some kind of infinite spell that prevents it from making any actions in combat, like for example the green beam spell.

@cyberium @xfurry @grz3s

### DIFF
--- a/doc/EventAI.txt
+++ b/doc/EventAI.txt
@@ -156,6 +156,8 @@ For all ACTION_T_RANDOM Actions, When a Particular Param is selected for the Eve
 48   ACTION_T_CHANGE_MOVEMENT               MovementType, WanderDistance    Change the unit movement type (Param1). If the movement type is Random Movement (1), the WanderDistance (Param2) must be provided. If the movement type is Waypoint Movement (2), Param2 is PathId.
 49   ACTION_T_DYNAMIC_MOVEMENT              EnableDynamicMovement           Enable dynamic movement behavior (1 = on; 0 = off)
 50   ACTION_T_SET_REACT_STATE               ReactState                      Change react state of the creature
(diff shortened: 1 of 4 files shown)


Ulduar:
@xfurry quotes:

We can't change the dbscript engine just to accommodate ONE boss. We are not blizz, and that's why our scripting system is in C++ and not any fancy complex spell + script thingy.

dbscripts are purposed to handled non combat events only and eventAI for combat only.
You can have a mix of both, in for this boss it will simply NOT work!

There are many ways in which we can improve dbscript or eventAI engines, but doing this just for the sake of such a complex boss, is ridiculous.

So if you want to submit a C++ script for this guy, I'm happy to help.

@killerwife  For one npc they will not be added. Here and so!
They prefer to handle this in SD2.

killerwife:
I am gonna point out a simple counterargument:

This spell is used all over outland to script green beams, and they are not interrupted on entering combat. As a result this is not for "one mob", if it were, I would not even submit this. The Pause Waypoints is merely a port of DBSCRIPT behaviour to EAI, to enable same possibilities (consistency if you will). Besides, there is plenty of examples where you need to interrupt spellcasting in script.

Ulduar:
https://github.com/cmangos/mangos-cata/pull/47#issuecomment-153840464
ACTION_T_PAUSE / UNPAUSE_MOVEMENT (maybe)

cyberium:
@killerwife please update accordinaly eai documentation before merging this.

killerwife:
I think I can do better, I will update documentation and then I will update wiki on missing eai commands.

PR #130, opened 2016-09-23
Fix blink visual and implement target type

killerwife:


robinsch:
Just curious: What was wrong with the blink visual?

killerwife:
@robinsch TARGET_55 wasnt implemented and we werent sending dest location for effect 1 which is responsible to show small flickering light on the ground

cyberium:
merged thanks

PR #131, opened 2016-09-24
Miscellaneous pet fixes

killerwife:
@phatcat

### DIFF
--- a/src/game/NPCHandler.cpp
+++ b/src/game/NPCHandler.cpp
@@ -481,6 +481,7 @@ void WorldSession::SendStablePet(ObjectGuid guid)
     data << uint8(GetPlayer()->m_stableSlots);
 
     uint8 num = 0;                                          // counter for place holder
+    uint32 firstSlot = PET_SAVE_FIRST_STABLE_SLOT;
 
     // not let move dead pet in slot
     if (pet && pet->isAlive() && pet->getPetType() == HUNTER_PET)
@@ -493,10 +494,12 @@ void WorldSession::SendStablePet(ObjectGuid guid)
         data << uint8(0x01);                                // client slot 1 == current pet (0)
         ++num;
     }
+    else
+        firstSlot = 0;
 
     //                                                     0      1     2   3      4      5        6
     QueryResult* result = CharacterDatabase.PQuery("SELECT owner, slot, id, entry, level, loyalty, name FROM character_pet WHERE owner = '%u' AND slot >= '%u' AND slot <= '%u' ORDER BY slot",
-                          _player->GetGUIDLow(), PET_SAVE_FIRST_STABLE_SLOT, PET_SAVE_LAST_STABLE_SLOT);
+                          _player->GetGUIDLow(), firstSlot, PET_SAVE_LAST_STABLE_SLOT);
 
     if (result)
     {
--- a/src/game/NPCHandler.cpp
+++ b/src/game/NPCHandler.cpp
@@ -33,6 +33,7 @@
 #include "Guild.h"
 #include "GuildMgr.h"
 #include "Chat.h"
+#include "World.h"
 
 enum StableResultCode
 {
@@ -578,8 +579,18 @@ void WorldSession::HandleStablePet(WorldPacket& recv_data)
 
     Pet* pet = _player->GetPet();
 
+    if (!pet) // if pet not by your side, load from DB for a moment
+    {
+        pet = new Pet;
+        if (!pet->LoadPetFromDB(_player, 0))
+        {
+            delete pet;
+            SendStableResult(STABLE_ERR_STABLE);
+            return;
+        }
+    }
     // can't place in stable dead pet
-    if (!pet || !pet->isAlive() || pet->getPetType() != HUNTER_PET)
+    else if (!pet->isAlive() || pet->getPetType() != HUNTER_PET)
     {
         SendStableResult(STABLE_ERR_STABLE);
         return;
@@ -611,6 +622,9 @@ void WorldSession::HandleStablePet(WorldPacket& recv_data)
 
     if (free_slot > 0 && free_slot <= GetPlayer()->m_stableSlots)
     {
+        if (pet->GetCharmInfo()->GetPetNumber() == _player->GetTemporaryUnsummonedPetNumber())
+            _player->SetTemporaryUnsummonedPetNumber(0);
+
         pet->Unsummon(PetSaveMode(free_slot), _player);
         SendStableResult(STABLE_SUCCESS_STABLE);
     }
@@ -633,14 +647,16 @@ void WorldSession::HandleUnstablePet(WorldPacket& recv_data)
     }
 
     uint32 creature_id = 0;
+    uint32 slot = 0;
 
     {
-        QueryResult* result = CharacterDatabase.PQuery("SELECT entry FROM character_pet WHERE owner = '%u' AND id = '%u' AND slot >='%u' AND slot <= '%u'",
+        QueryResult* result = CharacterDatabase.PQuery("SELECT slot,entry FROM character_pet WHERE owner = '%u' AND id = '%u' AND slot >='%u' AND slot <= '%u'",
                               _player->GetGUIDLow(), petnumber, PET_SAVE_FIRST_STABLE_SLOT, PET_SAVE_LAST_STABLE_SLOT);
         if (result)
         {
             Field* fields = result->Fetch();
-            creature_id = fields[0].GetUInt32();
+            slot = fields[0].GetUInt32();
+            creature_id = fields[1].GetUInt32();
             delete result;
         }
     }
@@ -665,9 +681,25 @@ void WorldSession::HandleUnstablePet(WorldPacket& recv_data)
         return;
     }
 
-    // delete dead pet
+    // delete temporary pet
     if (pet)
         pet->Unsummon(PET_SAVE_AS_DELETED, _player);
+    else 
+    {
+        pet = new Pet; // load dismissed pet temporarily
+        if (pet->LoadPetFromDB(_player, 0))
+            pet->Unsummon(PetSaveMode(slot), _player);
+        else
+        {
+            if (pet->TryLoadPetFromDB(_player, 0) == SPELL_FAILED_TARGETS_DEAD)
+            {
+                delete pet;
+                SendStableResult(STABLE_ERR_STABLE);
+                return;
+            }
+            delete pet;
+        }
+    }
 
     Pet* newpet = new Pet(HUNTER_PET);
     if (!newpet->LoadPetFromDB(_player, creature_id, petnumber))
@@ -679,6 +711,9 @@ void WorldSession::HandleUnstablePet(WorldPacket& recv_data)
     }
 
     SendStableResult(STABLE_SUCCESS_UNSTABLE);
+
+    if ((sWorld.getConfig(CONFIG_BOOL_PET_UNSUMMON_AT_MOUNT) && _player->HasAuraType(SPELL_AURA_MOUNTED)) || _player->HasAuraType(SPELL_AURA_MOD_FLIGHT_SPEED_MOUNTED))
+        newpet->Unsummon(PET_SAVE_AS_CURRENT); // TODO: remove adding to world if mounted
 }
 
 void WorldSession::HandleBuyStableSlot(WorldPacket& recv_data)
@@ -731,7 +766,17 @@ void WorldSession::HandleStableSwapPet(WorldPacket& recv_data)
 
     Pet* pet = _player->GetPet();
 
-    if (!pet || pet->getPetType() != HUNTER_PET)
+    if (!pet) // if pet not by your side, load from DB for a moment
+    {
+        pet = new Pet;
+        if (!pet->LoadPetFromDB(_player, 0))
+        {
+            delete pet;
+            SendStableResult(STABLE_ERR_STABLE);
+            return;
+        }
+    }
+    else if (pet->getPetType() != HUNTER_PET || !pet->isAlive())
     {
         SendStableResult(STABLE_ERR_STABLE);
         return;
@@ -765,8 +810,8 @@ void WorldSession::HandleStableSwapPet(WorldPacket& recv_data)
         return;
     }
 
-    // move alive pet to slot or delete dead pet
-    pet->Unsummon(pet->isAlive() ? PetSaveMode(slot) : PET_SAVE_AS_DELETED, _player);
+    // move pet to slot
+    pet->Unsummon(PetSaveMode(slot), _player);
 
     // summon unstabled pet
     Pet* newpet = new Pet;
@@ -776,7 +821,11 @@ void WorldSession::HandleStableSwapPet(WorldPacket& recv_data)
         SendStableResult(STABLE_ERR_STABLE);
     }
     else
+    {
         SendStableResult(STABLE_SUCCESS_UNSTABLE);
+        if ((sWorld.getConfig(CONFIG_BOOL_PET_UNSUMMON_AT_MOUNT) && _player->HasAuraType(SPELL_AURA_MOUNTED)) || _player->HasAuraType(SPELL_AURA_MOD_FLIGHT_SPEED_MOUNTED))
+            newpet->Unsummon(PET_SAVE_AS_CURRENT); // TODO: remove adding to world if mounted
+    }
 }
 
 void WorldSession::HandleRepairItemOpcode(WorldPacket& recv_data)
--- a/src/game/Pet.cpp
+++ b/src/game/Pet.cpp
@@ -141,7 +141,7 @@ SpellCastResult Pet::TryLoadFromDB(Player* owner, uint32 petentry, uint32 petnum
     }
 
     uint32 summon_spell_id = fields[21].GetUInt32();
-    SpellEntry const* spellInfo = sSpellTemplate.LookupEntry<SpellEntry>(summon_spell_id);
+    SpellEntry const* spellInfo = sSpellStore.LookupEntry(summon_spell_id);
 
     bool is_temporary_summoned = spellInfo && GetSpellDuration(spellInfo) > 0;
 
--- a/src/game/Pet.cpp
+++ b/src/game/Pet.cpp
@@ -90,7 +90,7 @@ void Pet::RemoveFromWorld()
     Unit::RemoveFromWorld();
 }
 
-bool Pet::TryLoadFromDB(Player* owner, uint32 petentry, uint32 petnumber, bool current)
+SpellCastResult Pet::TryLoadFromDB(Player* owner, uint32 petentry, uint32 petnumber, bool current)
 {
     uint32 ownerid = owner->GetGUIDLow();
 
@@ -120,7 +120,7 @@ bool Pet::TryLoadFromDB(Player* owner, uint32 petentry, uint32 petnumber, bool c
             ownerid, PET_SAVE_AS_CURRENT, PET_SAVE_LAST_STABLE_SLOT);
 
     if (!result)
-        return false;
+        return SPELL_FAILED_NO_PET;
 
     Field* fields = result->Fetch();
 
@@ -129,7 +129,7 @@ bool Pet::TryLoadFromDB(Player* owner, uint32 petentry, uint32 petnumber, bool c
     if (!petentry)
     {
         delete result;
-        return false;
+        return SPELL_FAILED_NO_PET;
     }
 
     CreatureInfo const* creatureInfo = ObjectMgr::GetCreatureTemplate(petentry);
@@ -137,7 +137,7 @@ bool Pet::TryLoadFromDB(Player* owner, uint32 petentry, uint32 petnumber, bool c
     {
         sLog.outError("Pet entry %u does not exist but used at pet load (owner: %s).", petentry, owner->GetGuidStr().c_str());
         delete result;
-        return false;
+        return SPELL_FAILED_NO_PET;
     }
 
     uint32 summon_spell_id = fields[21].GetUInt32();
@@ -149,7 +149,7 @@ bool Pet::TryLoadFromDB(Player* owner, uint32 petentry, uint32 petnumber, bool c
     if (current && is_temporary_summoned)
     {
         delete result;
-        return false;
+        return SPELL_FAILED_NO_PET;
     }
 
     PetType pet_type = PetType(fields[22].GetUInt8());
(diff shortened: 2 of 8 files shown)


Phatcat:
build issues would have to get sorted, of course.

src/game/Spell.cpp
```diff
@@ -5100,9 +5100,7 @@ SpellCastResult Spell::CheckCast(bool strict)
                     else
                     {
                         Pet* dbPet = new Pet;
-                        if (dbPet->LoadPetFromDB((Player*)m_caster, 0))
-                            return SPELL_CAST_OK;           // still returns an error to the player, so this error must come from somewhere else...
-                        else
+                        if (!dbPet->TryLoadFromDB((Player*)m_caster, 0))
```
Phatcat: if there is something to load, we continue to SummonPet, which is fine, but this means that the check in summonPet is redundant, because if the hunter has nothing to load, then he will return spell failed instead and never go through to SummonPet -> this should also make the whole skip bool redundant.

src/game/SpellEffects.cpp
```diff
@@ -4594,15 +4594,17 @@ void Spell::EffectSummonPet(SpellEffectIndex eff_idx)
...
     {
         switch(m_caster->getClass())
         {
             case CLASS_HUNTER:
             {
-                // Everything already taken care of, we are only here because we loaded pet from db successfully
-                delete NewSummon;
-                return;
+                // Everything already taken care of, we are only here because we loaded pet from db successfully                
+                skip = NewSummon->LoadPetFromDB((Player*)m_caster, 0);;
```
Phatcat: okay, here's an issue, you got two ;; ending this.
also, this check should hopefully be done in CheckCast; if the hunter has no pet he should be stopped before reaching here; he should NOT go through the summon function.

src/game/Spell.cpp
killerwife: This must not be done here, because Effect sends SPELLOG, which will be implemented very soon.

Phatcat: Still can't have hunters going through the summon function ever, you have to return them in their own function (take a look at how it was done back in april, before any of my work - hunter's pets have always been returned out of summonPet early as their pets are not created by that function, only summoned via it, hunter's pets are created through Tamebeast or whatever it's called -> most of what is in SummonPet is for creating the pet, this is why warlocks only go through it once with each pet.

Phatcat: that was poorly worded, you can of course have them go through summonPet, but if you use the function NewSummon->LoadPetFromDB((Player*)m_caster, 0) and the pet is dead then you're summoning the pet dead, which is not blizzlike, it shouldn't be allowed to be summoned dead, so if you don't plan to stop that in checkCast you must stop it before doing LoadPetFromDB

killerwife: Ok, I will write that tomorrow.

Phatcat: @cyberium told me back when I suggested doing the check to see if the pet is dead in SummonEffect that it should be done in CheckCast and we shouldn't return any cast errors inside the spell effects themselves.

https://github.com/cmangos/mangos-tbc/commit/2271ecbeffa490f61b6ea4cc5910011ecd738c39

killerwife:
Fixed build and added numerous stable master issues fixes.

src/game/SpellEffects.cpp
```diff
@@ -4594,15 +4594,17 @@ void Spell::EffectSummonPet(SpellEffectIndex eff_idx)
 
     Pet* NewSummon = new Pet;
 
+    bool skip;
```
Phatcat: yeah, this skip system is still wrong, hunters must NEVER get to line 4650 or they will always be returned in 4655, along with a false flagged error.

src/game/NPCHandler.cpp
```diff
@@ -575,8 +579,18 @@ void WorldSession::HandleStablePet(WorldPacket& recv_data)
 
     Pet* pet = _player->GetPet();
 
+    if (!pet) // if pet not by your side, load from DB for a moment
```
Phatcat: It's not just 'loaded for a moment', not unless you unsummon it again because LoadPetFromDB spawns it.

src/game/NPCHandler.cpp
```diff
@@ -728,7 +756,17 @@ void WorldSession::HandleStableSwapPet(WorldPacket& recv_data)
 
     Pet* pet = _player->GetPet();
 
-    if (!pet || pet->getPetType() != HUNTER_PET)
+    if (!pet) // if pet not by your side, load from DB for a moment
```
Phatcat: It's not just 'loaded for a moment', not unless you unsummon it again because LoadPetFromDB spawns it.

killerwife: I do unsummon it again later. Unless you hack your outgoing packet, in which case you will lose pet.

src/game/Pet.h
```diff
@@ -167,8 +167,8 @@ class MANGOS_DLL_SPEC Pet : public Creature
 
         bool Create(uint32 guidlow, CreatureCreatePos& cPos, CreatureInfo const* cinfo, uint32 pet_number);
         bool CreateBaseAtCreature(Creature* creature);
-        bool LoadPetFromDB(Player* owner, uint32 petentry = 0, uint32 petnumber = 0, bool current = false);
-        bool TryLoadFromDB(Player* owner, uint32 petentry = 0, uint32 petnumber = 0, bool current = false);
+        bool LoadPetFromDB(Player* owner, uint32 petentry = 0, uint32 petnumber = 0, bool current = false, bool ignoreDead= false);
```
Phatcat: bool ignoreDead= false

bool ignoreDead = false

very small one.

src/game/NPCHandler.cpp
Phatcat: well, as long as it's unsummoned again I don't see an issue with it, if it's never visually displayed to the player.

src/game/Pet.h
killerwife: done

killerwife:
Forgot to delete variable and forgot to load non-hunter pets from DB even if dead. Incoming fix in a moment.

killerwife:
Added.

killerwife:
@Phatcat added more refinements

Fixed shadowfiend being saved to database.

robinsch:
Related to Shadowfiend I would rather try something linking the pet to the spell and don't postion shadowfiend next to the summoner.

killerwife:
Shadowfiend is positioned next to target, not summoner, and shadowfiend is not supposed to be saved to DB at all, so I do not see the need for any linkage, since he spawns with predefined states every time, without relation to previous settings.

robinsch:
Well I don't know how it works currently. I only had the bug that Shadowfiend was spawned next to caster and fixed it this way.

killerwife:
No such bug is present in cmangos at this moment to my knowledge.

Phatcat:
> Fixed shadowfiend being saved to database.

you sure we don't want to keep saving it to and loading it from the db, and handle this;

>  he spawns with predefined states every time, without relation to previous settings.

in SaveToDB/LoadFromDB instead? creating a new pet is heavy on the server, especially since it spawns with all the same stats each time, we really want to avoid having to re-calculate when we can just choose which things to load to and from the db and what not to.

killerwife:
Loading from DB is much much more expensive. So no, we do not want to save shadowfiend at all.

Phatcat:
What do you mean? you do know that in order to do the calculations we have to load values from a db anyway? ...

killerwife:
You do not, the values of petlevelstats are always in memory after launch. Character DB data (like character_pet table) is queried during runtime. And then consider the fact that Character DB is not local, even minor latency of 50ms, and in that time you can create millions of objects. Saving in Character DB is always worse than creating new pet.

Phatcat:
So creature_template is loaded for every single creature is loaded into memory after launch as well? because creating a new pet loads that data into CreatureInfo.

killerwife:
Yes creature_template is loaded once and is static in memory. Whole mangos database part works like that.

Phatcat:
Okay. In that case I guess it's fine to not be loading or saving the shadowfiend then, even though I think it's a waste it has to go through Pet::Create and Pet::InitStatsForLevel each time, but I guess it's the same for creature pets, those don't get loaded/saved either.

src/game/SpellEffects.cpp
```diff
@@ -7147,10 +7155,20 @@ void Spell::EffectSummonDeadPet(SpellEffectIndex /*eff_idx*/)
...
         return;
+
+    bool wasDespawned = false;
     Player* _player = (Player*)m_caster;
     Pet* pet = _player->GetPet();
     if (!pet)
-        return;
+    {
+        pet = new Pet;
+        if (!pet->LoadPetFromDB(_player, 0, 0, false, true))
```
Phatcat: If we can't load pet doesn't that mean there is no pet, not that it has been despawned? - also, it should be possible to resurrect your pet when it's not around, while it shouldn't be possible to call it close at all if it's dead.

not quite sure what exactly you're checking for here?

src/game/SpellEffects.cpp
```diff
@@ -3835,7 +3835,7 @@ bool Spell::DoSummonPet(SpellEffectIndex eff_idx)
     {
         // Load pet from db; if any to load
         if (m_caster->getClass() != CLASS_PRIEST
```
Phatcat: This may be something that changes with the rework of EffectSummonType, because priests and mage's don't summon their pets via SummonPet or SummonGuardian, they summon them via EffectSummon . (at least in vanilla / tbc) - if they now go through SummonPet (which is fine if they are) then you need to surpress them getting saved to the db, for consistency you know, this....

src/game/SpellEffects.cpp
```diff
@@ -4642,76 +4647,79 @@ void Spell::EffectSummonPet(SpellEffectIndex eff_idx)
...
-        NewSummon->SetUInt32Value(UNIT_NPC_FLAGS, cInfo->NpcFlags);
-        NewSummon->SetUInt32Value(UNIT_FIELD_FLAGS, cInfo->UnitFlags);
-
-        // Notify Summoner
-        if (m_originalCaster && (m_originalCaster != m_caster)
-            && (m_originalCaster->GetTypeId() == TYPEID_UNIT) && ((Creature*)m_originalCaster)->AI())
-            ((Creature*)m_originalCaster)->AI()->JustSummoned(NewSummon);
-        else if ((m_caster->GetTypeId() == TYPEID_UNIT) && ((Creature*)m_caster)->AI())
-            ((Creature*)m_caster)->AI()->JustSummoned(NewSummon);
+            NewSummon->SavePetToDB(PET_SAVE_AS_CURRENT);
```
Phatcat: here ^

src/game/Player.cpp
```diff
@@ -15841,8 +15841,9 @@ void Player::SaveToDB()
         _SaveStats();
 
     // save pet (hunter pet level and experience and all type pets health/mana).
-    if (Pet* pet = GetPet())
-        pet->SavePetToDB(PET_SAVE_AS_CURRENT);
+    if (getClass() != CLASS_PRIEST)
```
Phatcat: Wouldn't it be easier to place this check inside SavePetToDB itself (do it as a check before loading any data)?

killerwife: Yes, thats purely a design decision. Though I like it more this way tbh.

src/game/SpellEffects.cpp
killerwife: Priest Shadowfiend execution never goes through this code if I am not mistaken, hence the check isnt required here.

Phatcat:
Sorry, there were a few comments that didn't apply (I'm a newb, I know) - I removed them again to not cause any confusion as they added nothing to the conversation, sorry.

src/game/SpellEffects.cpp
killerwife: The delete should never occur, since its all checked in checkCast, this is simply cleaning code in case of some sort of fatal failure. (someone playing in debugger for example, or missing creature_template entry for pet already in DB)

src/game/Player.cpp
Phatcat: well, it's nothing that's gonna make or break a PR, but it would be nice to have it unified rather than having the same check placed different places for the same purpose, also makes debugging easier should there be issues in relation to it.

src/game/SpellEffects.cpp
Phatcat: Yeah, you're right, it doesn't, I realized that and removed my comment again as it brought nothing to the conversation. I wouldn't have had I known you were responding to them...

src/game/SpellEffects.cpp
Phatcat: well, if those initial checks are being done in checkcast and this is just a failsafe then I'm all for it, but shouldn't it output an error message as well, then?

src/game/Player.cpp
robinsch: I think logic wise while reading through the code it makes it a lot clearer to see checking it before the call of the function rather then in the function itself.

src/game/SpellEffects.cpp
killerwife: No because that would be against the flow of the code. We would be sending result of spell twice, once generically in Spell::cast and once in the handler. To catch an error here its too late. LoadFromDB will throw an error to the console, which is enough.

src/game/Player.cpp
killerwife: Exactly, it only influences 2 calls out of 7? I think. But as I said, this is not something utterly important.

src/game/SpellEffects.cpp
Phatcat: What? No, not send an error to the client, output a db/server error since we're not even supposed to be able to get here in the code..

killerwife: Already done.

src/game/Player.cpp
Phatcat: It's not, no, this way may also be lighter on the server actually, so nevermind my comment altogether.

robinsch:
@killerwife In which case should we not ignore dead pets. I get it for Hunter pets, you should call them even if they are dead, but isn't this a general thing? I really can't think of something that isn't temporary that should not be summonable when dead.

Phatcat:
>  I get it for Hunter pets, you should call them even if they are dead

No.

src/game/Pet.cpp
```diff
@@ -164,6 +247,16 @@ bool Pet::LoadPetFromDB(Player* owner, uint32 petentry, uint32 petnumber, bool c
         }
     }
 
+    if (!ignoreDead)
+    {
+        uint32 pet_health = fields[13].GetUInt32();
```
robinsch: you can move this outside, so you don't have to get it again fields[13].GetUInt32();

robinsch:
I don't get you. Your pet will despawn after a certain amount of time after being dead. How do you want to get it back again? You summon it and rezz it.

Phatcat:
No. you ress it and it's summoned after ress cast timer.

Phatcat:
if you try to call a dead pet you get error, at least on retail.

killerwife:
@robinsch hunter pets must not be summoned when dead thats exactly the reason why i wrote it. I can consider making the optional parameter true instead of false, but thats it. The logic is correct, after these changes

killerwife:
If it despawns you are supposed to cast revive pet directly from DB.

robinsch:
Oh wow, I didn't knew that.

Phatcat:
https://github.com/cmangos/issues/issues/1089 - It's notied somewhere in that issue with a time mark on joana's leveling video (it's from vanilla but still applies to this day to my knowledge)

Phatcat:
> I don't get you.

I think there's a lot of things you don't get to be honest.

src/game/Spell.cpp
```diff
@@ -5056,11 +5056,18 @@ SpellCastResult Spell::CheckCast(bool strict)
...
                 Creature* pet = m_caster->GetPet();
                 if (!pet)
-                    return SPELL_FAILED_NO_PET;
-
-                if (pet->isAlive())
-                    return SPELL_FAILED_ALREADY_HAVE_SUMMON;
-
+                {
+                    Pet* dbPet = new Pet;
+                    SpellCastResult result = dbPet->TryLoadFromDB((Player*)m_caster, 0);
```
killerwife: We always assume it in old LoadFromDB so yeah. We dont save creature pets since thats static (creatures dont level up).

Phatcat: unless anything except hunter players are supposed to know 982 (Revive Pet) then I think we're safe.

src/game/Pet.cpp
```diff
@@ -90,7 +90,90 @@ void Pet::RemoveFromWorld()
...
+            delete result;
+            return SPELL_FAILED_NO_PET;
+        }
+    }
+
+    uint32 pet_health = fields[13].GetUInt32();
+    if (!pet_health)
+    {
+        delete result;
+        return SPELL_FAILED_TARGETS_DEAD;
```
Phatcat: Yeah, I'm the one who put that one there originally as a placeholder since I couldn't find the right one either, should have probably added a comment in the code about that..

robinsch: ServerToClient: SMSG_PET_TAME_FAILURE (0x26DA) Length: 1 ConnIdx: 1 Time: 10/04/2016 14:34:35.156 Number: 27
Reason: Dead (10)

This is from retail.

robinsch: Can confirm this is working fine. Just ignore the "Your target is dead" message.https://i.gyazo.com/a4de5ecc4104eea230b044b886f8033a.png

killerwife: I will test it out when I am able to. Hopefully today.

killerwife: oki,

Phatcat: That's nice =)

robinsch: here is a full list of reasons, I can't confirm them for TBC. Only for WOTLK.

https://gist.github.com/robinsch/95f0527d24c38454791dff82842b13cc

Phatcat: we can test them against TBC and classic, hopefully it fits across all cores, but having it right for wrath is still better than wrong for all :+1:

killerwife: I dont see an issue with that, I will have a field day to test which are showable in tbc client. Then I will pr it.

killerwife: @robinsch one thing, next time you might not wanna post a gist with your real name, I myself do not care, but you might

killerwife:
@Phatcat @robinsch error fixed, refactor added

src/game/Spell.cpp
```diff
@@ -5082,9 +5082,8 @@ SpellCastResult Spell::CheckCast(bool strict)
                 Creature* pet = m_caster->GetPet();
                 if (!pet)
                 {
-                    Pet* dbPet = new Pet;
-                    SpellCastResult result = dbPet->TryLoadFromDB((Player*)m_caster, 0);
-                    delete dbPet;
+                    Pet dbPet;
+                    SpellCastResult result = dbPet.TryLoadFromDB((Player*)m_caster, 0);
                     if(result != SPELL_FAILED_TARGETS_DEAD)
```
Phatcat: This needs a space between if and (

also you maybe wanna rewrite it to:

`if (!dbPet.TryLoadFromDB((Player*)m_caster, 0))`

src/game/Pet.cpp
Phatcat: Also you forgot to apply the recent changes to this function.

src/game/Spell.cpp
killerwife: I really dont. That will not do what I want. but will do formatting

killerwife: WTF this happened in merge, on other repo its right

killerwife: @Phatcat it doesnt return bool

Phatcat: So maybe don't be **this much** on the defensive with me, I am only trying to help..

Phatcat: That is weird, it just did on my end, sorry.

`if (dbPet.TryLoadFromDB((Player*)m_caster, 0) != SPELL_FAILED_TARGETS_DEAD)`

would be what I would have suggested in that case instead (again, since you don't use result for anything outside the scope)

killerwife: that works

src/game/Spell.cpp
```diff
@@ -5125,15 +5124,23 @@ SpellCastResult Spell::CheckCast(bool strict)
                     if (Creature* pet = m_caster->GetPet())
                     {
                         if (!pet->isAlive() || pet->isDead()) // this one will not play along; tried and retried countless times....
```
Phatcat: probably one of these checks can be removed now it actually functions.

src/game/Pet.cpp
robinsch: @killerwife Hey, thanks for the notice. I don't care, all my changes and PR's ref to the email so, yeah I don't care either.

killerwife:
@Phatcat are we waiting for something apart from removing one of those isalive checks? Because I am done with this.

Phatcat:
Don't be like that.

I asked you directly, on IRC.

> [13:04] <Phatcat> You let me know what your pet PR is good to go, then I'll get a move on merging it into the core - the only one concern I personally have with it is still the lack of purpose for the skip bool.

you outright ignored me asking you that, and now you're asking me here what's the hold up? I'll tell you.

Your skip bool still has no purpose. Get rid of it, it's ugly.

So removing either the isAlive or the !isDead check is simply above you as you are 'done with this' as you say?

so short anseer, yes, we're waiting for something. Someone to be exact. We're waiting for YOU.

killerwife:
@Laizerox Help me out here with explaining why I am keeping skip boolean in effect handler of EffectSummonPet. Hint: SpellLog.

Phatcat:
Then give it a purpose in effectHandler, or if it already has a purpose, explain it to me;

"If you can't explain it to a six year old, you don't understand it yourself." - Albert Einstein

Laizerox:
@Phatcat the idea behind the summon handler is in future and I think its now on some sort also is that it should be generating a LogExecute (refer to my patch for it) now if you return directly in handler when summon pet from DB then proper LogExecute wont be generated or sent for player ever.

Phatcat:
Unless someone else besides me has anything to note, I am content with it in it's current state so I am done complaining at least and will look into merging it very soon.

cyberium:
Ok let me do it then because lot of stuff will change around those code in upcoming charm code. I'll do it right after it. If you merge it before i will be forced to rebase my patch :/

Phatcat:
I was wondering if it was in conflict with your upcoming work. Well, you're gonna have to rebase killerwife's work most likely then but roger that, it's all yours.

killerwife:
Ok then that means I am done here.

cyberium:
Its never done you can continue to if you see something. Even when this will be merged.

src/game/Spell.cpp
```diff
@@ -5412,6 +5424,29 @@ SpellCastResult Spell::CheckCast(bool strict)
...
+            {
+                ((Player*)m_caster)->SendPetTameFailure(PETTAME_ANOTHERSUMMONACTIVE);
+                return SPELL_FAILED_DONT_REPORT;
+            }
+            else
+            {
+                Pet pet;
+                if (pet.TryLoadPetFromDB((Player*)m_caster, 0) == SPELL_FAILED_TARGETS_DEAD)
+                {
+                    ((Player*)m_caster)->SendPetTameFailure(PETTAME_ANOTHERSUMMONACTIVE);
```
Fabi: and here?

cyberium:
merged thanks @killerwife

PR #133, opened 2016-10-03
Implement proc system flags

killerwife:
PROC_FLAG_DONE_SPELL_NONE_DMG_CLASS_POS
PROC_FLAG_TAKEN_SPELL_NONE_DMG_CLASS_POS
PROC_FLAG_DONE_SPELL_NONE_DMG_CLASS_NEG
PROC_FLAG_TAKEN_SPELL_NONE_DMG_CLASS_NEG
Rename
PROC_FLAG_DONE_SPELL_MAGIC_DMG_CLASS_POS
PROC_FLAG_TAKEN_SPELL_MAGIC_DMG_CLASS_POS
PROC_FLAG_DONE_SPELL_MAGIC_DMG_CLASS_NEG
PROC_FLAG_TAKEN_SPELL_MAGIC_DMG_CLASS_NEG

Can be expanded upon in the future. Enables correct handling of things
like Sap interrupting Stealth

Also https://gist.github.com/killerwife/35f9ce64842c4a10337cc01166479718 needs to be added to core SQL. (but apparently that requires a demonic sacrifice of some sort, so I am gonna leave that to the unfortunate reviewer)

### DIFF
--- a/src/game/Player.cpp
+++ b/src/game/Player.cpp
@@ -14811,6 +14811,7 @@ void Player::_LoadAuras(QueryResult* result, uint32 timediff)
                     holder->SetTrackedAuraType(TRACK_AURA_TYPE_NOT_TRACKED);
 
                 AddSpellAuraHolder(holder);
+                holder->SetState(SPELLAURAHOLDER_STATE_READY);
                 DETAIL_LOG("Added auras from spellid %u", spellproto->Id);
             }
             else
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -694,8 +694,16 @@ void Spell::prepareDataForTriggerSystem()
         default:
             if (IsPositiveSpell(m_spellInfo->Id))           // Check for positive spell
             {
-                m_procAttacker = PROC_FLAG_SUCCESSFUL_POSITIVE_SPELL;
-                m_procVictim   = PROC_FLAG_TAKEN_POSITIVE_SPELL;
+                if (m_spellInfo->DmgClass & SPELL_DAMAGE_CLASS_NONE) // if dmg class none
+                {
+                    m_procAttacker = PROC_FLAG_DONE_SPELL_NONE_DMG_CLASS_POS;
+                    m_procVictim = PROC_FLAG_TAKEN_SPELL_NONE_DMG_CLASS_POS;
+                }
+                else
+                {
+                    m_procAttacker = PROC_FLAG_DONE_SPELL_MAGIC_DMG_CLASS_POS;
+                    m_procVictim = PROC_FLAG_TAKEN_SPELL_MAGIC_DMG_CLASS_POS;
+                }
             }
             else if (m_spellInfo->HasAttribute(SPELL_ATTR_EX2_AUTOREPEAT_FLAG))   // Wands auto attack
             {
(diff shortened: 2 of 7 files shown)


src/game/Player.cpp
```diff
@@ -14811,6 +14811,7 @@ void Player::_LoadAuras(QueryResult* result, uint32 timediff)
                     holder->SetTrackedAuraType(TRACK_AURA_TYPE_NOT_TRACKED);
 
                 AddSpellAuraHolder(holder);
+                holder->SetState(SPELLAURAHOLDER_STATE_READY);
```
Warlockbugs: This line fixes issues with Combustion talent after logout.

killerwife:
@robinsch Stealth has two ways of removal, AuraInterruptFlags and ProcFlags. Stealth had a proc flag AOE_PROC which was supposed to be procced by Sap, which as you can see, is bonkers. Thats why I did a little digging, turns out its not an AOE_PROC flag, but a NON-DAMAGE spell proc flag. Most of the code is not related to Sap triggering Stealth proc, but it promptly shows what the proc flag should actually do.

To answer your classic question, I am unaware how classic handles procs, because Stealth does not have proc flags in vanilla DBC (I looked at QSW). Maybe stealth proc flags were serverside in classic? Dunno, it might even be handled completely differently.

Also the patch fixes proccing of saved auras on login.

robinsch:
If you are refering to Rogue A Sap's Rogue B. Rogue B stealth should be removed then I wouldn't handle this at all this way. All harmful spells should remove Stealth (exception MD, Distract and Earthbind Totem, added in 2.2(?))

killerwife:
To my current knowledge, stealth has "all harmful spells" proc flags, hence this is the way to handle it. As of now, I am unaware of any removal rules apart from AuraInterruptFlags or ProcFlags, (and maybe some parent-child spells) that were used by blizzard.

robinsch:
Some input that (https://github.com/TrinityCore/TrinityCore/commit/5f24fd6c51350b391bb30b53bd813c1ec25f3248) is the way to go.

Apply a DOT to a stealthed target.
Apply the arena shadow sight to a stealthed target.
Any non damageing spell to a stealthed target.

Laizerox:
Stealth (Rank 4) for example has AuraIF: 0x00003C06
Which means following:
- AURA_INTERRUPT_FLAG_DAMAGE
- AURA_INTERRUPT_FLAG_CAST
- AURA_INTERRUPT_FLAG_TALK
- AURA_INTERRUPT_FLAG_USE
- AURA_INTERRUPT_FLAG_MELEE_ATTACK
- AURA_INTERRUPT_FLAG_UNK13

killerwife:
@robinsch I still think you are wrong, BUT I will give you the benefit of the doubt and I will look at the three cases you described. If I apply arena shadow sight to target, stealth should break? Sap falls into non damaging category, and I already told you proc flags are the way to go there. You just need to configure it properly.

robinsch:
@Laizerox Only damage flag is harmful. I think everybody can use spellwork here.

@killerwife Shadow Sight should remove Stealth. When I've worked with mangos there were no "apply aura internal flag". There was a case the proc system returned somewhere because no damage was done and that why nothing proc'ed. This also resulted in a issues where trinkets couldnt proc on druids because they we're applying hots that did no healing at apply.

killerwife:
@robinsch Nowadays procs happen if damage wasnt done, hence you enter proc system on something like Sap correctly. It is all a matter of configuring SpellFamilyMask of Stealth procs (which is serverside on blizzard)

robinsch:
@killerwife Oh my bad, then ignore all my comments. Do you know when this was changed? I would like to take a look at it because I fixed it with a special internal proc flag (was like 2.5y ago).

killerwife:
@robinsch this line procs with sap

https://github.com/cmangos/mangos-tbc/blob/master/src/game/Spell.cpp#L1071

PR #134, opened 2016-10-09
Sd2improvements

killerwife:
Added two parameters to DamageTaken and DamageDealt to enable scripting of special events.

Event A) Shade of Aran must only break drinking on direct damage.
Event B) Gallywix during quest script, must always take 50% of health from Ambushes.

EAI part for testing of q.2478 https://gist.github.com/killerwife/4bf3a68a46796f904e65f659bf81ec7e will be added to tbc-db when PR goes through.

@xfurry @cyberium Please add this as soon as possible :)


cyberium:
merged thanks

PR #135, opened 2016-10-14
Fix showing proper quest giver marks

killerwife:
Previously if a quest giver had non-daily/weekly repeatable quests, a daily exclamation mark was shown above his head. Correct behaviour is to show blue question mark in case of such quests. Example is commendations questgiver in ironforge near middle.

Might need testing/changing(or not apply at all) for classic branch, as there are no daily quests there.

### DIFF
--- a/src/game/QuestDef.h
+++ b/src/game/QuestDef.h
@@ -101,12 +101,12 @@ enum QuestStatus
 enum __QuestGiverStatus
 {
     DIALOG_STATUS_NONE                     = 0,
-    DIALOG_STATUS_UNAVAILABLE              = 1,
-    DIALOG_STATUS_CHAT                     = 2,
-    DIALOG_STATUS_INCOMPLETE               = 3,
-    DIALOG_STATUS_REWARD_REP               = 4,
-    DIALOG_STATUS_AVAILABLE_REP            = 5,
-    DIALOG_STATUS_AVAILABLE                = 6,
+    DIALOG_STATUS_UNAVAILABLE              = 1,             // Grey Exclamation Mark
+    DIALOG_STATUS_CHAT                     = 2,             // No marker
+    DIALOG_STATUS_INCOMPLETE               = 3,             // Grey Question Mark - quest taken
+    DIALOG_STATUS_REWARD_REP               = 4,             // Blue Question Mark - non-daily repeatable available
+    DIALOG_STATUS_AVAILABLE_REP            = 5,             // Blue Exclamation Mark - daily available
+    DIALOG_STATUS_AVAILABLE                = 6,             // Yellow Exclamation Mark - quest available
     DIALOG_STATUS_REWARD2                  = 7,             // no yellow dot on minimap
     DIALOG_STATUS_REWARD                   = 8,             // yellow dot on minimap
     DIALOG_STATUS_UNDEFINED                = 100            // Used as result for unassigned ScriptCall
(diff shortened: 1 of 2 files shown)


Phatcat:
Possibly also look at https://github.com/cmangos/issues/issues/1098
> 2:12
**The quest to turn in 'Shaman Training' has a blue question mark in the gossip menu while it should be yellow.**
(ref unified-db/Database#251 (comment))

and https://github.com/classicdb/database/issues/899
> 3:30:21
Kor'ghan is supposed to go straight to the quest and skip the gossip menu, on cmangos it shows the default one.
**The question mark above his head should not be blue, but rather yellow ...**
He is supposed to do both /talk on initial contact and /talk when you press continue
He should also have a follow-up quest, on cmangos he doesn't...

Pretty please? :3

killerwife:
Gossip menu: nothing to do with me, thats a general issue elsewhere.
Korghan gives a repeatable quest afaik, and at least in TBC there is no yellow question mark for repeatables. Shaman Training has issue elsewhere, i fixed marks over head. Will look if it happens on tbc later today.

killerwife:
Shaman Training is somehow related to wotlk client, there is a new quest flag and I have no clue how it interacts with client.

Finding the Antidote from Korghan is a repeatable quest. In TBC and WOTLK repeatable non daily/weekly quests have ALWAYS blue question mark. If you find evidence to whether this didnt change in TBC/WOTLK, I will make it work so that this repeatable quest shows yellow question mark as well. (thats when blue question mark was used according to gamepedia). If it did change, then this is a vanilla bug only. Currently there is very little to no evidence on whether a repeatable quest has yellow question mark under certain circumstances as well. This would be an easy change, but I need evidence.

Phatcat:
actually, that "Finding the Antidote from Korghan" - may be fine on tbc/wrath but it's bugged on classic (evident by joanas leveling video). I just thought that since it was bugged there that there may be something up with it for tbc as well, guess not.

killerwife:
Nah, this is one of those things that vid of one expansion doesnt make sure its that way for another.

Phatcat:
Yeah, I can respect that totally, but..... it makes no sense that that quest would be repeatable, though, not even in later expansion. It's a leveling quest. It makes no sense to be able to repeat.

killerwife:
Shaman training is not repeatable, the Korghan quest is. Shaman training has some sort of "starting zone quest" flag I know nothing about.

Phatcat:
Yes. that's the one I'm talking about. I know shaman training shouldn't be repeatable. but really. Have you played the mission Finding the antidote? cause I'm telling you, it would make no sense that that quest would be repeatable.

Phatcat:
https://www.youtube.com/watch?v=Z4IT4O6S1wo <- it's not repeatable in 2010 either. Having a hard time believing it started out not being repeatable, then was repeatable for tbc and wrath, then stopped being repeatable again.

killerwife:
You cant actually see the question mark be removed from the quest giver, which is what is most important. If it doesnt get removed, then its repeatable with yellow question mark, if it gets removed, its repeatable with special condition. Try to find out if it gets removed from the quest giver after finishing quest. If yes, well then we have a problem cos we need conditional quest showing, and we dont support that.

Phatcat:
Sure, I can try to look out for that, my initial search didn't give me much to go by besides that one very poorly rated video there.

Phatcat:
If the top comment here is to be believed; http://www.wowhead.com/quest=813/finding-the-antidote

then the yellow exclaimation mark we see appearing over his head in the joana's leveling video is actually the quest becoming available again; so it's a repeatable with yellow markers which is only available when you're on this quest; http://www.wowhead.com/quest=812/need-for-a-cure

namreeb:
Can you add a commit to this PR which fixes the compile?

killerwife:
@namreeb should build now, was related to an unmerged PR which was merged between now and then

PR #137, opened 2016-10-23
Refactor and add usage of sSpellScriptTargetStorage

killerwife:
TARGET_RANDOM_ENEMY_CHAIN_IN_AREA
TARGET_RANDOM_FRIEND_CHAIN_IN_AREA
TARGET_RANDOM_UNIT_CHAIN_IN_AREA
now support spell_script_target table entries

Enables
/*
http://jira.vengeancewow.com/browse/TBC-1174
Nightlash hitting players with Spirit Steal
*/

DELETE FROM spell_script_target WHERE entry = 3477;
INSERT INTO spell_script_target (entry, `type`, targetEntry, inverseEffectMask) VALUES
(3477, 1, 1772, 0),
(3477, 1, 1773, 0);

### DIFF
--- a/src/game/Spell.cpp
+++ b/src/game/Spell.cpp
@@ -1606,6 +1606,46 @@ void Spell::SetTargetMap(SpellEffectIndex effIndex, uint32 targetMode, UnitList&
             if (tempTargetUnitMap.empty())
                 break;
 
+            SQLMultiStorage::SQLMSIteratorBounds<SpellTargetEntry> bounds = sSpellScriptTargetStorage.getBounds<SpellTargetEntry>(m_spellInfo->Id);
+            if (bounds.first != bounds.second)
+            {
+                for (UnitList::const_iterator iter = tempTargetUnitMap.begin(); iter != tempTargetUnitMap.end();)
+                {
+                    bool found = false;
+
+                    if ((*iter)->GetTypeId() == TYPEID_UNIT)
+                    {
+                        for (SQLMultiStorage::SQLMultiSIterator<SpellTargetEntry> i_spellST = bounds.first; i_spellST != bounds.second; ++i_spellST)
+                        {
+                            if (i_spellST->CanNotHitWithSpellEffect(effIndex))
+                                continue;
+
+                            // only creature entries supported for this target type
+                            if (i_spellST->type == SPELL_TARGET_TYPE_GAMEOBJECT)
+                                continue;
+
+                            if ((*iter)->GetEntry() == i_spellST->targetEntry)
+                            {
+                                if (i_spellST->type == SPELL_TARGET_TYPE_DEAD && ((Creature*)(*iter))->IsCorpse())
+                                {
+                                    found = true;
+                                }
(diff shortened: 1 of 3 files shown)


xfurry:
I disagree.
I don't see any reason why TARGET_RANDOM_.... should have script targets.
This quest seems to rather be an exception, and I don't support adding script targets to these spell just for one scenario.
We should research what is actually happening in the script, and I'm pretty sure we'll find a proper solution.
Most of the times, these creatures, use an AoE dummy effect spell, which already supports script target, to select a target and then triggers the actual spell on the selected target. This happens all the time, and it's probably the same scenario.

killerwife:
First of all, the spell is supposed to hit only gnolls. Second of all, there are spells that are single target and are supposed to support only a set of specific entry targets, which supports the idea that blizz is able to script targeting on every spell. For me personally it doesnt matter, I can use my spell scripting for this very example.

Whichever way you wanna go is up to you, I added this as a "generic" solution, and I do not see why it would be bad. Either way you need to restrict this one specific spell to only target gnolls.

ItsmeAga:
I think the TARGET_4 which spirit steal has means Target nearby ally(based on factions probably), so it would be just a combination of the target mode and the faction.

killerwife:
target random nearby ally is target 3, target 4 is target random nearby unit. So the REAL question is, do we want this to support spell_script_target, or change target type in spell.dbc in future :)

xfurry:
@killerwife I think we need to do some more research here, before we jump to the conclusion.
When I implemented script target for TARGET_NARROW_FRONTAL_CONE I had a lot of evidence that these kind of spells are actually using specific entries to acquire the targets. These involved many sniffs from various scripted events from 3.x and 4.x

I think that TARGET_RANDOM_UNIT_CHAIN_IN_AREA might be reasonable to support script target, as this kind of target mode is faction independent.
But for TARGET_RANDOM_ENEMY_CHAIN_IN_AREA and TARGET_RANDOM_FRIEND_CHAIN_IN_AREA I don't see any reason. These are specifically used to select a friend or enemy in area, regardless of the typeId of the object (unit or player). And there are quite a lot of spells with these target modes. So I think someone would have noticed the need for a script target before, if it was required.

killerwife:
@xfurry i added for all three so that the code is as tidy and compact as possible. (design decision) If the table entries are not set we evaluate one condition, its hardly a problem to not merge db code that would be considered bad. You can already do that on 20 places. But as I said, there are more solutions to the same issue, and on my end it is done since all three were mentioned. I simply dont care which one will be utilized, simply decide and I will use that one. I PRed this one because I found it the most generic and consistent with decisions in other places of the core. Amen.

xfurry:
Fair enough. :smile: 
Let me look at this in detail and I will get to a conclusion.
Will get back to you later / tomorrow.

ItsmeAga:
spell 7103 also has target 4 and the tooltip says "10% chance to spread contagion to allies when hit". Also the other spells which use target 4 all are of the "wandering plague" kind, where you can infest your allies around you, but it won't jump to hostile ones

PR #138, opened 2016-10-24
Fix and cleanup RollMeleeOutcomeAgainst

killerwife:
@cyberium

### DIFF
--- a/src/game/Unit.cpp
+++ b/src/game/Unit.cpp
@@ -1617,7 +1617,9 @@ void Unit::CalculateMeleeDamage(Unit* pVictim, CalcDamageInfo* damageInfo, Weapo
             }
 
             // calculate values
-            int32 diff = damageInfo->target->GetDefenseSkillValue() - GetWeaponSkillValue(damageInfo->attackType);
+            int32 attackerRating = GetWeaponSkillValue(damageInfo->attackType);
+            attackerRating = attackerRating > getLevel() * 5 ? getLevel() * 5 : attackerRating; // cap rating at level * 5, done in 2.0.1
+            int32 diff = damageInfo->target->GetDefenseSkillValue() - attackerRating;
             float lowEnd  = baseLowEnd - (0.05f * diff);
             float highEnd = baseHighEnd - (0.03f * diff);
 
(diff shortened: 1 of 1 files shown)


src/game/Unit.cpp
```diff
@@ -2393,10 +2393,10 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(const Unit* pVictim, WeaponAttackT
 
     if (miss_chance > 0)
     {
-        tmp = miss_chance - skillBonus;
-        if (roll < (sum += tmp))
+        miss_chance -= skillBonus;
```
Phatcat: so a miss chance of 0 in all cases. that's nice.

src/game/Unit.cpp
```diff
@@ -2470,25 +2473,44 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(const Unit* pVictim, WeaponAttackT
...
-            tmp = parry_chance - skillBonus;
+            parry_chance -= skillBonus;
 
             // Reduce parry chance by attacker expertise rating
             if (GetTypeId() == TYPEID_PLAYER)
-                tmp -= int32(((Player*)this)->GetExpertiseDodgeOrParryReduction(attType) * 100);
+                parry_chance -= int32(((Player*)this)->GetExpertiseDodgeOrParryReduction(attType) * 100);
 
-            if (roll < (sum += tmp) && (pVictim->GetTypeId() == TYPEID_PLAYER
-                || !(((Creature*)pVictim)->GetCreatureInfo()->ExtraFlags & CREATURE_EXTRA_FLAG_NO_PARRY)))
```
Phatcat: ´Why did you remove the extra flag from here?

src/game/Unit.cpp
```diff
@@ -2499,54 +2521,34 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(const Unit* pVictim, WeaponAttackT
...
-
-        if (roll < (tmp = (tmp > 4000) ? 4000 : tmp))
-        {
-            DEBUG_FILTER_LOG(LOG_FILTER_COMBAT, "RollMeleeOutcomeAgainst: GLANCING <%d, %d)", sum - tmp, tmp);
-            return MELEE_HIT_GLANCING;
-        }
-    }
-
-    if (crit_chance > 0)
+    if (crit_chance > 0 && roll < (sum += crit_chance))
```
Phatcat: Why did you remove skill bonus from crit chance?

Phatcat: I'd ask why that was being handled in unit but that is down to me for not having checked there before applying the other patch. This may have affected crit chance, indeed. but not in the volumes you speak of.

Phatcat:
If it really matters that much to you to have mobs glancing even before they get the chance to block then sure, we can move it back. also I see the calculation for glancing indeed should be * 100, not * 200, you're right, I was looking at the original code in classic where I did the PR and the calculation is slightly different there; it seems to have a 2 jammed in there for some unnamed reason.

Changes proposed in the PR look fine to me.

ItsmeAga:
Since it is a cleanup you could remove the MELEE_HIT_BLOCK_CRIT part aswell, there are nor crit blocks in TBC and it is never used so it only reduces your actual blockchance by the critchance

killerwife:
@ItsmeAga I am adding that removal to the PR right away :) thx

Phatcat:
that 2 it had jammed in there wasn't for no reason;
http://wowwiki.wikia.com/wiki/Glancing_blow
In classic the chance for landing a glancing was 10 + 2% per skill difference, like my calculation is, not 10 + 1% per skill difference as it was changed to for TBC, as your calculation is.

Phatcat:
Merged.

PR #139, opened 2016-10-26
General charm/pet fixes

cyberium:
- Now creatures can be charmed.
- Now players can be charmed.
- We can now charm/possess even if we have already a summoned pet.
- Charmed pet have their own threatlist. But their original one before they were charmed is preserved.
- Charmer have now attack spell in pet bar even if he charmed a player.
- Its now possible to charm a player while dueling.
- Charmed players/creature will not be teleported if areatrigger is triggered.
- Uncharm before telport.
- We can now stable unsumoned pet.
- If we are in flight while loading the pet will be set as temporary unsumoned and will be correctly resummoned at landing.
- Priest pet is now not saved to db.
- Its now not possible to summon/stable a dead pet. If its not summoned we are now able to revive it from DB.
- Client error message are improved in different case like trying to summon a dead pet.
- Now in combat state is applied only after spell hit. (before it was impossible to cast a spell with not in combat attribute)

With @Phatcat help and some @killerwife commits. Refer -> https://github.com/cmangos/mangos-classic/pull/210 for better commit history


Lillecarl:
It's alot for one commit, but i trust you on this one! ;)

cyberium:
Never trust anyone. Check it and verify it work then report new issue! :)

xfurry:
@cyberium code and overall design looks very good at a first glance.
Main part, is that we need to test it thoroughly.

ItsmeAga:
Should be split into multiple commits, also it contains parts of https://github.com/cmangos/mangos-tbc/pull/131

src/game/Unit.cpp
```diff
@@ -8593,18 +8594,50 @@ void Unit::CleanupsBeforeDelete()
 
 CharmInfo* Unit::InitCharmInfo(Unit* charm)
 {
-    if (!m_charmInfo)
-        m_charmInfo = new CharmInfo(charm);
+    if (m_charmInfo)
```
stfx: redundant null check before delete, except if you have it there to prevent double deletes but this is usually bad design: http://stackoverflow.com/questions/615355/is-there-any-reason-to-check-for-a-null-pointer-before-deleting

> setting the pointer explicitly to null after deleting might actually make sense depending on the architecture of the code, precisely to prevent double-deletes. Since this only saves double-deletes on the same pointer though, it's highly questionable whether this is a good design. I wouldn't ever rely on it.

killerwife:
1. Cant unstable pet when mounted.
2. Cant switch pets when mounted.
3. If you mount, stable pet, unmount and teleport away, pet is wrongly summoned to your side.
4. Bit more detailed pet duplication:
a) have one pet summoned, one pet in stables
b) mount and stable pet
c) unmount, unstable pet you stabled
d) pet is duplicated and shown in window twice

killerwife:
1. Cannot move pet to slot #2 when having one pet, both when mounted and unmounted
2. Pet scale is wrong, they are small as hell
3. Pet weirdly behaves in stables, if it was in slot 2 and u unstabled slot 1 it moved to slot 1 from slot 2. (http://imgur.com/a/GnyFl)
4. It is possible to permanently lose pet by following:
A) Tame pet
B) Dismiss it
C) Tame a new one without moving the old one in stables - this should throw error message
@cyberium happy hunting

Phatcat:
number 4 has been there since forever. not so much pet as it is spell related. its an annoying bug none the less but it has to be fixed in checkcast or something like that.

ItsmeAga:
I'd suggest using a FSM(finite state machine) to get a safer design so things like #4 are not possible

killerwife:
Number 4 ties into the same logic we are fixing at stable master, else I would not be mentioning it.

cyberium:
1-3 are same: i faced that but didnt give much attention to it mainly because lack of knowledge of how it should work. So Thank for the input, those issues will be adressed.
2: its not relevant for this PR, i dont know much about the corect scale i have to apply but i think @Phatcat already worked on it right? I can confirm they seem a bit too little at low level and maybe a bit too big in high level. That should be a matter of tweaking.
4: Not directly related to this PR too but nice found. It should be fixable as @Phatcat write by adding some condition in checkcast.

Many Thanks for those ingame feedback.
That said i know about more problem related to charm/Possess but it seem no one care more about them except @Phatcat so i can conclude its good enough to be pushed after i'll fix last above issues.

killerwife:
I have been struggling to get any free time from fixing consequences of my aura reshuffle. As of now, that seems to be done.

Also the checkcast condition is related since you took my PR and incorporated it into it. I will write the condition today and post it here, for you to add it as well.

Phatcat:
I never directly worked with the scaling, but I know where it's applied; it's applied in Pet::InitstatsForLevel; it shouldn't be too much of an issue tinkering with those values, finding the right one however... I just frankly didn't know we had an issue with that.

ItsmeAga:
hunter pets stats are correct, warlock pets not they got individual stamina to health ratios and such things

killerwife:
Afaik pets dont account stats into hp/mana as of now.

killerwife:
@cyberium fix for number 4 https://gist.github.com/killerwife/c75d53f20d8c0d463cd046138e206b5d

I am unsure about the error message, but the rest is tested.

Phatcat:
InitStatsforLevel is the name of the function, it deals with other things that just 'stats';
https://github.com/cmangos/mangos-tbc/blob/master/src/game/Pet.cpp#L1121

Phatcat:
> fix for number 4 https://gist.github.com/killerwife/c75d53f20d8c0d463cd046138e206b5d

not enough to change else if to;
`else if (Pet::TryLoadFromDB((Player*)m_caster) != SPELL_FAILED_NO_PET)` ?

Also could the error message be
SPELL_FAILED_ALREADY_HAVE_SUMMON ?

killerwife:
bad targets and no pet must result in tame pet going through, whereas pet dead and spell cast ok must result in tame pet not going through, so the answer is no

Phatcat:
what are you talking about?

your proposed check checks if the pet is either dead or there is a pet, then you opt for the error
my proposed checks if there is a pet to load at all, if there is then it opts out for the error.

can you give any examples of the hunter having a pet and still being allowed to tame a new?

i dont see how your check has anything to do with bad targets

killerwife:
Eh you need to look at Pet::TryLoadDB to understand what I wrote. In there you will see what BAD_TARGETS means. Also you need to account for possibility of DB having wrong pet saved, temporary and such, since this condition is run on vanilla tame spells which are charms, since @cyberium is using the attribute. Hence back to my previous statement.

Phatcat:
we dont (or at the very least shouldnt) save temp pets, and i did look at the tryloadfromdb, where do you think i got the results from?

okay, tame spell is not charm
taming rod is charm....

killerwife:
SPELL_ATTR TAME_PET search for spells with this, you will find apart from tame pet that they are charms

Phatcat:
conversation derailed; what you're saying is no argument to keep your check, its not even a counter argument to doing my check, it's simply semantics. I don't care, but get your facts straight;

Hunter pet != charm. Tame beast != taming rod. Tame beast != charm.

killerwife:
Please go check what fulfils HasAttribute(SPELL_ATTR_EX2_TAME_BEAST)

Phatcat:
I don't care that it has one of the same spell attributes, sharing of attributes does not make equal.

If what you're saying is that both taming rod and tame beast go through the same spell checkcast then you have a very complex way of saying such, but still that would not make a single difference to anything that I am saying. Exactly how does that attribute invalidate anything I've said or the code change I proposed?

killerwife:
‎That the attribute is used directly above my changed code, which you did not bother to look. If tryloadfromdb fails with bad targets it suggests wrong data in db, hence tame pet should succeed. I suggest you go ingame and test all eventual possible outcomes first, instead of trying to make the code only shorter. When you define what bad targets outcome does in core (when playing) then come back and explain to me why I didnt include it in the condition the way you suggest. Amen.

Phatcat:
Still, that attribute has nothing what so ever to do with anything in this context, you are right I didn't quite understand the different results from the TryLoadFromDB; it's not like the result is exactly obvious by what it means, and you're right, no I didn't study your code, but you do not know what I 'bother to look at' or not.

I was not trying to make the code shorter, but simpler; you however seem to have a real issue with every time I suggest any kind of simplification, fitting or not.

> When you define what bad targets outcome does in core (when playing) then come back and explain to me why I didnt include it in the condition the way you suggest. Amen.

I'm gonna ask you nicely never to speak to me like that again.

I asked you to justify and explain your code changes. I very much expect a straight-forward and non condasending answer next time.

when I ask stuff such as this

> can you give any examples of the hunter having a pet and still being allowed to tame a new?

and get answered this:

> Eh you need to look at Pet::TryLoadDB to understand what I wrote. In there you will see what BAD_TARGETS means.

when you could have simply given me the direct answer you did later;

> If tryloadfromdb fails with bad targets it suggests wrong data in db, hence tame pet should succeed.

perhaps you should stop to think that maybe I did look at the code, and maybe I didn't fully understand, or I wouldn't be confused about it.

cyberium:
it seem that for 1 and 3 its not possible/normal.
Client doesnt send anything when we try to switch pet from slot1 to slot2.
When you unstable a pet and the pet from slot2 switch to slot1 it seem normal too and only visual client side.
So i dont know. For me it look like its correct.

killerwife:
from slot 1 to slot 2 is indeed not possible. But the other interactions should be working i think, as right now they are counter intuitive. But I digress :)

cyberium:
merged thanks to all

PR #142, opened 2016-11-11
Added/Fixed texts for TBC dungeons

Qingle:
Also added notes.

I look to update all of TBC and then Vanilla shortly after.


Qingle:
Nevermind I'm a noob, will do a full pull when it's finished

PR #143, opened 2016-11-22
Make all projects compile with MP and add EditAndContinue build config

killerwife:
@namreeb


killerwife:
@namreeb doesnt wanna add another property sheet when we will switch to cmake in the future. I think that was what he said.

namreeb:
I think we agreed on IRC to postpone this for now, until VS2017 hits.

PR #145, opened 2016-11-25
Fix compilation on Mac and implement CI for Mac

Lillecarl:
Might use some find tuning and standardising, but it's definitely an improvement over what we got atm.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,16 @@ if(UNIX)
     find_package(MySQL REQUIRED)
   endif()
 
+  # If OpenSSL path isn't specified on mac we set the one that homebrew uses
+  # since that's what most people will be using.
+  if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    if (NOT OPENSSL_ROOT_DIR)
+      set(OPENSSL_ROOT_DIR /usr/local/opt/openssl/)
+    endif()
+    if (NOT OPENSSL_INCLUDE_DIR)
+      set(OPENSSL_INCLUDE_DIR /usr/local/opt/openssl/include)
+    endif()
+  endif()
   find_package(OpenSSL REQUIRED)
   find_package(ZLIB REQUIRED)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,7 +129,7 @@ if(UNIX)
 
   # If OpenSSL path isn't specified on mac we set the one that homebrew uses
   # since that's what most people will be using.
-  if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  if (APPLE)
     if (NOT OPENSSL_ROOT_DIR)
       set(OPENSSL_ROOT_DIR /usr/local/opt/openssl/)
     endif()
--- a/contrib/extractor/System.cpp
+++ b/contrib/extractor/System.cpp
@@ -89,7 +89,7 @@ const char* CONF_mpq_list[] =
     "patch-5.MPQ",
 };
 
-static char* const langs[] = {"enGB", "enUS", "deDE", "esES", "frFR", "koKR", "zhCN", "zhTW", "enCN", "enTW", "esMX", "ruRU" };
+static char const* langs[] = {"enGB", "enUS", "deDE", "esES", "frFR", "koKR", "zhCN", "zhTW", "enCN", "enTW", "esMX", "ruRU" };
 #define LANG_COUNT 12
 
 void CreateDir(const std::string& Path)
@@ -190,7 +190,7 @@ uint32 ReadMapDBC()
         map_ids[x].id = dbc.getRecord(x).getUInt(0);
         strcpy(map_ids[x].name, dbc.getRecord(x).getString(1));
     }
-    printf("Done! (%u maps loaded)\n", map_count);
+    printf("Done! (%u maps loaded)\n", std::uint32_t(map_count));
     return map_count;
 }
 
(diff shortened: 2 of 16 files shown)


CMakeLists.txt
```diff
@@ -127,6 +127,16 @@ if(UNIX)
     find_package(MySQL REQUIRED)
   endif()
 
+  # If OpenSSL path isn't specified on mac we set the one that homebrew uses
+  # since that's what most people will be using.
+  if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
Lillecarl: Might wanna change this to:

IF(APPLE)

Lillecarl:
I guess I could've left the warning fixes out, but now they're there, and it's better that they're there then nowhere.

Lillecarl:
Okay idk what I've missed from "b16" but I'll figure it out at some point...

Lillecarl:
This closes cmangos/issues#1153

Blumfield:
The king has return, welcome back to your home

Lillecarl:
I don't think I deserve being called a king, but thanks!

contrib/vmap_extractor/vmapextract/mpq_libmpq.cpp
```diff
@@ -82,7 +82,7 @@ size_t MPQFile::read(void* dest, size_t bytes)
     if (eof) return 0;
 
     size_t rpos = pointer + bytes;
-    if (rpos > size)
```
cyberium: ??

contrib/vmap_extractor/vmapextract/model.cpp
```diff
@@ -24,7 +24,7 @@
 #include <algorithm>
 #include <cstdio>
 
-Model::Model(std::string& filename) : filename(filename), vertices(0), indices(0)
+Model::Model(std::string& filename) : vertices(0), indices(0), filename(filename)
```
cyberium: ??

contrib/vmap_extractor/vmapextract/loadlib/loadlib.h
```diff
@@ -19,31 +19,16 @@
...
-typedef int64_t            int64;
-typedef int32_t            int32;
-typedef int16_t            int16;
-typedef int8_t             int8;
-typedef uint64_t           uint64;
-typedef uint32_t           uint32;
-typedef uint16_t           uint16;
-typedef uint8_t            uint8;
-#endif
+#include <cstdint>
```
cyberium: I think this may broke other kind of build like powerpc

contrib/vmap_extractor/CMakeLists.txt
```diff
@@ -19,6 +19,14 @@ ADD_DEFINITIONS("-Wall")
 ADD_DEFINITIONS("-ggdb")
 ADD_DEFINITIONS("-O3")
 
+if (APPLE)
```
cyberium: This worked before for other apple so this mean it may broke them now. I insist on the "may" word. No idea if it will realy or not.

contrib/extractor/System.cpp
```diff
@@ -89,7 +89,7 @@ const char* CONF_mpq_list[] =
     "patch-5.MPQ",
 };
 
-static char* const langs[] = {"enGB", "enUS", "deDE", "esES", "frFR", "koKR", "zhCN", "zhTW", "enCN", "enTW", "esMX", "ruRU" };
+static char const* langs[] = {"enGB", "enUS", "deDE", "esES", "frFR", "koKR", "zhCN", "zhTW", "enCN", "enTW", "esMX", "ruRU" };
```
cyberium: this realy fix something?

contrib/vmap_extractor/vmapextract/loadlib/loadlib.h
Fabi: who cares about powerpc?

src/game/SpellMgr.h
```diff
@@ -1135,6 +1136,7 @@ inline bool IsSpellSpecificUniquePerTarget(SpellSpecific specific)
         case SPELL_DRINK:
         case SPELL_FOOD_AND_DRINK:
             return true;
+        default: break;
```
Fabi: Then it should be a bigger PR with adding a default where it is missing :D

src/tools/genrevision/genrevision.cpp
```diff
@@ -133,7 +133,7 @@ bool extractDataFromGit(std::string filename, std::string path, bool url, RawDat
         else
             strcpy(data.rev_str, hash_str);
     }
-    else if (entriesFile = fopen((path + ".git/HEAD").c_str(), "r"))
+    else if ((entriesFile = fopen((path + ".git/HEAD").c_str(), "r")))
```
Lillecarl: A Clang warning.

Lillecarl:
I added a few clang warning fixes that are not appreciated, PR invalid.

Lillecarl:
To make it clear: I'm redoing it from my Windows machine, more used to that still =)

contrib/vmap_extractor/vmapextract/model.cpp
evil-at-wow: Compiler warnings again. Class members are initialized in the order they are declared, no matter what order you're using in the definition. Some compilers, like GCC and clang, are warning you that the order you've written the members in, is not what you actually get.

contrib/vmap_extractor/vmapextract/loadlib/loadlib.h
evil-at-wow: It's a C++ standard library header since C++ 11 so this should be fine for any modern PowerPC compiler. Also, cala's not using PowerPC but an Intel based Mac so no worries there.

contrib/vmap_extractor/CMakeLists.txt
Lillecarl: I tested on my own apple though.

contrib/extractor/System.cpp
Lillecarl: Yes another warning.

PR #146, opened 2016-11-26
Implement TriggerCastFlags

killerwife:


ItsmeAga:
Not sure whatfor we need those flags, afaik spell attributes exist for those cases

killerwife:
@ItsmeAga for serverside settings of triggered spells. For "ignore hit calculation" there is an attribute BUT it is quite ambiguous whether the attribute is missing from clientside OR if there is some greater logic behind it. I want to use this system for finding that "greater logic". For "can target unselectable" there is no attribute and was previously only skippable by triggered spells, which is somewhat wrong because triggered does so many more things.

Laizerox:
@ItsmeAga also noteworthy note dbc data is incomplete and spells either miss or may also have wrong atteibutes.

PR #155, opened 2016-12-13
Tbc-db column implementation exports

killerwife:
Two features required:

1. Overriding a vendor template during an event (lets say valentine) where you need new items sold

2. Maximum reputation threshold for rewarding

Some quests in TBC stop giving reputation in the middle of honored for example.


cyberium:
merged

PR #157, opened 2016-12-17
Activate creatures and objects during cinematics

krullgor:
port work of @r00ty-tc in TrinityCore/TrinityCore@87b4533046d337fe05d7891c3632d34f8811b73e & TrinityCore/TrinityCore@de918a0f644794beebff2234b53980b577b27e39

before & after comparison
[http://imgur.com/a/qlguT](http://imgur.com/a/qlguT)

### DIFF
--- a/contrib/extractor/System.cpp
+++ b/contrib/extractor/System.cpp
@@ -58,11 +58,12 @@ uint32 maxAreaId = 0;
 enum Extract
 {
     EXTRACT_MAP = 1,
-    EXTRACT_DBC = 2
+    EXTRACT_DBC = 2,
+    EXTRACT_CAMERA = 4
 };
 
 // Select data for extract
-int   CONF_extract = EXTRACT_MAP | EXTRACT_DBC;
+int   CONF_extract = EXTRACT_MAP | EXTRACT_DBC | EXTRACT_CAMERA;
 // This option allow limit minimum height to some value (Allow save some memory)
 // see contrib/mmap/src/Tilebuilder.h, INVALID_MAP_LIQ_HEIGHT
 bool  CONF_allow_height_limit = true;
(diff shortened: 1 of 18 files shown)


killerwife:
I really need to test this one soon-ish. Really curious.

cala:
I'm interested by this PR but did not/do not have time to test this yet. One question for @cyberium will this work be redundant once your new map system is complete?

r00ty-tc:
@cala I'm not too familiar with mangos/cmangos. I certainly don't know what the new map system is. However, I don't think this can be avoided, because the client doesn't send any position information during the cinematic. It has to be derived serverside from the M2 files (hence this change). This actually does a crude job (not taking into account the curves for example).

So, in order for the server to know which creatures to activate and send information about to the player the server needs to derive the current location in the cinematic, which this provides.

One thing to note is that when I do get time I intend to change the TC version to use a virtual camera (it requires some core work) rather than the invisible creature currently used. Checking sniffs, it's clear it never used any kind of object and was all done serverside with some "virtual" position.

VladimirMangos:
Maybe more clear for checks and code create cinematic mgr for time cinmematic activation instead just set active cinematic and delete at  cinematic finish? This is seldom event so why have it in other time...

killerwife:
@r00ty-tc have you had time to do the virtual camera or is your V1 still the same in TC? I haven't been able to keep up.

r00ty-tc:
I haven't done much on TC or mangos at all recently. I am still intending to do this, but it's backed behind another project (Pooling, now that dynamic spawning is in). Which I'm weaving some design work on in the small amount (sadly) of spare time I have.

Although since this doesn't need anything near as much complexity, just a bit of investigation on my part. I might try to take a look some time soon.

As of right now, it's not changed beyond a few minor changes I made (unsure if that was pre-post my previous comment).

PR #163, opened 2016-12-23
Add visual to script

killerwife:
('45787','0','101','21','6','1','0','4','0','0','0','1','Talbuk Tag
Sleep'); is already in wotlk DBC

Rest is handled in ACID, fixes an ACID error as well


killerwife:
Found a way to do it using ACID only.

PR #166, opened 2016-12-26
Script Ghost-O-Plasm round up quest

killerwife:
Required SQL to test
```sql 
UPDATE gameobject_template SET type=1,data2=11796480 WHERE entry=177746;
UPDATE creature_template SET FactionAlliance=0,FactionHorde=0 WHERE entry=11560;
DELETE FROM creature_template_addon WHERE entry=11560;
INSERT INTO creature_template_addon(entry, mount, bytes1, b2_0_sheath, b2_1_flags, emote, moveflags, auras) VALUES(11560,0,0,1,16,0,0,'18146');```

### DIFF
--- a/src/scriptdev2/scripts/kalimdor/desolace.cpp
+++ b/src/scriptdev2/scripts/kalimdor/desolace.cpp
@@ -678,6 +678,74 @@ bool QuestAccept_npc_rigger_gizelton(Player* pPlayer, Creature* pCreature, const
     return true;
 }
 
+enum
+{
+    SPELL_CURSE_OF_MAGRAMI = 18159,
+    SPELL_GHOST_SPAWN_IN   = 17321,
+    SPELL_BLUE_AURA        = 17327,
+    SPELL_GREEN_AURA       = 18951,
+
+    FACTION_HOSTILE        = 16,
+};
+
+struct npc_magrami_spectre : public ScriptedAI
+{
+    npc_magrami_spectre(Creature* pCreature) : ScriptedAI(pCreature) { Reset(); }
+
+    uint32 m_uiCurseTimer;
+
+    void Reset() override
+    {
+        m_uiCurseTimer = 4000;
+    }
+
+    void JustAIInit() override
+    {
+        m_creature->CastSpell(m_creature, SPELL_GHOST_SPAWN_IN, TRIGGERED_NONE);
+        m_creature->CastSpell(m_creature, SPELL_BLUE_AURA,TRIGGERED_NONE);
+    }
+
+    void MovementInform(uint32 uiMovementType, uint32 uiData) override
+    {
+        if (uiMovementType == POINT_MOTION_TYPE)
+        {
+            if (uiData == 1)
+            {
+                float x, y, z;
+                m_creature->GetPosition(x, y, z);
+                m_creature->GetMotionMaster()->MoveRandomAroundPoint(x, y, z, 3.f);
+                m_creature->RemoveAurasDueToSpell(SPELL_BLUE_AURA);
+                m_creature->CastSpell(m_creature, SPELL_GREEN_AURA, TRIGGERED_NONE);
+                m_creature->SetFactionTemporary(FACTION_HOSTILE, TEMPFACTION_NONE);
+            }
+        }
+    }
+
+    void UpdateAI(const uint32 uiDiff) override
+    {
+        if (!m_creature->SelectHostileTarget() || !m_creature->getVictim())
+            return;
+
+        if (m_uiCurseTimer)
+        {
+            if (m_uiCurseTimer <= uiDiff)
+            {
+                m_uiCurseTimer = 0;
+                DoCastSpellIfCan(m_creature->getVictim(), SPELL_CURSE_OF_MAGRAMI);
+            }
+            else
+                m_uiCurseTimer -= uiDiff;
+        }
+
+        DoMeleeAttackIfReady();
+    }
+};
+
+CreatureAI* GetAI_npc_magrami_spectre(Creature* pCreature)
+{
+    return new npc_magrami_spectre(pCreature);
+}
+
 void AddSC_desolace()
 {
     Script* pNewScript;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by Rushor** [comment]
nice even with emotes <3

**Comment by killerwife** [comment]
I love how clang compiled yet i made a mistake.

File: src/scriptdev2/scripts/kalimdor/desolace.cpp [hunk 1]
```diff
@@ -678,6 +678,74 @@ bool QuestAccept_npc_rigger_gizelton(Player* pPlayer, Creature* pCreature, const
...
+    npc_magrami_spectre(Creature* pCreature) : ScriptedAI(pCreature) { Reset(); }
+
+    uint32 m_uiCurseTimer;
+
+    void Reset() override
+    {
+        m_uiCurseTimer = 4000;
+    }
+
+    void JustAIInit() override
```
xfurry: Do we have such a function which we can override?
And if yes, whey not just use Reset()?

Also, please use DoCastSpellIfCan() below.

src/scriptdev2/scripts/world/world_map_scripts.cpp
```diff
@@ -156,7 +156,8 @@ struct world_map_kalimdor : public ScriptedMap
         {
             for (auto guid : eventData.summonedMagrami)
                 if (Creature* magrami = instance->GetCreature(guid))
-                    magrami->ForcedDespawn();
+                    if(magrami->isAlive()) // dont despawn corpses with loot
```
xfurry: Codestyle: ```if (...)```

src/scriptdev2/scripts/kalimdor/desolace.cpp
killerwife: I just fixed that, I have that locally to fix the EscortAI issue that is still not fixed in cmangos. Also there is a good reason I did not use DoCastSpellIfCan, I used raw CastSpell so that it cannot fail, WITHOUT using triggered flag. If you notice, I used DoCastSpellIfCan with the combat spell.

xfurry: Well, my understanding is that if you don't specify any triggered flag in DoCastSpellIfCan(), it will use automatically TRIGGERED_NONE.
If this isn't the case now, then we need to fix it.

src/scriptdev2/scripts/world/world_map_scripts.cpp
killerwife: Added a commit to fix it, will squash at the end.

src/scriptdev2/scripts/kalimdor/desolace.cpp
killerwife: Yes it is, but DoCastSpellIfCan contains redundant checks mostly done in Spell::CheckCast, hence if its not a combat spell, but a script spell, there is really no point in using it, ever since SD2 was merged with core.

boxa:
Good work!
@killerwife, tell me please, which means data2=11796480 here?
UPDATE gameobject_template SET type=1, _**data2=11796480**_ WHERE entry=177746;

from GameObject.h
// 1 GAMEOBJECT_TYPE_BUTTON
uint32 autoCloseTime; // data2 - secs till autoclose = autoCloseTime / IN_MILLISECONDS

11796480 -- 3+ hours. what is this for? typo?

killerwife:
its in different units, its 65556 * seconds, look in gameobject_template wiki page for button

boxa:
@killerwife thx, i see now -- tbc and wotlk have different data field values
tbc:  seconds * 65556
wotlk:  seconds * 1000
will know now =)

killerwife:
Type was incorrect (as per sniff), and data was incorrect based on animation that happens in TBC with it. I adjusted it based on that.

src/scriptdev2/scripts/kalimdor/desolace.cpp
xfurry: I don't disagree, but I think that DoCastSpellIfCan() looks cleaner and nicer. :smile:

xfurry:
> Type was incorrect (as per sniff), and data was incorrect based on animation that happens in TBC with it. I adjusted it based on that.

Ok. 
I was just asking, because you requested something similar for the Warlock quest in Dire Maul, and in the end it turned up that we were able to trigger the animation and make everything work fine, without changing the type of the gameobject.

killerwife:
Actually if I remember correctly, WE DID change type that time as well, when evil-at-wow gave us sniff that showed that it was wrong as well :D Or am I tripping?

xfurry:
> Actually if I remember correctly, WE DID change type that time as well, when evil-at-wow gave us sniff that showed that it was wrong as well :D Or am I tripping?

No, we didn't change that. :wink: 
I couldn't find it anywhere in the sniff. Besides it just worked using the right commands in SD2. Check my commit and you will see. https://github.com/cmangos/mangos-wotlk/commit/a0e24efbd2861217ac9bf852009624ec5936ce90 https://github.com/cmangos/wotlk-db/commit/6ac1654b1054bdecb089a8174ff9f7ba544e22a4

PS. [offtopic] what happened to @Phatcat Haven't seen him for some time.

killerwife:
@cyberium revoked his rights due to a situation that occured on IRC. Also offtopic, but please remember to fix the failing conditions for Dreadsteed quest.

Phatcat:
@xfurry
Yeah, I only joined this project cause it was educational in nature (yeah, right...) and when you realize that most of who is doing the 'education' is yourself (even thugh this is the first c++ or mmo project you've worked on) ,and that not a single person cares about what you know or your betterment, then it's time to look for other ventures. Also I may have been a jerk and an a-hole but I always treated people's work with respect, I maybe not treated their person with such, because they were acting in a way that deserved no respect (demonstrably lying, putting themselves on a pedistal, generally being elitist twats). I howerver never belittled anyone over their work, their dedication towards wow emulation.
It would be a lie to say that people treated my work with respect, especially in the end. And when it's internal, people on a team you're supposed to be a 'part of' that basically make fun of your work behind your back, who belittle you for your general lack of knowledge about said mechanic, who none of them have no interest what so ever in educating you properly about it.

Yeah, you guys have fun with this, but I am out,.

xfurry:
> Also offtopic, but please remember to fix the failing conditions for Dreadsteed quest.

Give me some details please. I think I might have missed this one.

killerwife:
There are those 3 GOs, if none are active, the event fails afaik. I think Mantislord explained it on the commit in which you added it.

cala:
>  just fixed that, I have that locally to fix the EscortAI issue that is still not fixed in cmangos.

@killerwife I did not play an escort quest recently and do not have a PC at hand so the answer may be pretty obvious but what is this issue with escortAI ? Can we easily fix it in CMaNGOS too ?

killerwife:
Well try City of Light quest in mangos-tbc, whether it is working properly. I do not have an example for vanilla I am afraid. There are several possible fixes, it is more of a design decision than any real difficulty. At least it was when I hacked it on vengeance.

cala:
Thanks @killerwife I'll look at this quest to see what it is about and then we will discuss what to do on cmangos.

PR #170, opened 2017-01-09
Tempest Keep Update

Qingle:
### - Movement has been improved, has the chance to do one of 3 things:
1. Stay at the current platform and summon an ember.
2. Move to the next platform and summon an ember.
3. Move to the middle of the room and Flame Quill.

### - Updated the threat system to do the following:
1. Check first if there are any melee in range.
2. If no melee are in range, target the person at range with the highest threat.
3. Use Flame Buffet if they are out of range for auto attack.

## Void Reaver:
- Fixed a bug where he **wouldn't** use Arcane Orb if everyone is within 18 yards.

## High Astromancer Solarian:
- Implemented the AoE arcane damage for Wrath of the Astromancer.
- Made it so that she does Mark of Solarian on every Blinding Light.
- Fixed Timers.


Qingle:
Closing this until I can sort out my repos.

PR #172, opened 2017-01-11
Tempest Keep Update

Qingle:
### - Movement has been improved, has the chance to do one of 3 things:
1. Stay at the current platform and summon an ember.
2. Move to the next platform and summon an ember.
3. Move to the middle of the room and Flame Quill.

### - Updated the threat system to do the following:
1. Check first if there are any melee in range.
2. If no melee are in range, target the person at range with the highest threat.
3. Use Flame Buffet if they are out of range for auto attack.

## Void Reaver:
- Fixed a bug where he **wouldn't** use Arcane Orb if everyone is within 18 yards.

## High Astromancer Solarian:
- Implemented the AoE arcane damage for Wrath of the Astromancer.
- Made it so that she does Mark of Solarian on every Blinding Light.
- Fixed Timers.

Note: Squashed to one.

### DIFF
--- a/src/scriptdev2/scripts/outland/tempest_keep/the_eye/boss_alar.cpp
+++ b/src/scriptdev2/scripts/outland/tempest_keep/the_eye/boss_alar.cpp
@@ -363,8 +363,7 @@ struct boss_alarAI : public ScriptedAI
                 }
             }
             else
-                m_PlatformMoveTimer -= uiDiff;
-            }
+                m_uiPlatformMoveTimer -= uiDiff;
         }
         // Combat phase
         else if (m_uiPhase == PHASE_TWO)


xfurry:
Poke @killerwife to review.
I can suggest some small improvements, but I don't have time to go into details at the moment.

killerwife:
Actually I was the one who gave him the incentive to look at it. This goes hand in hand with the complete SQL revamp of TK. So yeah, I will provide feedback once I am done with my school/work on the attunement.

Qingle:
I think I have an idea of how it can be improved, especially with the Solarian AoE.

The dummy has a range, so if I can update Spell.cpp to use the range to get a targetList, would that work?
As well as that I could use the SpellEffects.cpp#L323 to do the damage split?

robinsch:
No it's not EffectDMG.

Qingle:
Okay thank you for the insight, any specific improvements you can think of robinsch?
I'm new to this stuff still :)

src/scriptdev2/scripts/outland/tempest_keep/the_eye/boss_alar.cpp
```diff
@@ -156,17 +157,64 @@ struct boss_alarAI : public ScriptedAI
                 pSummoned->AI()->AttackStart(pTarget);
         }
     }
-
+    
     void SummonedCreatureJustDied(Creature* pSummoned) override
     {
-        // drain 3% of boss health when the ember dies
+        // drain 2% of boss health when the ember dies
```
Patrick-Edouard: You changed the comment, but not the value in the code bellow.

According to wow wiki it's indeed 2% and not 3.

killerwife:
@Qingle I dont see anything inherently wrong about any of your changes. Only disputable thing is that Wrath of the Astromancer Dot implementation, but not because of your code, but because we know so little about it.

Qingle:
@killerwife the problem i see about the DoT is that it doesn't actually show up on the combat log, which means it may be implemented in a different way. I'll do some digging and see if there is another way to go about it

PR #175, opened 2017-01-16
[SpellAura] Dose flame touched aura on flame sear

Patrick-Edouard:
The Eredar twins from Sunwell Plateau.

Flame sear dots should add stacks of Flame Touched debuff.

In the current state of mangos, it only applies an initial flame touched when it lands, but not on DoT ticks. (Can provide youtube video if requiered).

The flame sear dot lasts for 6 second ticking every 500ms. But on youtube video we clearly see it stack one flame touched per second, for a total of 6 stacks. I assume there is a hiden cooldown mechanic that should maybe be managed with the dark flame auras.

### DIFF
--- a/src/game/SpellAuras.cpp
+++ b/src/game/SpellAuras.cpp
@@ -5724,21 +5724,6 @@ void Aura::PeriodicTick()
                         }
                         break;
                     }
-                    case 46771: 
-                    {                               // Flame Sear (SWP, eredar twins)
-                        if (target->GetTypeId() == TYPEID_PLAYER)
-                        {
-                            if (!target->HasAura(45343) &&
-                                !target->HasAura(47300)) {          // No Dark Flame applied. There are two auras
-                                if (m_periodicTick % 4 == 1) {      // Only doze the aura every 2 ticks (should maybe be one second cooldown)
-                                    target->CastSpell(target, 45348,
-                                                      TRIGGERED_OLD_TRIGGERED, nullptr,
-                                                      this); // apply a doze of Flame Touched
-                                }
-                            }
-                        }
-                        break;
-                     }
                     default:
                         break;
                 }
--- a/src/game/SpellAuras.cpp
+++ b/src/game/SpellAuras.cpp
@@ -5724,6 +5724,19 @@ void Aura::PeriodicTick()
                         }
                         break;
                     }
+                    case 46771: {                               // Flame Sear
+                        if (target->GetTypeId() == TYPEID_PLAYER) {
+                            if (!target->HasAura(45343) &&
+                                !target->HasAura(47300)) {          // No Dark Flame applied. There are two auras
+                                if (m_periodicTick % 4 == 1) {      // Only doze the aura every 2 ticks (should maybe be one second cooldown)
+                                    target->CastSpell(target, 45348,
+                                                      TRIGGERED_OLD_TRIGGERED, nullptr,
+                                                      this); // apply a doze of Flame Touched
+                                }
+                            }
+                        }
+                        break;
+                     }
                     default:
                         break;
                 }
--- a/src/game/SpellAuras.cpp
+++ b/src/game/SpellAuras.cpp
@@ -666,7 +666,10 @@ void AreaAura::Update(uint32 diff)
                     holder->SetInUse(false);
                 }
                 else
+                {
                     (*tIter)->AddSpellAuraHolder(holder);
+                    holder->SetState(SPELLAURAHOLDER_STATE_READY);
+                }
 
             }
         }
--- a/src/game/SpellAuras.cpp
+++ b/src/game/SpellAuras.cpp
@@ -5724,8 +5724,10 @@ void Aura::PeriodicTick()
                         }
                         break;
                     }
-                    case 46771: {                               // Flame Sear
-                        if (target->GetTypeId() == TYPEID_PLAYER) {
+                    case 46771: 
+                    {                               // Flame Sear (SWP, eredar twins)
+                        if (target->GetTypeId() == TYPEID_PLAYER)
+                        {
                             if (!target->HasAura(45343) &&
                                 !target->HasAura(47300)) {          // No Dark Flame applied. There are two auras
                                 if (m_periodicTick % 4 == 1) {      // Only doze the aura every 2 ticks (should maybe be one second cooldown)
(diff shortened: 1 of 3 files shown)


src/game/SpellAuras.cpp
```diff
@@ -5724,6 +5724,19 @@ void Aura::PeriodicTick()
                         }
                         break;
                     }
+                    case 46771: {                               // Flame Sear
+                        if (target->GetTypeId() == TYPEID_PLAYER) {
```
killerwife: Please use C++ brackets styling, not Java :P

src/game/SpellAuras.cpp
```diff
@@ -5724,6 +5724,19 @@ void Aura::PeriodicTick()
...
+                            if (!target->HasAura(45343) &&
+                                !target->HasAura(47300)) {          // No Dark Flame applied. There are two auras
+                                if (m_periodicTick % 4 == 1) {      // Only doze the aura every 2 ticks (should maybe be one second cooldown)
+                                    target->CastSpell(target, 45348,
```
killerwife: I believe you should apply different debuffs on 45343 presence, and different on 47300. Right now you apply only fire one.

killerwife:
Also on a general note, this code should be handled in UnitAuraProcHandler.cpp not PeriodicTick() function.

src/game/SpellAuras.cpp
Patrick-Edouard: The boss mecanic is the following:

- One fire lady does fire damage and applies fire debuff when it ticks
- One shadow lady does shadow damage and applies shadow debuff when it ticks
- When someone has a fire or a shadow debuff and take damage from the opposite school, he gains 3 seconds of immunity for all debuff applications

I assume one aura is triggered when switching from fire to shadow and the other the other why around.

This test seems legit to me

killerwife: The check currently does "if no dark flame aura and if no fire flame aura do stuff", afaik u want "if dark flame aura procs do stuff, if currently has flame touched, run immunity" "if fire flame aura procs, do stuff, if currently has shadow touched, run immunity" or am I wrong?

Patrick-Edouard: What I want is :

If flame sear (id: 46771) dot is ticking
And if there is no dark flame aura (id: 45343 and 47300) (3 second imunity to fire auras)
Then do stuff (increase fire aura stacks)

Qingle:
Another additional comment, shouldn't 'target' be 'unitTarget'?

Patrick-Edouard:
@Qingle I have no clue about that; but in this function everyone seem to use 'target' that is defined at its beginning :

`Unit* target = GetTarget();`

Qingle:
@Patrick-Edouard Yes apologies, I'm too used to doing Dummy Auras ^^

Patrick-Edouard:
It seems the spell 45342 also have a wrong behavior : applies a stack of flame touched on landing, then nothing on ticks. 

CF retail video : https://youtu.be/AKdBJFzvJcY?t=1m30s

Qingle:
Just looked at the spell, it applies a periodic damage effect, a confuse and also target everyone around the caster and applies damage to them.

Have you got QSpellWork and a Spell.dbc file to work with?

Patrick-Edouard:
@Qingle Yo,

No I haven't started to look in this aspect of wowemu yet...

To be clear, what was triggering me is that flame touched debuff (id: 45348) isn't stacking properly on fire spells ticks (for flame sear id:46771 and conflagration id:45342 ), instead it just stack on spell landing.

So I think the root of the evil is that the current flame touched / shadow touched / dark flame is handled in the boss script here:

https://github.com/cmangos/mangos-tbc/blob/master/src/scriptdev2/scripts/eastern_kingdoms/sunwell_plateau/boss_eredar_twins.cpp#L192

While it should be spell effect? So reworking it would mean reworking all the encounter spells (flame sear, conflagration, blaze, shadow nova, shadow blade, flame touched, shadow touched, dark flame 1, dark flame  2). And delete this SpellHitTarget method.

It seems a bit too big for me ^^

Qingle:
Then I feel as @killerwife said, this should be done in UnitAuraProcHandler.cpp

Patrick-Edouard:
@Qingle @killerwife 

I've tried to play with `spell_proc_event` and add some of twins spell into it in order to make it trigger `Unit::HandleProcTriggerSpellAuraProc` or other functions `UnitAuraProcHandler.cpp` but without any result..

Some of the spells that I can trace from `HandleProcTriggerSpellAuraProc` are not even in this table. So I don't even know if I'm looking in the right direction.

It seems that trinity is using a table named `spell_link_spell`to handle this mecanic, but I can't find any equivalent in mangos.

I'm most likely going spend a few hours this week end, if I don't get forward I'll give up and just make an issue ^^

killerwife:
Ok, ill try to look this evening.

killerwife:
Apply this patch https://gist.github.com/killerwife/e0665b3b12435abef543562d8bf16bb0 after that it enters 

SpellAuraProcResult Unit::HandleProcTriggerSpellAuraProc(Unit* pVictim, uint32 damage, Aura* triggeredByAura, SpellEntry const* procSpell, uint32 procFlags, uint32 procEx, uint32 cooldown)

and you can script away. The proc should do all, immunities, flame touched/shadow touched. Also since there are two Dark Flame auras, I think one should be named "Normal Flame Aura" and should work for fire spells.

Patrick-Edouard:
Ok, I made some changes to handle those procs.

It's still not perfect and some part are a bit hacky. One of the main concern is about the number of stacks. On Flame Sear there should be 6 stack of flame touched and I got 7 with this:
CF video : https://www.twitch.tv/vooml/v/116256396

Patrick-Edouard:
Second concern is about this: https://github.com/cmangos/mangos-tbc/pull/175/files#diff-a35c8aeb0556a8ce0acb08c987d3d23bR1744

The remove aura should be handled in the dark flame proc?

cyberium:
@killerwife sound like abandonned/ not correct? Close this if any.

Patrick-Edouard:
@cyberium well it's the blizzlike behavior. The server I was playing on merged this and since I had no feedback here, I had no incentive to keep posting.

killerwife:
@cyberium no, all that needs to be done is that it needs to be rebased properly, so that the changes are visible. This PR actually seems to be doing it properly, its just hard to see the changes.

cyberium:
@killerwife cleaned here -> https://gist.github.com/cyberium/89815c9630c21e9aaab90f06d1141a70
You have no excuse now :P

killerwife:
@cyberium if you fix up formatting a bit, then its indeed good to go. I remember helping @Patrick-Edouard a bit on IRC but I think due to school I forgot about it. My bad.

Patrick-Edouard:
@killerwife just a head up. Seems cyberium made the so called clean up. Where you expecting me to report it to the pull request or something?

killerwife:
I saw the cleaned up code. Someone, either you, or someone else has to punch it into git and commit it to the repo. I am spreading myself so thin lately that I don't keep up with everything.

PR #177, opened 2017-01-23
Implement only moving closer setting in MoveChase

killerwife:
Enables caster movement using MoveChase, usable for Gruul. Be careful when using with LOS.
@Drikish this is for you


OriginsDev:
Thanks Killerwife!!!!

PR #179, opened 2017-02-16
#1084 & #1283 - fix server hang on shutdown due to fgets is blocking

ulfgebhardt:
# fix server hang on shutdown due to fgets is blocking

This fixes the Issues https://github.com/cmangos/issues/issues/1084 and https://github.com/cmangos/issues/issues/1283 on Linux.

Im unsure and unable to test if this bug is present in Windows.

The reason for the Server hanging is the blocking character of fgets() on stdin. Therefore changing the stdin to non-blocking is a solution to the Problem.

Please look at this @Cyberium

### DIFF
--- a/src/mangosd/CliRunnable.cpp
+++ b/src/mangosd/CliRunnable.cpp
@@ -606,7 +606,15 @@ void CliRunnable::run()
     // print this here the first time
     // later it will be printed after command queue updates
     printf("mangos>");
-
+    
+#ifdef linux
+    //Set stdin IO to nonblocking - prevent Server from hanging in shutdown process till enter is pressed
+    int fd = fileno(stdin);  
+    int flags = fcntl(fd, F_GETFL, 0); 
+    flags |= O_NONBLOCK; 
+    fcntl(fd, F_SETFL, flags);
+#endif
+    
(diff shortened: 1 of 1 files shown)


ulfgebhardt:
Can you specify what is not working? This PR fixes the blocking character of stdin, which results in a segfault on shutdown - you need to press enter to shut down the programm. With this PR there is no segfault and no need to press enter.

All this applies to linux only, since the linux specific fcntl() fucntion is used - therfore the added code is in a 

```
#ifdef linux

#endif
```

preprocessor block. If you use Windows and have the described problem - this PR does not fi it!

DanielBallaSZTE:
I am using linux, and still have to hit enter (or any other character) for the server to halt.

KwazarFr:
Confirmed, using Debian and still blocked on "Halting process..."   have to press enter for exit

hlarsen:
@ulfgebhardt __linux__ works for me

https://sourceforge.net/p/predef/wiki/OperatingSystems/

ulfgebhardt:
I use Debian and the supplied fix works for me (at least in console).
The "#ifdef linux"  is used few lines below the suggested change in the same file.
I use gcc.

PR #180, opened 2017-02-18
#1177 double login account fix

ulfgebhardt:
# #1177 double login account fix

This fixes the issues https://github.com/cmangos/issues/issues/1177 and https://github.com/cmangos/issues/issues/1252

Please note that i moved

`const time_t currTime = time(nullptr);`

into 

```
bool ShouldLogOut() const
{
      return (_logoutTime > 0 && time(nullptr) >= _logoutTime + 20);
}
```

and therefore changed the api of ShouldLogOut. I did that to prevent a call to time(nullptr) every Update Cycle. This way its only called when _logoutTime is > 0.

Also i removed the ProcessLogout() from the PacketFilter. This causes Map::Update() to handle Playerlogouts, which seems to be a bad idea regarding this comment.

`// in Map::Update() we do not process player logout!`

This PR fixes the Problems with the double-login and connection loss. Anyhow im not able to tell if there is a Problem with Map::Update - might be problematic due to threatsafety and/or performance.

I need an expert to look at this @killerwife @cyberium

### DIFF
--- a/src/game/WorldSession.cpp
+++ b/src/game/WorldSession.cpp
@@ -288,20 +288,13 @@ bool WorldSession::Update(PacketFilter& updater)
         }
     }
 
-    // check if we are safe to proceed with logout
-    // logout procedure should happen only in World::UpdateSessions() method!!!
-    if (updater.ProcessLogout())
-    {
-        ///- If necessary, log the player out
-        const time_t currTime = time(nullptr);
+    // process logout if needed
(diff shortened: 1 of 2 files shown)


kotishe:
Whats news?

KwazarFr:
This is a really important PR.  Hope it's been modified and pushed asap.

KwazarFr:
We need an expert to look at this @killerwife @cyberium

cyberium:
This doesn't really fix the issue. We'll take care of this in near future

PR #182, opened 2017-03-29
Unify fleeing for assistance and normal fleeing

Ozerev:
This PR attempts to unify the behaviors of fleeing for assistance and normal fleeing. This is done by using the same movement generator and the same flags for both behaviors. It also fixes some bugs related to fleeing.

This should fix the following errors:

- Units will no longer keep autoattacking when fleeing for assistance. Previously they would attack you even though they were not targeting you.
- Units fleeing for assistance will now revert to normal fleeing if they can't find a path to their target.
- UNIT_FLAG_FLEEING should now properly be cleared after the unit stops fleeing. Previously it would just stay set, even after exiting combat.
- Fleeing enemies should no longer completely ignore pathfinding in some cases. At least one case of this happened because the generated paths were too short, so I have added a check for PATHFIND_SHORT in FleeingMovementGenerator. This does lead to enemies sometimes standing still for a while, since all generated paths are invalid, but it's probably better than them running through walls.
- Enemies fleeing for assistance will no longer ignore effects that prevent fleeing, like Curse of Recklessness.

Opinions?

### DIFF
--- a/src/game/Creature.cpp
+++ b/src/game/Creature.cpp
@@ -132,7 +132,7 @@ Creature::Creature(CreatureSubtype subtype) : Unit(),
     m_lootStatus(CREATURE_LOOT_STATUS_NONE),
     m_corpseDecayTimer(0), m_respawnTime(0), m_respawnDelay(25), m_corpseDelay(60), m_aggroDelay(0), m_respawnradius(5.0f),
     m_subtype(subtype), m_defaultMovementType(IDLE_MOTION_TYPE), m_equipmentId(0),
-    m_AlreadyCallAssistance(false), m_AlreadySearchedAssistance(false),
+    m_AlreadyCallAssistance(false), m_AlreadySearchedAssistance(false), m_AllowCallAssistance(true),
     m_isDeadByDefault(false), m_temporaryFactionFlags(TEMPFACTION_NONE),
     m_meleeDamageSchoolMask(SPELL_SCHOOL_MASK_NORMAL), m_originalEntry(0),
     m_creatureInfo(nullptr), m_ai(nullptr), m_gameEventVendorId(0)
@@ -728,6 +728,9 @@ void Creature::RegenerateHealth()
 void Creature::DoFleeToGetAssistance()
 {
     if (!getVictim())
+        return;
+
+    if (HasAuraType(SPELL_AURA_PREVENTS_FLEEING))
         return;
 
     float radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_FLEE_ASSISTANCE_RADIUS);
@@ -1831,7 +1834,7 @@ void Creature::SendAIReaction(AiReaction reactionType)
 void Creature::CallAssistance()
 {
     // FIXME: should player pets call for assistance?
-    if (!m_AlreadyCallAssistance && getVictim() && !isCharmed())
+    if (!m_AlreadyCallAssistance && AllowCallAssistance() && getVictim() && !isCharmed())
     {
         SetNoCallAssistance(true);
 
--- a/src/game/FleeingMovementGenerator.cpp
+++ b/src/game/FleeingMovementGenerator.cpp
@@ -47,7 +47,7 @@ void FleeingMovementGenerator<T>::_setTargetLocation(T& owner)
     PathFinder path(&owner);
     path.setPathLengthLimit(30.0f);
     path.calculate(x, y, z);
-    if (path.getPathType() & PATHFIND_NOPATH)
+    if (path.getPathType() & (PATHFIND_NOPATH | PATHFIND_SHORT))
     {
         // path not found recheck later
         i_nextCheckTime.Reset(50);
--- a/src/game/Unit.cpp
+++ b/src/game/Unit.cpp
@@ -9596,7 +9596,7 @@ void Unit::SetIncapacitatedState(bool apply, uint32 state, ObjectGuid casterGuid
 
     Player* controller = GetCharmerOrOwnerPlayerOrPlayerItself();
     const bool control = controller ? controller->IsClientControl(this) : false;
-    const bool movement = (state != UNIT_FLAG_STUNNED);
+    const bool movement = (state != UNIT_FLAG_STUNNED) || !apply;
     const bool stun = !!(state & UNIT_FLAG_STUNNED);
     const bool fleeing = !!(state & UNIT_FLAG_FLEEING);
 
--- a/src/game/Creature.h
+++ b/src/game/Creature.h
@@ -671,7 +671,9 @@ class MANGOS_DLL_SPEC Creature : public Unit
         void CallAssistance();
         void SetNoCallAssistance(bool val) { m_AlreadyCallAssistance = val; }
         void SetNoSearchAssistance(bool val) { m_AlreadySearchedAssistance = val; }
+        void SetAllowCallAssistance(bool val) { m_AllowCallAssistance = val; }
         bool HasSearchedAssistance() const { return m_AlreadySearchedAssistance; }
+        bool AllowCallAssistance() const { return m_AllowCallAssistance; }
         bool CanAssistTo(const Unit* u, const Unit* enemy, bool checkfaction = true) const;
         bool CanInitiateAttack() const;
 
@@ -778,6 +780,7 @@ class MANGOS_DLL_SPEC Creature : public Unit
         // below fields has potential for optimization
         bool m_AlreadyCallAssistance;
         bool m_AlreadySearchedAssistance;
+        bool m_AllowCallAssistance;
         bool m_isDeadByDefault;
         uint32 m_temporaryFactionFlags;                     // used for real faction changes (not auras etc)
 


Warlockbugs:
Movement generators are a constant source of numerous issues with AI and npc behavior. Thanks for taking the time to improve the situation. There is, however, a number of complications, why changes to retreating routine were avoided in cMaNGOS for quite some time.

*  SetIncapacitatedState's current rendition originally meant strictly for spells, to remove duplicating code and concentrate it on a single screen. Existing usage for creature retreat is a temporary placeholder coming from pre-existing code, until entire "Retreat/Seek for help" routine is thoroughly researched. Generally speaking, if one has to call something with spellId 0, that means bad design (a hack) and underlying systems need a rework. I suspect, that In such cases, an overhaul with a dedicated API is preferred, rather than further extensions on something that is clearly not fit for intended usage.

* Random direction fleeing for help is likely a more "true to the source" scenario for typical seek for assistance instead of stock pinpoint-accurate move into another creature. Which leads us to the next bullet point.

* It is possible, that entire creature retreating reaction is actually just a hidden server-side self-fear spell with periodic AoE trigger "Call for help" kind of effect on friendly units. That's theory i stick with within the "Everything is a spell" paradigm we should follow. 

This changeset is rather trying to conceal the problem with existing APIs and needs to be covered with "TODOs" at least to avoid shutting the door for original research on the subject.

@cyberium 
@killerwife

killerwife:
I have one point I need to mention as well. Currently our system for "call for help" is slightly flawed. We have no way to turn it off. Would be nice if you could incorporate this into the PR, i.e. a mob is in combat, its near his allies but does not call for help if a certain setting is set (this is merely for scripting purposes).

Also while I agree with WBs "everything is spell" approach, it might be that we will never find this particular spell even if it exists. I suggest you figure out a better way to handle the SetIncapacitatedState thing, in a way that will be more intuitive and other than that the PR is good with the current code.

Ozerev:
Ok, I have made some changes to SetIncapacitatedState to reduce the number of ugly optional arguments and moved some parts out into the calling functions. 

I'm a bit curious about the purpose of these lines:
```
if (apply)
        CastStop(GetObjectGuid() == casterGuid ? spellID : 0);
```
I would like to add a comment about it since it's not obvious why this check exists. Are there spells that stun/fear/confuse the caster that should not be interrupted or something?

I also added another bug fix, if you stun an enemy right before a fear expires, its movement will be really strange, it will move as if it wasn't in combat but still attack you if you come close.

Also, do you know of anything specific that should be tested with these changes? I have tested overlapping a bit, like what happens if you stun then fear, or root a fleeing enemy etc. If you know of anything that might break let me know and I'll test it.

@killerwife About the first part, do you mean something like SetNoCallAssistance that is only used by scripts, or disabling the whole "seek assistance" behavior?

killerwife:
The former, a setting which would enable disabling assistance calling in scripts :) Also, apparently there is a way to break sheep. They can turn mobs into passive non-reactionary husks. Maybe you might wanna try to look at that.

Warlockbugs:
I fixed spell overlapping and many other side effects some time ago: https://github.com/cmangos/mangos-tbc/pull/114
You are introducing more code duplication and likely additional bugs to the routine at this point.
As an example for spellID check, it is for various self-stuns, such as Curator's Evocation.

The real problem was and still is interaction between AI and movement generator.

Ozerev:
@Warlockbugs Did you try stunning a creature when fear or confusion/sheep expires? (Hint: it makes weird things happen). 

This is not meant to be a blizzlike implementation of fleeing, it is meant to be a bugfix for the following bugs:

- Units kept attacking enemies while fleeing for assistance, because the CanReactInCombat() check does not return true when fleeing for assistance.
- AssistanceMovementGenerator had no way to handle failed paths.
- AssistanceMovementGenerator did not care how long it took to reach it's target, there was no time limit or path length limit, which could lead to unintended behavior (although probably very rarely).
- Fleeing enemies sometimes ignored terrain/walls and just ran through stuff.
- Units fleeing for assistance ignored things like Curse of Recklessness.
- Units that were stunned when fear or confusion expired reverted to non-combat movement.

I simply thought that the easiest way to fix some of these issues was to merge "fleeing for assistance" with "fleeing". But if you are fine with keeping the current behavior until someone, some time in the future (it's been what, 10 years, since TBC was released? I'm sure it will happen any day now), does a perfect blizzlike implementation then sure, I can abandon this and go do something else.

@killerwife I added m_AllowCallAssistance to Creature. I know of one issue with sheep, if the unit is stunned when sheep expires it will not chase its target, it will just walk slowly balk to it's original position and look at its target. This is already fixed in this PR.

Warlockbugs:
I am more concerned about spell handlers, which you just modified, much less than fleeing for assistance mechanics. The problem with changing spell handlers is the need for an exhaustive test for all corner cases and obscure clashes (such as mentioned Curator's Evocation, which requires additional CastStop check). If you are in the mood to test all possible cases again instead of "hit an run and let others handle the fallout" mindset i have no issues with the change.

Ozerev:
I see. I have reverted all but some minor bugfixes, since the other changes would require either
- Extending the usage of SetIncapacitatedState, which was not desirable
- Modifying SetIncapacitatedState and retesting everything _manually_, which wont happen
- Creating a duplicate SetIncapacitatedState used for fleeing only, which is way too ugly

Anyway the current changes fixes fleeing through walls, passive behavior if units are stunned when fear or confusion expires, and seeking assistance no longer ignores flee prevention effects. It is still a change in SetIncapacitatedState, so decide among yourselves if you want it or not.

Warlockbugs:
One of the major reasons SetIncapacitatedState was designed in the way it currently exists, is completely broken movement generator mutations. This method performs forced "manual" movement generator management for CC spells instead of "natural" movement generator mutation to plug the underlying holes in movement generators. I am sorry if you took my criticism personally or something, but dissecting these handlers really is opening Pandora's box. I hope, this explains why this area is a major timesink in cMaNGOS and was avoided for quite some time. To fix this situation once and for all, it requires walking over all movement generators and AI interactions, which has to be done at some point.
Thanks for taking time to fix these issues. As long as it doesn't cause additional issues, it should be fine by me.

Ozerev:
OK, good to know that the movement system should be avoided for now

Warlockbugs:
Bump. Any other reviews, comments?

Warlockbugs:
Bump again. Anyone else?
On a side note, recent discovery confirmed that fleeing combat is indeed a built-in serverside script outside of spell system. So, the dreaded rework of movement generators will need to take this into account.

Tobschinski:
poke @cala - you should know everything about fleeing from combat

src/game/Creature.cpp
```diff
@@ -132,7 +132,7 @@ Creature::Creature(CreatureSubtype subtype) : Unit(),
     m_lootStatus(CREATURE_LOOT_STATUS_NONE),
     m_corpseDecayTimer(0), m_respawnTime(0), m_respawnDelay(25), m_corpseDelay(60), m_aggroDelay(0), m_respawnradius(5.0f),
     m_subtype(subtype), m_defaultMovementType(IDLE_MOTION_TYPE), m_equipmentId(0),
-    m_AlreadyCallAssistance(false), m_AlreadySearchedAssistance(false),
+    m_AlreadyCallAssistance(false), m_AlreadySearchedAssistance(false), m_AllowCallAssistance(true),
```
cyberium: 3 bool for callings assistances, i dont like that, its time to revamp it correctly. Add a structure, an enum or whatever to  unify call assistance parameters.

src/game/Creature.cpp
```diff
@@ -730,6 +730,9 @@ void Creature::DoFleeToGetAssistance()
     if (!getVictim())
         return;
 
+    if (HasAuraType(SPELL_AURA_PREVENTS_FLEEING))
```
cyberium: +1

src/game/Creature.cpp
```diff
@@ -1831,7 +1834,7 @@ void Creature::SendAIReaction(AiReaction reactionType)
 void Creature::CallAssistance()
 {
     // FIXME: should player pets call for assistance?
-    if (!m_AlreadyCallAssistance && getVictim() && !isCharmed())
+    if (!m_AlreadyCallAssistance && AllowCallAssistance() && getVictim() && !isCharmed())
```
cyberium: AllowCallAssistance seem not trivial name for me. I would like something more natural like CanCallAssisstance or AllowedToCallAssisstance.

src/game/FleeingMovementGenerator.cpp
```diff
@@ -47,7 +47,7 @@ void FleeingMovementGenerator<T>::_setTargetLocation(T& owner)
     PathFinder path(&owner);
     path.setPathLengthLimit(30.0f);
     path.calculate(x, y, z);
-    if (path.getPathType() & PATHFIND_NOPATH)
+    if (path.getPathType() & (PATHFIND_NOPATH | PATHFIND_SHORT))
```
cyberium: Ok, This is a bit too much, PATHFIND_SHORT is returned pretty often and in a large majority of case it dont cause any trouble. I think this change will have more bad side than good side. Remember even in retail there was some weird fleeing case ;)

src/game/Unit.cpp
```diff
@@ -9596,7 +9596,7 @@ void Unit::SetIncapacitatedState(bool apply, uint32 state, ObjectGuid casterGuid
 
     Player* controller = GetCharmerOrOwnerPlayerOrPlayerItself();
     const bool control = controller ? controller->IsClientControl(this) : false;
-    const bool movement = (state != UNIT_FLAG_STUNNED);
+    const bool movement = (state != UNIT_FLAG_STUNNED) || !apply;
```
cyberium: At first view it seem correct but i will delegate that change to @Warlockbugs because i know first view is 90% of the time not enough about that portion of the code.

src/game/Creature.cpp
killerwife: Current cmangos nomenclature can use either CanCallAssistance or IsAllowedToCallAssistance so that we stay consistent with getters.

src/game/Unit.cpp
Warlockbugs: This line is correct.

cyberium:
@Warlockbugs i'll close this but feel free to reopen it or create a new one if you judging its important.

PR #184, opened 2017-04-02
Don't take fall damage if SPELL_AURA_MOD_FLIGHT_SPEED_MOUNTED is present

Ozerev:
Flying mounts use SPELL_AURA_MOD_FLIGHT_SPEED_MOUNTED, which calls
SetCanFly. This means that activating instant flying mounts, like Flying
Broom and Swift Flying Broom, causes you to take fall damage in the air.

To reproduce the issue this fixes:

1. .additem 33176
2. Use Flying Broom
3. Fly up for a while
4. Remove the Flying Broom buff
5. Use Flying Broom again after a few seconds, but before hitting the ground
6. Die

### DIFF
--- a/src/game/Player.cpp
+++ b/src/game/Player.cpp
@@ -20211,7 +20211,7 @@ void Player::HandleFall(MovementInfo const& movementInfo)
     // 14.57 can be calculated by resolving damageperc formula below to 0
     if (z_diff >= 14.57f && !isDead() && !isGameMaster() && !HasMovementFlag(MOVEFLAG_ONTRANSPORT) &&
             !HasAuraType(SPELL_AURA_HOVER) && !HasAuraType(SPELL_AURA_FEATHER_FALL) &&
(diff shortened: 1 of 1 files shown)


killerwife:
How does Swift Flight Form work? Did that one suffer from this issue too?

Ozerev:
Swift Flight Form uses SPELL_AURA_FLY, which was already included. As far as I can see, the only mounts that were affected are Flying Broom, Swift Flying Broom, Flying Reindeer and Headless Horseman's Reins (does this even exist? .additem says invalid id but atlas shows it). So not exactly a big issue.

killerwife:
Headless Horsemans Reins - possibly missing cos WOTLK launched after Hallows end, and 3.0.1 patch was october 18 i think (right in middle of hallows end), so needs investigation :)

Lillecarl:
Can confirm issue with broom.
bool IsFreeFlying() const { return HasAuraType(SPELL_AURA_MOD_FLIGHT_SPEED_MOUNTED) || HasAuraType(SPELL_AURA_FLY); }

This function checks both flying cases.

I approve.

Warlockbugs:
Merged, thanks.

PR #188, opened 2017-05-25
Implement GossipSelect with items as senders

Lillecarl:
This allows you to create items that handle GossipMenus just like with creatures.


killerwife:
Hm, where in the game is this required?

Lillecarl:
Nowhere at all, it'd be to allow people to implement their own things.

killerwife:
Ugh, I believe there is a strict "no custom things" policy when it comes to scripting. But ask @xfurry and @cyberium . I would prefer to not be the one making this decision.

Lillecarl:
We'll see what they say :-)

cyberium:
Thank you, but we have enough to deal instead of this new custom feature.

PR #189, opened 2017-05-25
Fix lots of debug warnings

Lillecarl:
Don't merge this yet. I'm only recompiling code over and over, must test application before a merge is appropriate.


cyberium:
Lot of this are already applied by @Warlockbugs so it need some update.

PR #192, opened 2017-06-01
Implement script for q.10408

killerwife:
Complementary DB patch
https://github.com/cmangos/tbc-db/commit/2c5ef940b4b82884022d4fb594ecfb62d01eddfe

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/netherstorm.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/netherstorm.cpp
@@ -1580,6 +1580,231 @@ CreatureAI* GetAI_npc_dimensius(Creature* pCreature)
     return new npc_dimensiusAI(pCreature);
 }
 
+enum
+{
+    NPC_ENERGY_BALL                 = 20769,
+    NPC_SALHADAAR                   = 20454,
+
+    SPELL_FLOAT                     = 37613, // actual name C_C_O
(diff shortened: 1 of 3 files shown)


src/game/AI/ScriptDevAI/scripts/outland/netherstorm.cpp
```diff
@@ -1580,6 +1580,232 @@ CreatureAI* GetAI_npc_dimensius(Creature* pCreature)
...
+        }
+        if (m_uiAttackTimer)
+        {
+            if (m_uiAttackTimer <= uiDiff) // on attack start cast spell that handles everything
+            {
+                m_uiAttackTimer = 0;
+                std::list<Player*> playerList;
+                GetPlayerListWithEntryInWorld(playerList, m_creature, 100.0f);
+                std::list<Player*>::const_iterator itr = playerList.begin();
+                if (itr != playerList.end())
```
evil-at-wow: At this point itr is (still) playerList.begin(), so an alternative (and arguably more clear) version could be to use !playerList.empty() in the condition and write AttackStart( playerList.front() ) in the body.

killerwife: nice, seems like making a PR instead of just merging was a good idea :D

evil-at-wow: Let's say it's not a big deal either way, but it was a good excuse to show I'm still alive :wink:

PR #193, opened 2017-06-29
Some speedups while build

GiR-Zippo:
Using gold-linker for gcc if found (speedup linking)      
https://en.wikipedia.org/wiki/Gold_(linker)
This could be implemented for Clang too.

Setting g3dlite, gsoap and framework as OBJECT instead of STATIC
https://cmake.org/Wiki/CMake/Tutorials/Object_Library

Ignore Kdev-Projectfiles in .gitignore


evil-at-wow:
I think there's a subtle but very important difference between using OBJECT and STATIC. If I understood what CMake does well, using OBJECT you will create your final executable (realmd or mangosd) by adding all object files from those libraries. Unless the gold linker works differently from ld, this means object files that aren't really needed will be added as well, making your executables potentially bigger and slower. Using STATIC you will link your final executable against the static libraries (basically a container of object files) that have been created. And that means the linker will only add object files from the static library if they contain symbols that are referenced from the object files used to link your final executable. So yes, I can see why that part speeds up linking, but it has a penalty at runtime. And I'd prefer the link time penalty instead of the runtime penalty...

Let's add a minimal demo to demonstrate what I'm trying to tell. 
Assume two source files, a.cpp and b.cpp, each defining a function (the source files for your libraries):

```c++
int f1()
{
  return 1;
}
```

```c++
int f2()
{
  return 2;
}
```

And a main source file (for the executable) that only uses one of these functions:

```c++
extern int f1();

int main()
{
  return f1();
}
```

Now behold what I get on my machine:
```
# Compile library sources
$ g++ -o a.o -c a.cpp
$ g++ -o b.o -c b.cpp

# Create a static library with both object files as well.
$ ar crs libtest.a a.o b.o

# OBJECT case - linking using all object files.
$ g++ -o test-object main.cpp a.o b.o
# STATIC case - linking using the static library we've just created.
$ g++ -o test-static main.cpp -ltest -L.

# See, the object executable is bigger...
$ ls -l test-*
-rwxrwxr-x 1 user user 11904 29 jun 22:37 test-object
-rwxrwxr-x 1 user user 11848 29 jun 22:37 test-static

# And that's because OBJECT has included the object files with the unnecessary symbols!
$ nm test-static | grep f2
$ nm test-object | grep f2
00000000004004bc T _Z2f2v
```

GiR-Zippo:
Yep the bigger size caused by unused functions was a point I doesn't had in mind.
```
-rwxrwxr-x 1 user user 8640 Jun 30 02:08 test-object
-rwxrwxr-x 1 user user 8579 Jun 30 02:09 test-static
```

Indeed OBJECT links unused functions:
```
000000000040050c <_Z2f2v>:
  40050c:	55                   	push   %rbp
  40050d:	48 89 e5             	mov    %rsp,%rbp
  400510:	b8 02 00 00 00       	mov    $0x2,%eax
  400515:	5d                   	pop    %rbp
  400516:	c3                   	retq   
  400517:	66 0f 1f 84 00 00 00 	nopw   0x0(%rax,%rax,1)
  40051e:	00 00 
```

But all calls-opcodes are near calls in this example there should no impact on the execution speed. Cuz memory is accessed by the addressbus/memorycontroler (depends on the architecture)
```
00000000004004f6 <main>:
  4004f6:	55                   	push   %rbp
  4004f7:	48 89 e5             	mov    %rsp,%rbp
  4004fa:	e8 02 00 00 00       	callq  400501 <_Z2f1v>
  4004ff:	5d                   	pop    %rbp
  400500:	c3                   	retq  
```
If there where any far calls instead of near, so I would worry. Far-Calls are using 4 Bytes and therefore more cycles to processed.

But I have to agree with you, that a small file is better than a bigger one.

cyberium:
Ok so we'll stick at the actual way. Thank both of you.

PR #194, opened 2017-07-09
Linux OpenSSL 1.0 check

Darexon:
Arch Linux provides an OpenSSL-1.0 package but CMake doesn't use it. This patch fixes that.


CMakeLists.txt
```diff
@@ -251,6 +251,15 @@ if(UNIX)
       set(OPENSSL_INCLUDE_DIR /usr/local/opt/openssl/include)
     endif()
   endif()
+  # Some Linux distributions provide an OpenSSL-1.0 package alongside 1.1+
+  if (UNIX)
+    if (NOT OPENSSL_ROOT_DIR)
```
Warlockbugs: Checking for version number can also be useful:
`if (NOT OPENSSL_ROOT_DIR OR NOT OPENSSL_VERSION VERSION_LESS "1.1")`

CMakeLists.txt
```diff
@@ -251,6 +251,15 @@ if(UNIX)
...
+      set(OPENSSL_ROOT_DIR /usr/lib/openssl-1.0)
+    endif()
+    if (NOT OPENSSL_INCLUDE_DIR)
```
Warlockbugs: Same here

Warlockbugs:
Tested on Arch and i can confirm that this change is enough to play ball with installed OpenSSL 1.0 package. Thank you.

PR #200, opened 2017-07-24
Global unit ownership redesign

Warlockbugs:
# Functionality changes:
* Initial Pet, Guardian and Companion fields and flags are cleaned up according to gathered retail data.
* Unifies logical concept of unit ownership and it's fields; allows overrides within WorldObject hierarchy.
* Adjusted ownership API for Pets, Totems, Guardians and Companions.
* Added recursive owner API specifically for current implementations of IsFriendlyTo and IsHostileTo to resolve some issues with summons like elementals.

# Refactoring details:
* Introduced new term "Master", which resolves either to charmer or owner in one action, therefore:
  - GetCharmerOrOwnerGuid() becomes GetMasterGuid()
  - GetCharmerOrOwner() becomes GetMaster()
* Introduced a new term "Beneficiary" for cases when an instant action to get unit's charmer, owner or itself was required, therefore:
  - GetCharmerOrOwnerOrSelf() becomes GetBeneficiary()
  - GetCharmerOrOwnerPlayerOrPlayerItself() becomes GetBeneficiaryPlayer()
* Usage cases of following unit methods were carefuly examined and replaced with fitting flag checks instead; methods were removed as excessive as a result:
  - isCharmedOwnedByPlayerOrPlayer()
  - IsCharmerOrOwnerPlayerOrPlayerItself()
  - GetCharmerOrOwnerOrOwnGuid()

# Quality of Life changes:
* Client is now able to correctly detect and display totems as a result of all changes done.

# Video:
Demonstration of client being able to detect totems after these changes using nameplate options as an example.
[YouTube preview](https://youtu.be/jvetGKAOe3Q)

# Notes:
* All changes already stood trial on a live realm, so this is a peer review request for final merge.
* Classic and WoTLK ports, as usual, are ready and awaiting approval of this design.


killerwife:
It works. ^^

src/game/Entities/Unit.cpp
```diff
@@ -6054,44 +6058,78 @@ void Unit::ModifyAuraState(AuraState flag, bool apply)
...
+    Unit* beneficiary = GetBeneficiary();
+    if (beneficiary)
+        return (beneficiary->GetTypeId() == TYPEID_PLAYER ? static_cast<Player*>(beneficiary) : nullptr);
+    return (GetTypeId() == TYPEID_PLAYER ? static_cast<Player*>(this) : nullptr);
+}
 
-    return GetTypeId() == TYPEID_PLAYER ? (Player const*)this : nullptr;
+Unit* Unit::GetSummoner() const
+{
+    if (ObjectGuid guid = GetSummonerGuid())
```
cyberium: ObjectGuid const& guid (even if GetUnit argument will need to be changed too if we would make it more sence)

Warlockbugs: Guid struct still fits under POD and can be passed with no harm in this case, i agree from code style standpoint.

src/game/AI/BaseAI/GuardAI.cpp
```diff
@@ -138,6 +138,6 @@ void GuardAI::AttackStart(Unit* u)
 
 void GuardAI::JustDied(Unit* killer)
 {
-    if (Player* pkiller = killer->GetCharmerOrOwnerPlayerOrPlayerItself())
+    if (Player* pkiller = killer->GetBeneficiaryPlayer())
```
Warlockbugs: Since its usage mostly lies within XP/loot/honor rewards, I couldn't find a better term to describe the unit which behaves on behalf of another in this most common context.

Warlockbugs:
Regarding documentation, i can note that GetSummoner(), GetCreator(), GetCharmer() are getters for corresponding guid fields, while GetOwner(), GetMaster(), GetBeneficiary() are all additional logical constructs for the former.

src/game/AI/BaseAI/GuardAI.cpp
killerwife: @cyberium I had the same argument, but he is right, thats what it essentially does and I wasn't able to find a reason against that name, believe me, I tried ^^

Warlockbugs:
Bump. Anything else apart from naming? Naming debate can go on forever and it's easy to rename it at any later point if someone comes up with a better term or more use cases will be uncovered.

cyberium:
I see nothing bad. Go ahead for me.

src/game/Entities/Pet.cpp
```diff
@@ -615,6 +614,17 @@ void Pet::DeleteFromDB(Unit* owner, PetSaveMode slot)
     }
 }
 
+void Pet::SetOwnerGuid(ObjectGuid owner)
+{
+    switch (uint32(m_petType))
+    {
+        case SUMMON_PET:
+        case HUNTER_PET:
+            SetSummonerGuid(owner);
```
Exxenoz: I miss a break here. :-(

Warlockbugs: Added.

PR #201, opened 2017-07-31
Add optional Lockless Allocator

Warlockbugs:
Adds an option to build with alternative memory allocator, disabled by default. This is a much more lightweight option than previously used TBB and targeted mostly towards multi-threaded workloads.

- Comparisions done by developer: https://locklessinc.com/benchmarks_allocator.shtml
- Comparisions done by another project: https://www.andrew.cmu.edu/user/apodolsk/418/finalreport.html

The library source is released under GPLv3 license. 
Bundles CMake-ready source originating from a fork at: https://github.com/VladX/lockless-allocator/

I should come clean and say that i didn't do any benchmarks myself with it as of yet, this is just an experiment.

Tagging: @cyberium @namreeb @Laizerox @killerwife @stfx @VladimirMangos

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -1,5 +1,6 @@
 option(DEBUG                "Include additional debug-code in core" OFF)
 option(WARNINGS             "Show all warnings during compile"      OFF)
+option(USE_LOCKLESS_MALLOC  "Use lockless memory allocator"         OFF)
 option(POSTGRESQL           "Use PostgreSQL"                        OFF)
 option(PCH                  "Use precompiled headers"               ON)
 option(BUILD_GAME_SERVER    "Build game server"                     ON)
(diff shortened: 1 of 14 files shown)


namreeb:
If you want to consider other allocators, it is probably something worth discussing before doing.  I myself had always wanted to try jemalloc (https://github.com/jemalloc/jemalloc).  It is targeted to high performance and specifically avoidance of fragmentation.  It is used and maintained by people from Facebook and FreeBSD.

Warlockbugs:
I've checked a bunch of allocators (including jemalloc, hoard, but not limited to them) and most of them are not CMake-friendly or platform-friendly (skip entire platforms, windows and freebsd being most common sacrifices) at all. llalloc is the only one that appears portable and relatively painless to integrate, that is i tried first. This is just an optional experiment and not a proposal for default build setting.

namreeb:
There is an effort underway to transition jemalloc to the cmake build system, though it is not necessary in order to use it.  That effort can be found here: https://github.com/jemalloc/jemalloc-cmake/

cyberium:
@Warlockbugs hit a good point, Lockless have nice (better?) performance and it support all systems.
It seem however abandoned? Maybe the code is working pretty well now but what about the future?
Also the code base is smaller than jemalloc wich is nice.

Jemalloc cmake, its me or its not a priority for them? They have not merged last master change.
We highly depend of cmake and probably more in the future.
Also we can discuss about set the chosen one ON by default or not. If the improvement are real then i doesn't see why it should be disabled by default.

Warlockbugs:
The greatest argument against alternative allocators being set to default is that they are designed around speed and not security in mind, so usage can result in additional side effects (such as crashes) and should not be embraced as default ("safety first").

cyberium:
Sound like we are not ready for this. So better closing this for now.

PR #205, opened 2017-08-31
Crossfaction battlegrounds

Lillecarl:
Most things should work, if accepted and ported to WotLK we'd have to take achievements into account.

Very recommended addon for those who care to play battlegrounds with their friends no matter what faction they're on.


cyberium:
Thank you, but we have enough to deal instead of this new custom feature.

Warlockbugs:
As i've managed to get WotLK-like basic crossfaction relations part working properly (https://github.com/cmangos/issues/issues/596 and https://github.com/cmangos/issues/issues/1297) which allowed us to get full UI functionality support for crossfaction, this PR can now be resurrected for a second take.

Yatzii93:
So right now I have a version or a mix around of Lillie's work. Bloody brilliant stuff, really can't thank him enough. It's the sole reason pvp is a thing on my server. The one big issue to note that I haven't really got around to fixing is when a non-undead gets changed to an undead. 

Basically if changed to an undead you will be completely unable to use any form of chat, whereby you will recieve an error along the lines of "You do not speak that language". It has nothing to do with the actual language skills ingame. Having all the languages learnt does nothing to mitigate this issue.

I think @Lillecarl mentioned before it was a client side issue and he might of found a way around it. It's a bit of a jungle on his repos at this point :P so I've been unable to track it down. Would love for @Lillecarl to spread some wisdom here as it's one of the last remaining issues I experience.

Oh there is also the odd time where a player will be seen as friendly to the oposing team even though visually they have been cross-factioned. It's a pretty random rare bug that I haven't quite pinpointed it's origins. Relogging resolves the issue considering it reapplies the crossfaction code.

Yatzii93:
Ok I've settled for just removing undead as a possible faction change.

Lillecarl:
The problem you're experiencing with undeads are related to that we remove the language spell from them and learn a new one. What should instead be implemented is that translation happens serverside.

Lillecarl:
https://wowdev.wiki/DB/LanguageWords
There's basically everything required to implement serverside language translation, except the stormlib function SStrHash, which is required to get blizzlike results.

Daedalus-code:
How does someone implement this?
Where do I start xP

Lillecarl:
@Daedalus-code
https://github.com/Lillecarl/TrinityCore/commit/2b5a62ef0e3416fa7cdd2bdb25f4a19d3d612c47

This is a decent (needs cleanup) reference of what has to be done.
Be aware that language skill updates must be sent before ui reload and player invalidation opcode must be sent after ui reload.

Ui reload happens upon teleportation.

yuanf225:
So we still can't use crossfactionBG right now?

MantisLord:
@yuanf225 No, without a patch like the one in this PR, xfaction BGs are not possible. The core currently prevents queuing if your group contains both Horde & Alliance members.

All the other xfaction features in the mangos config work mostly okay though (groups/raid, guild. chat, trade, AH, etc).

I'd love to see this get some cleanup/love and merge eventually.

Lillecarl:
@MantisLord This was deemed too custom a long long time ago. If anyone with write access is interested I'd be happy to help upstream this though 😄

MantisLord:
@Lillecarl 
IMO this is no more custom than the rest of the TwoSideEnabled features already present/supported in the core and config.

I'd be curious to get @cyberium 's thoughts on this or anyone else who may be against.

AnonXS:
I mean there speaks nothing against implementing an option like this as long as it does not disrupt the correct blizzlike implementation, hence this feature needs to be built around present and future battleground code rewrites, not the other way round.

Lillecarl:
@AnonXS While this is far from perfect, we don't really change a lot to implement this, it's mostly just another queue inserter (the language skill parts should be thrown to hell,  It could be trimmed down way further on 2.4.3 and 3.3.5, but then the implementation would differ from 1.12.1 where there are no arenas and therefore no arenafaction field to set.

AnonXS:
It's comin though i think.

PR #207, opened 2017-09-01
Update PlayerbotAI.cpp

dunjeon:
Fixes bots not responding to attack commands from direct tell or in party chat.


cyberium:
look like @dunjeon code is right. Something missing while porting to the core?

dunjeon:
Mac build fails on travis-ci, but linux and windows both compile without error.

cyberium:
Don't worry about OSX now its know and will be fixed :)

cala:
Merged, thanks. :+1:

PR #208, opened 2017-09-03
Use vector of uints for spell list instead of cstrings

Lillecarl:


src/game/Chat/Level3.cpp
```diff
@@ -1399,615 +1399,611 @@ bool ChatHandler::HandleCooldownClearArenaCommand(char*)
...
-        "3930",
-        "3977",
-        "3925",
-        "136",
-        "228",
-        "5487",
-        "43",
-        "202",
-        "0"
+    static std::vector<uint32> allSpellList
```
Lillecarl: Should std::array be used instead maybe? And ofc the same on gmSpellList

Lillecarl: Agreed

killerwife: change it and i will merge this this week. I promise.

stfx: I fixed it for you

PR #209, opened 2017-09-12
Fixed Crash from Warlock AI bot

talamortis:


cala:
Thank you for this PR and pointing out this issue. I fixed it in a slightly different way to guarantee the warlock bot always pick a demon to summon (at least if one is known) .

talamortis:
np but it also did what you was explaining 👍

PR #212, opened 2017-10-05
Adjust DurationIndex 33897 to 33896

AnonXS:
https://github.com/cmangos/tbc-db/pull/138

Otherwise the npcs wont start attacking/moving again after 33896 expires.


killerwife:
This means it needs a script I think :) This spell was pulled from client data from 2.0.x so its as legit as it gets. I have seen this before.

Saltgurka:
Solved with https://github.com/cmangos/mangos-tbc/commit/f2f7348165de54523bc1df2c6917049d0be7063d

PR #216, opened 2017-11-11
Fix for the 'Column count doesn't match value count at row 1' error in sql/base/realmd.sql

berserkingyadis:
Since executing the file sql/base/realmd gives this error:

`ERROR 1136 (21S01) at line 73: Column count doesn't match value count at row 1
`

this PR adds default values for the (newly added) `token` colum of the table `realmd.account`. If other default values are wanted, ill change the PR.


killerwife:
I thought I included that, I will resolve all errors tomorrow, including the @NeatElves one

PR #217, opened 2017-11-17
[WIP] Quest: The Booterang: A Cure For The Common Worthless Peon

berserkingyadis:
Ill commit my progress on the booterang quest in this work-in progress PR.

Currently researching the sniff data.

### DIFF
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -820,6 +820,7 @@ UPDATE creature_template SET ScriptName='npc_spawned_oronok_tornheart' WHERE ent
 UPDATE creature_template SET ScriptName='npc_domesticated_felboar' WHERE entry=21195;
 UPDATE creature_template SET ScriptName='npc_shadowmoon_tuber_node' WHERE entry=21347;
 UPDATE creature_template SET ScriptName='npc_veneratus_spawn_node' WHERE entry=21334;
+UPDATE creature_template SET ScriptName='npc_disobedient_dragonmaw_peon' WHERE entry=23311;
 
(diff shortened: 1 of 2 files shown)


killerwife:
Check if there isnt a trigger creature at the position of those crystals. That way you might just make them move to the trigger creature. If there isn't one, you will likely need to make waypoints for each mob to follow after being hit with booterang (time consuming, but easy and inevitable).

berserkingyadis:
Have not heard of trigger creatures before, ill check out the db. 

Is there a way to find them with something like ".gob near"? Also, ive been searching for a way to query  db for gobjects, creatures, ... within a certain 2D Box boundary.

Does the coord-data from the sniffs translate to our coords? 
`Position: X: -5063.487 Y: 460.1463 Z: -11.3996`

killerwife:
There is no such way really. The coords in our DB might not be correct either. My latest work usually has sniff coordinates, but thats not true for all the data in the DB. If its not there, it needs to be added. Trigger creatures are creatures that are target of script stuff. For example TARGET_SCRIPT spells do use them often. Green beams across netherstorm use trigger creatures too.

Saltgurka:
To clarify, because this was something that took me a while to understand when I initially got into emu, trigger creatures are just normal creatures spawned in the world. The only thing differentiating them is that they are invisible except when in GM mode (uses hollow models) and they are usually used in scripts or as targets for spells as killerwife described above. As such there is no easy way to find them except to look in the world with .gm on or by checking in sniffs.

As for the coordinates you posted, yes you can insert them into our DB without the need for any conversion. They translate 1:1 to our coordinates

PR #219, opened 2018-02-16
Changed FndBoost.cmake to support new naming scheme of boost 1.66

DennisWG:
With boost version 1.66 the default naming scheme of the compiled library files changed. At least on Windows, it now contains the used architecture (32 or 64) to help choosing the right libraries.

This change manages the new file name layout in addition to the previous ones.


killerwife:
Is this official boost macro from latest cmake?

DennisWG:
It's not but I guess that would make sense, wouldn't it? I've updated it.

killerwife:
The protocol when making a pull request is that you rebase your branch and squash commits into lowest common denominator. Since all your commits are one thing, you should squash it into one commit and then force push into your branch.

PR #220, opened 2018-02-16
Implemented helper methods to convert between entities

DennisWG:
Implemented various helper methods to easily and safely convert between entities.

TrinityCore is using these since forever. They make code much simpler, safer and more readable.
I'll make heavy use of these functions in upcoming pull requests in order to clean-up the code base.

### DIFF
--- a/src/game/Entities/Object.h
+++ b/src/game/Entities/Object.h
@@ -569,6 +569,12 @@ class Object
 
         Loot* loot;
 
+        inline bool IsPlayer() const { return GetTypeId() == TYPEID_PLAYER; }
+        inline bool IsCreature() const { return GetTypeId() == TYPEID_UNIT; }
+        inline bool IsUnit() const { return isType(TYPEMASK_UNIT); }
+        inline bool IsGameObject() const { return GetTypeId() == TYPEID_GAMEOBJECT; }
+        inline bool IsCorpse() const { return GetTypeId() == TYPEID_CORPSE; }
+
     protected:
         Object();
 


killerwife:
Right now I don't want to comment on whether it will be merged or not, but why reinterpret_cast and not static_cast.

DennisWG:
I recalled that static_cast may cause slicing when used to down-cast. So I looked it up and it turns out that it causes slicing when up-casting.  I've changed the code to use static_cast instead, thanks for the hint!

DennisWG:
Okay, I've thought about it some more and no, I can't use static_cast here, since we've only forward declared the other types at this point. You might be able to work the code to be able to use static_cast but that's too much work for what we're achieving.

killerwife:
I am going to write a lengthy discussion based on what I, @Warlockbugs and @cyberium have said in regards to the topic.

First of all, I want to say that functions IsCreature/IsPlayer are completely fine.

When it comes to the casts, this solution was brought up years ago in cmangos and wasn't introduced for a reason. You are basically doing a checked cast which then you AGAIN need to check for nullptr.

In most cases in the core you are absolutely certain what the target type is, you just need a simple recast. In this case, this code will be 1 check slower than the current alternative. In addition to that, in those places, static_cast is an option.

@Warlockbugs first point was that reinterpret_cast is completely forbidden by his school of training. So the alternatives were to make it virtual and overriden, to which I said you might as well go with dynamic_cast.

What I want to illustrate here, is that casting in C++ is an important and very very complicated topic. By hiding it in a helper function, you offload the design decision on which cast to use onto the designer of the helper. This is not good. Even in C++14 you still need to mindfully use casts, and that can't be done generically, else there would not be 5 types of casts in C++.

So in the end, the IsCreature/IsPlayer can be merged, and the consensus is not to merge the rest due to the reasons explained above. There is no problem with merging TC code, but only if it is really needed.

Please, if you can, strip off the cast code, and push it into the same PR, without remaking the PR. It can be done using "git commit --amend" or "git reset HEAD^" and then "git push -f". The reason to not remake the PR is that it doesnt pollute the numbering. Numbers are shared among PRs and issues on github.

killerwife:
I only meant the bool functions, not the casts for player/creature. But yeah, you got the PR thingy right this time.

Lillecarl:
"reinterpret_cast<> is not guaranteed to result in the same bit pattern. "The mapping performed by reinterpret_cast<> is implementation defined." (C++03 5.3.10). However, the standard notes that "it is intended to be unsurprising". – Michael Burr Apr 30 '09 at 3:35"

From StackOverFlow.

ratkosrb:
> When it comes to the casts, this solution was brought up years ago in cmangos and wasn't introduced for a reason. You are basically doing a checked cast which then you AGAIN need to check for nullptr.

I am late to the party, but may i suggest something else to address this. Instead of having ToCreature as a member function in Object class, you simply make it a free function defined in the header like this.

```
inline Creature* ToCreature(Object* object)
{
    return object && object->GetTypeId() == TYPEID_UNIT ? reinterpret_cast<Creature*>(object) : nullptr;
}
```

This reduces all the checks you have to do to 1 if statement.

`if (my_creature_ptr = ToCreature(my_unit_ptr))
`

Assignment operator will return false if ToCreature returns nullptr, so you don't have to do any other checks.

killerwife:
The whole point is that you are doing redundant checks. I.e. in most cases when you are retyping, "object" is redundant, "object->GetTypeId() == TYPEID_UNIT" is redundant, and reinterpret_cast can be static_cast instead. And after that you still need to check the return value. Like for example in scripts, where you start an escort, you always know the starter GUID belongs to a player, so you never have to do typeid checks.

The main problem, is that this "hides" it behind a neat function, and delegates it away from the person who writes that code. And casting in C++ is a crucial decision making process. Once you add this neat function, everyone will use it without considering the redundant checks one has to make.

Lillecarl:
@ratkosrb While i'd love to see something like this in CMaNGOS, it's not the way things are done here, @killerwife provides some valid points. They would improve productivity for new people slightly, but would also introduce some overhead.

"goal must always be to provide the best code that we can." 😄

PR #222, opened 2018-02-17
Implement CMSG_GROUP_SWAP_SUB_GROUP

DennisWG:
This Opcode allows to directly swap the sub groups of two players who are in the same raid. It's a lot easier to just show you what tis does, so here are some videos:

Before:
https://www.youtube.com/watch?v=xrh8U1idatg

After:
https://www.youtube.com/watch?v=mAv8HTVTKKk

### DIFF
--- a/src/game/Groups/GroupHandler.cpp
+++ b/src/game/Groups/GroupHandler.cpp
@@ -886,3 +886,44 @@ void WorldSession::HandleOptOutOfLootOpcode(WorldPacket& recv_data)
     if (unkn != 0)
         sLog.outError("CMSG_GROUP_PASS_ON_LOOT: activation not implemented!");
 }
+
+void WorldSession::HandleGroupSwapSubGroupOpcode(WorldPacket& recv_data)
+{
+    std::string playerName1, playerName2;
+
+    recv_data >> playerName1;
+    recv_data >> playerName2;
+
+    Group* group = GetPlayer()->GetGroup();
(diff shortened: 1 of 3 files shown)


killerwife:
This wasn't implemented? Wow. Does this PR apply to both vanilla and wotlk?

DennisWG:
I don't see a reason why the packet structure should've changed but I only have the TBC client installed so I can't check.

src/game/Groups/GroupHandler.cpp
```diff
@@ -886,3 +886,44 @@ void WorldSession::HandleOptOutOfLootOpcode(WorldPacket& recv_data)
...
+        else
+        {
+            if (ObjectGuid guid = sObjectMgr.GetPlayerGuidByName(playerName))
+                return guid;
+            else
+                return ObjectGuid();
+        }
+    };
+
+    ObjectGuid guid1 = getGuid(playerName1);
```
Warlockbugs: Exiting the function as soon as either name query fails may be useful. One full name search can be avoided if first name is unknown.
```
if (!guid)
    return;
```

cala:
The feature is required in Classic and currently missing. I can't tell if the packet structure or even OPCode is the same thought.
But I don't see what changes occurred between Classic and TBC that would have require it to be different.

Warlockbugs:
Checked classic, packet structure is the same, so this patch can be applied without any alterations.

PR #223, opened 2018-02-28
Make Traps Visible to Owners

dunjeon:
The owner of a trap should be able to see it when placed (hunters)


dunjeon:
Also works on WoTLK, but untested on Classic

killerwife:
Sorry. The whole issue was caused by my mistake. I overlooked one crucial relations commit when exporting.

PR #225, opened 2018-03-21
Fix memleak in SessionMap and session duplication

Thema89:
This restores the original logic in that function and fixes #1177 #1252 and #1347 but will also suffer for #1359 problems:
yet, it's the good way to go. We need to fix that too asap.

### DIFF
--- a/src/game/World/World.cpp
+++ b/src/game/World/World.cpp
@@ -211,6 +211,8 @@ World::AddSession_(WorldSession* s)
             // prevent decrease sessions count if session queued
             if (RemoveQueuedSession(old->second))
                 decrease_session = false;
+            // do not remove replaced session from queue if listed
+            delete old->second;
         }
     }
 


killerwife:
As much as I would like to help on this matter, this is out of my area. I will try to ruffle some feathers, so that someone networking competent fixes this.

killerwife:
@cyberium said he will look at it this week. If he does not, @Thema89 Please write down instructions for a stupid person (me) how to reproduce and evaluate this PR.

Thema89:
@killerwife you can reproduce it very easily in few steps, here's how to reproduce that bugs and how to reproduce this PR fix:

--------------------------------------------------------------------------------
**Quickest way to reproduce #1177 player session memleak**

1. make and compile clean CMaNGOS master branch
2. login with an account (default 'player' account is enough, but an account with .tele permission is preferred) and create two characters: let's say Alice and Bob.
3. login with Alice and keep that game open in background
4. open another wow windows and login with the same account (this will disconnect the first client as intended)
5. login world with Bob and go to the same place where Alice was when the first client disconnected, and were Alice is **not** supposed to be anymore... you'll find your Alice character still inside the world (here's the memleak: that player unit shouldn't be in core memory anymore and surely must not be there) even if the first client has been disconnected. 
6. **bug reproduced!**
--------------------------------------------------------------------------------

**What happens if you login with the same character?** Many reported bugs, such as:
- player's unit won't despawn from world, and will remain in world until crash/reboot/shutdown
- player's unit is un-kickable too (as .kick closes player's session, but that unit's session is already closed!)
- empty bags, no money, no speaking languages
- rollbacks and many other DB "connections-lookalike" errors
- any other bug reported in that issues.

_That duplication is more obvious by using another characters, Bob, to see your Alice char still online. This is the vector that made me investigate and eventually let me found that the problem was in SessionMap pointers handling. More investigation in that function revealed that my PR is already implemented in other mangos-based project (such as TC and mangos itself) and that CMaNGOS is the only project who doesn't have that delete anymore._

 --------------------------------------------------------------------------------
**How to test this PR:**
- compile this PR and do the same steps described above. You won't be able to leave your character ingame anymore, you will never login again with empty bags, and so on: every bug in #1177 #1252 and #1347 and listed here will not be reproducible anymore: this fixes them all.
--------------------------------------------------------------------------------

_I'm testing it since February and I confirm that this simple PR definitively solves all of that problems.
So... yes, this solves 80% of problems, but this also let us know that there were **two** independent and similar bugs, and #1359 is **not** related with #1177 (and #1177 wasn't related to BOOST problems), as you can easily see that #1359 is still reproducible after this PR.
But we can continue with that bug in its proper issue comment section, I believe :D_

pkdetlefsen:
@cyberium I can currently reproduce [cmangos/issues#1359](https://github.com/cmangos/issues/issues/1359) by doing the following:

1. Start first wow client and log into account but stay at character selection screen.
2. Start second wow client and log into the same account.
3. Server crashes with the following error:

```
WorldSocket::HandleAuthSession: client build 5875, account TEST, clientseed AC0B1CBE
WorldSocket::HandleAuthSession: (s,v) check s: CFE5BB388AD495B4660CB2FAE0CB655915F9A21D16ACE1406422148E52AEA2F7 v: 3DCEB199F17B867276D66992FA60E21745AFEEBDD39D7B31C09A89AA0A69B81B
WorldSocket::HandleAuthSession: Client 'TEST' authenticated successfully from *.*.*.*.
Segmentation fault
```

It also crashes if you start logging into a character on client #1 and then log into the same account using client #2.

If the character is fully logged in on client #1 then the server won't crash when logging in using client #2 no matter if the character state changed or not.

System: Debian 9

I have applied this PR so I'm not affected by [cmangos/issues#1177](https://github.com/cmangos/issues/issues/1177) anymore.

Thema89:
I can reproduce what @pkdetlefsen reported (using Debian 9 too). But please notice that they are different problems, so I'd say #1177 is now handled properly. Now we need to investigate on #1359 ;)

PR #227, opened 2018-03-25
Mok Morokk Flags

kaimou1357:


Warlockbugs:
Please, use new specialized helpers instead of raw flags manipulations in such cases.
```
        void Unit::SetImmuneToNPC(bool state);
        void Unit::SetImmuneToPlayer(bool state);
```

PR #229, opened 2018-05-11
Commands: Add .gobject activate command.

henhouse:
This command was never in MaNGOS-based emus, but it is an extremely useful and honestly at times, necessary, command in order to work more easily and productively in-game. I feel it has a place to be added.

### DIFF
--- a/src/game/Chat/Level2.cpp
+++ b/src/game/Chat/Level2.cpp
@@ -1184,6 +1184,36 @@ bool ChatHandler::HandleGameObjectNearCommand(char* args)
     return true;
 }
 
+bool ChatHandler::HandleGameObjectActivateCommand(char* args) {
+    // number or [name] Shift-click form |color|Hgameobject:go_id|h[name]|h|r
+    uint32 lowguid;
+    if (!ExtractUint32KeyFromLink(&args, "Hgameobject", lowguid))
+        return false;
+
+    if (!lowguid)
+        return false;
+
+    GameObject* obj = nullptr;
+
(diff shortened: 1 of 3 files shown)


src/game/Chat/Level2.cpp
```diff
@@ -1184,6 +1184,36 @@ bool ChatHandler::HandleGameObjectNearCommand(char* args)
...
+        obj = GetGameObjectWithGuid(lowguid, go_data->id);
+
+    if (!obj)
+    {
+        PSendSysMessage(LANG_COMMAND_OBJNOTFOUND, lowguid);
+        SetSentErrorMessage(true);
+        return false;
+    }
+
+    uint32_t const autoCloseTime = obj->GetGOInfo()->GetAutoCloseTime() ? 10000u : 0u;
```
killerwife: supposed to be uint32
also specifying unsigned is redundant with such low numbers

henhouse: I honestly just adopted it exactly from Trinity without really thinking about the redundancy. Indeed, it's a bit odd since that function already returns a uint32 I'm not sure why he or she that implemented it cast them. Would you like me to update it?

killerwife: I am updating it on my local as we speak. It is also missing return true at the end.

henhouse: Oh my bad, I guess I was a bit hasty when throwing this together. Thanks for pointing these things out.

PR #231, opened 2018-05-13
Misc. group fixes

henhouse:
Fixes two relatively minor issues with groups.

- Raid assistants should not be able to kick the raid leader.
- GMs should have no restrictions when it comes to inviting a player to a group.

### DIFF
--- a/src/game/Groups/GroupHandler.cpp
+++ b/src/game/Groups/GroupHandler.cpp
@@ -258,6 +258,13 @@ void WorldSession::HandleGroupUninviteGuidOpcode(WorldPacket& recv_data)
     if (!grp)
         return;
 
+    // raid assistant cannot kick leader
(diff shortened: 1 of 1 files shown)


Warlockbugs:
GM invitation is more on the custom side of things, but i think we should allow it.

henhouse:
I can't imagine it would be. If players in DND can still receive whispers from GMs on retail (and is the case in cmangos), why would GMs be disallowed to help half of the player base on a realm? By this same logic, they should be disallowed to whisper members of the opposing faction, too. There's no way Blizzard would inconvenience their GM support staff by restricting their actions of messaging/grouping to aid them in their work.

That's my way of seeing it anyway, I just don't feel it makes sense from a GM point-of-view. They're characters in the world, yes, but they're not players. They should not be bound by the restrictions of faction limitations by any means.

PR #232, opened 2018-05-13
Pet: Clear arena auras when entering arena

henhouse:
Fixes the ability to hide non-arena buffs on a pet by dismissing them before entering the arena battle, and summoning them once inside.

### DIFF
--- a/src/game/Entities/Pet.cpp
+++ b/src/game/Entities/Pet.cpp
@@ -348,6 +348,10 @@ bool Pet::LoadPetFromDB(Player* owner, uint32 petentry /*= 0*/, uint32 petnumber
     // load spells/cooldowns/auras
     _LoadAuras(timediff);
 
+    // remove arena auras if in arena
+    if (map->IsBattleArena())
+        RemoveArenaAuras();
+
     // init AB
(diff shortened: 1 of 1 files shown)


killerwife:
Is this blizzlike behaviour tho? Are pets dismissed on entering arena?

AnonXS:
This is a fix to bypass sneaking in pets into arena battles which are fully buffed. i dont know when or how blizzard called clear auras on pets in arena battles (maybe repeated checks or it even checks dismissed pets and clears their auras).

killerwife:
@henhouse put this code AFTER _LoadSpells. That way no shenanigans due to spell scripts and whatnot can't occur.

henhouse:
Thanks for the extra information @AnonXS. Like he said, the fix is for sneaking in buffs into the arena. Your pet should have a clear state, just like the player does via `RemoveArenaAuras()`. Else, you could have a bunch of friends buff your pet, dismiss, enter the arena, and summon again with all the buffs present.

As for dismissing pet when you enter, I believe that is correct, yes, but I didn't check for that. I think they already dismiss and I am pretty sure that is correct behavior.

PR #233, opened 2018-05-13
Arena: Fixed arena quit/leave exploits

henhouse:
Fixes issue where you could leave or be kicked from an arena team while in queue or while in a match resulting in evading penality and/or crash.

### DIFF
--- a/src/game/Arena/ArenaTeamHandler.cpp
+++ b/src/game/Arena/ArenaTeamHandler.cpp
@@ -208,6 +208,9 @@ void WorldSession::HandleArenaTeamLeaveOpcode(WorldPacket& recv_data)
     if (!at)
         return;
 
+    if (_player->InBattleGroundQueue() || _player->InArena())
+        return;
+
     if (_player->GetObjectGuid() == at->GetCaptainGuid() && at->GetMembersSize() > 1)
     {
         // check for correctness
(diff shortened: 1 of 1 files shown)


Warlockbugs:
I wonder what is the server's response to this action would be on retail. Removing team from queue is aslo acceptable outcome. Did you by chance do any research on this?

henhouse:
I don't know how retail handled this, and considering teams are a thing of the past on retail now, it's not easy to know. The few videos we have as research nowadays will not likely show players intentionally trying to exploit and abuse on YouTube, so I don't think there's much material to research from.

We can only speculate, and by this, there are several approaches you could take. Should it remove the team from the queue instead? Sure, that's definitely a solution as well but requires some additional code to handle, and is [exactly how TrinityCore handles it](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/game/Battlegrounds/ArenaTeam.cpp#L325). Returning from the function call additionally works, and I thought was simply the easier fix. Correct me if I am wrong, but sadly there are no client message errors that relate to these issues, so it's safe to say they didn't proc an error message.

There's no way these exploits were possible and able to be abused on retail, there had to be some checks in place. It makes no sense to kick a member of your team while in a match with them. Many servers patched this issue over the years and for good reason. There's no way it was possible on retail. MaNGOS in 2009 still had the exploit where you could disband the team while in a match. I see that was fixed now, so someone recognized this was unallowed behavior.

If something more in the lines of TrinityCore is what should be the acceptable solution, let me know what you think and I will easily amend this PR. I didn't actually check TC before making this since it was based on existing work we patched nearly a decade ago. There's always many solutions to a single problem, and theirs seems a bit more proper these days, perhaps.

PR #234, opened 2018-05-15
Log: Add arena logging

henhouse:
Adds an arena log file (disabled by default) to log matches and the stats of the match, team creation, disbandment, etc.

**PR Note**: I combined the "ArenaExtendedLog" feature into one. If people disagree and think these should still remain separate, let me know. I just don't see why not to have it included these days. The "extended" (on/off) feature for those who may not be aware was is simply the addition of the individual player stats and info in the match. Eg: player info, damage done, healing done, kills, IP address, etc. Without this enabled it normally would only report team creation & disbandment, and match started/match ended. No player stats.

### DIFF
--- a/src/game/Arena/ArenaTeam.cpp
+++ b/src/game/Arena/ArenaTeam.cpp
@@ -97,6 +97,7 @@ bool ArenaTeam::Create(ObjectGuid captainGuid, ArenaType type, std::string arena
     CharacterDatabase.CommitTransaction();
 
     AddMember(m_CaptainGuid);
+    sLog.outArena("New ArenaTeam created [Id: %u] [Type: %u] [Captain GUID: %u]", GetId(), GetType(), GetCaptainGuid());
     return true;
 }
 
@@ -184,6 +185,8 @@ bool ArenaTeam::AddMember(ObjectGuid playerGuid)
         // hide promote/remove buttons
         if (m_CaptainGuid != playerGuid)
             pl->SetArenaTeamInfoField(GetSlot(), ARENA_TEAM_MEMBER, 1);
+
+        sLog.outArena("Player: %s [GUID: %u] joined arena team type: %u [Id: %u].", pl->GetName(), pl->GetGUIDLow(), GetType(), GetId());
     }
     return true;
 }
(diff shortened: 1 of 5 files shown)


killerwife:
While I agree with @Warlockbugs on the fact that Grepping is a solution I also do run a server. The way I could see this merged is by making it for all Battleground related things, not just arenas. Can you do that @henhouse or is there a specific reason why the arena file has so few entries for logging.

henhouse:
Indeed it can, but then why is there already such separation between log files already? From experience, these logs files can get immensely large unless you begin to break them up. I see CMaNGOS has a few features to create individual logs (per day?) as one solution on some. 

In most cases, basic users are not going to know what grepping is, nor how to easily navigate a behemoth file if everything is output into one. Separation gives clarity and is likely why MaNGOS kept character logs away from Server.log when it was implemented. TrinityCore also implemented arena logging as a separate file for the same reasons.

I believe it's somewhat good to follow the example of other similar projects once a pathway for a useful administrative feature has been paved. Many come to expect similar solutions to be implemented similarly or entirely. I personally am not a massive fan of the entire reworks they did for logging, but the lack of some arguably critical logs is certainly a motivation for adoption.

henhouse:
Ah sorry @killerwife, you responded right as I was writing.

It's definitely possible you could merge them under one "battleground" log file, but even I would have a preference for them to remain separate, but that's just my personal taste. I am trying to think of what BG log files would have to offer... You have a BG stats feature, which is really nice! So the only thing a log would give edge on is player joining/leaving info... perhaps it could be extended to do flag caps, etc. which would then be useful for potential abuse review cases. But if you have a server with active BGs/Arena it will be difficult to follow the file sequentially by looking at it if they were in one... My personal opinion would be to have a Battleground.log file then as well. That's just me, though.

killerwife:
Hmmm, okay. I don't entirely disagree, especially since I have experience with going through insanely big files. One thing that does indeed come to mind is making this "file creation business" a bit more automated through an array of files and enums. You would still have a function for it for ease of access, but the other logic could be automated. Just stating my improvements here.

Lillecarl:
If I were to put my own saying into this I'd prefer seeing logging done into a database table where we could set up a few basic columns of information such as:
date, event type, accountid (can be null), playerid( (can be null), xyz (o?) I'd argue it's more powerful then grepping, considering we already have a database connection to a structured database, might just use some structure? :)

killerwife:
The problem for me with that is performance and volume of logging and lack of tooling for it. I usually deal with logs size of several megabytes, which would eventually hog my SSDs. But thats my perspective, maybe someone else has other ideas.

Lillecarl:
@killerwife If you have lots of players and debug logging on it's gonna generate a lot of data indeed, but we shall not underestimate the speed of pg or MySQL. (especially with async inserts)

How many MB/s are we talking for you guys on average?

cyberium:
Hello, iam not against more log of any sort but i would prefer to use LOG FILTER system. It will add log to main log but it will be possible to disable/enable it in game without to restart the server.

For now i will close this if someone else realy need it well see.

PR #237, opened 2018-07-06
Fix Rend Blackhand gate if Gyth is killed before..

BillytheOrc:
Rend Blackhand

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/shadowmoon_valley.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/shadowmoon_valley.cpp
@@ -17,7 +17,7 @@
 /* ScriptData
 SDName: Shadowmoon_Valley
 SD%Complete: 100
-SDComment: Quest support: 10451, 10458, 10480, 10481, 10514, 10540, 10588, 10781, 10804, 10854, 11020.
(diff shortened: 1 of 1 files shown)


killerwife:
1. Tabs
2. How would it be possible to skip it. Thats the critical issue here instead. It sounds like some faction change or flags or something missing.

BillytheOrc:
Sorry you don't like my dirty tabbing </3

It's possible to skip it at level 70 if people were to dualburst the boss with a big frostbolt or something.

BillytheOrc:
Oh, it should also be mentioned that Rend Blackhand needs a C++ script to start an instance encounter. Without that, this script always opens the gate. 

I'll work on a Rend script today, assuming I can fix my ugly tabbing :p

killerwife:
You polluted the PR with another commit. You need to make one branch per one PR and never PR from master branch.

Next, as mentioned on reddit, until Gyth summons Rend, Gyth needs to be invulnerable, and stop at 1 hp. It can be done using DamageTaken and examples are all over the core. Please clean up the PR and adjust the script accordingly.

cyberium:
This was abandoned so i'll just put a reminder to @killerwife if he is interested to take something from this.

PR #240, opened 2018-07-16
Substantial EAI rework

killerwife:
List of changes:
Fixes two bugs:
1. When EAI timer expires in combat, it should not reset until the cast is successful - now in line with SD2 == CAST_OK workflow and sniffs
2. Fixes issue when phase changes can chain each other when in subsequent events of same type - found by Grz3s on Spell hit event

New features:
Adds ranged mode which is meant to fully replace dynamic movement
It enables caster like behaviour, with several settings and chase distances
It eliminates stutter step, responds to silence and kick and is able to distance itself from the enemy
Adds support for proximity aoe spells like frost nova, which should only go off if someone is in range
Adds 2 new event parameters enabling some behaviour in LOS event and in the future propably more
Also makes event target more consistent across different events in regards to action invoker

Notes:
Core contains changes which will need to be applied to the ACID files in DB repos.
Will start with tbc-db and once a smooth transition is made will be ported to classic and wotlk.

### DIFF
--- a/doc/EventAI.txt
+++ b/doc/EventAI.txt
@@ -1,5 +1,5 @@
 ===========================================================================
-MaNGOS EventAI (Creature_AI) Documentation: (Last Updated: August 22, 2012)
+MaNGOS EventAI (Creature_AI) Documentation: (Last Updated: June 25, 2018)
 ===========================================================================
 
 EventAI allows users to create new creature scripts entirely within the database.
@@ -7,6 +7,7 @@ EventAI allows users to create new creature scripts entirely within the database
 For the AI to be used, you must first make sure to set AIname for each creature that should use this AI.
 UPDATE creature_template SET AIName = 'EventAI' WHERE entry IN (...);
 
+Note: Edit in Spaces, not Tabs
 
 =========================================
 Basic Structure of EventAI
@@ -28,6 +29,8 @@ event_param1                    Variables for the event (depends on event_type)
 event_param2                    Variables for the event (depends on event_type)
 event_param3                    Variables for the event (depends on event_type)
 event_param4                    Variables for the event (depends on event_type)
+event_param5                    Variables for the event (depends on event_type)
+event_param6                    Variables for the event (depends on event_type)
 
 action1_type                    Action #1 to take when the Event occurs (see "Action types" below)
 action1_param1                  Variables used by Action1 (depends on action_type)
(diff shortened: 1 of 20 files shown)


killerwife:
https://github.com/cmangos/tbc-db/tree/rework/EAI

killerwife:
@cala @cyberium @evil-at-wow and @Grz3s requesting opinion

AnonXS:
Hey @cala ,

All classic and tbc "casters" and "archers" have already been reworked, currently working on preparing a version with wotlk npcs for @Grz3s .

EVENT_T_RANGE only checks for current victim, TARGET_T_NEAREST_AOE_TARGET checks for poential targets in range to for example cast frostnova while chasing current victim, rooting other players/enteties in place. Its an overall smarter check for AoE Targeting Spells as they are then only executed when it actually hits something.

When i understand @killerwife correctly EVENT_T_RANGE  was only added due to the fact that spellrange values were not able to be taken into account befor and even then we will probably still need this Event in the future as there are npcs casting a MAIN_SPELL, which has for example a range of 0 to 40. but they do not use this spell when under a certain range limit (e.g 8yards, 5yards) so handling this in this event isnt actually that bad and prevents us to consistently check for range for vicitim to change phase to a phase where he doesnt cast this Spell.

Personally not too sure about the event_param5,6 additions, but i guess this was the only way to implement additional checks for other parameters for ONE Event. (recieve emote event)
-> maybe think about refactoring the Event befor adding additional columns to make it work.

killerwife:
@cala sorry I missed this in my summer havoc. The 2 param additions are due to LOS event not being scriptable enough, and us having to move it to C++ for that reason. Splitting it in the future is indeed a valid thing, but I was trying to tackle bugs here and it already outgrew all I wanted to do. Which leads me to why it took me 2 months to come back and start merging it (what I am doing when writing this comment).

The issue with guardians should be lowered, I think, because I refactored the code, ending up with much much less code duplicity.

TARGET_T_NEAREST_AOE_TARGET - uses different distance calculation. Single target uses 3D source combat reach + target combat reach. AOE uses source nothing target combatreach. In this manner it mimicks spell system targeting. Some spells go off regardless of whether there is a target nearby, but I found that this is not the case for at least some frost nova spells on some creatures.

Orders:

This is still a major WIP on my part, and it is not covered by this PR, but its coming in a more extensive way after it. The way I am implementing it, is that I am generalizing everything to UnitAI when possible. Atm its not possible for distancing, bcs of overriding MovementInform, but I will figure that one out eventually. Examples of AI wide events are call for help, fleeing and so on. In SD2 its working out for me splendidly, but that doesn't solve some of the most custom things, and its not portable to EAI in any reasonable manner due to the issue I call "assignment problem" (you cant assign to a variable in DB script code). Depends on what orders you have in mind.

Also I understand its a bit hard to read, but it got even more garbled in the STFX code maintenance, so I apologize for that.

PR #241, opened 2018-07-28
[s2375] Implement trainer greetings

Ghaster:
The goal of this commit is to properly handle SMSG_TRAINER_LIST, by adding trainer specific greetings at the top of the trainer window. Beforehand, this string was determined by LANG_NPC_TAINER_HELLO, but now an unique greeting for every trainer can be added in 'trainer_greeting' table.

Credits to Wyreth (https://github.com/TrinityCore/TrinityCore/issues/20493) for finding this issue originally and for finding alot of the greetings.

The greetings lines themselves will be added in another commit. But for now, you can see the changes in the images below:

Before:
before

After
after

### DIFF
--- a/sql/base/mangos.sql
+++ b/sql/base/mangos.sql
@@ -23,7 +23,7 @@ DROP TABLE IF EXISTS `db_version`;
 CREATE TABLE `db_version` (
   `version` varchar(120) DEFAULT NULL,
   `creature_ai_version` varchar(120) DEFAULT NULL,
-  `required_s2374_01_mangos_quest_template_rewmaxrep` bit(1) DEFAULT NULL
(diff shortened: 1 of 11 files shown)


killerwife:
Whooooaaaa, this is so cool. I will try to review asap.

killerwife:
Okay, looking through the changes, for future reference, when making a PR don't do a revision change yourself and let the person who is merging it handle that. Right now the argument is moot, because there are not too many DB revision changes, but it could potentially conflict with each other in high traffic occasions.

I am gonna go test this on my local and then probably merge it.

Ghaster:
Alrighty. It should be noted that this commit is also valid for Classic and WotLK.

killerwife:
Ofc, I will do that in a moment.

PR #242, opened 2018-07-29
Improve compile time

FastBlit:
Improve compile time by caching unchanged object files if ccache is available on the system, otherwise skips this optimization.

Must be set ahead of the project header, pulled cmake min required out also to perform the check on ccache (currently requires cmake 2.8). The setting for linker is ignored.

As documented here:
https://www.virag.si/2015/07/use-ccache-with-cmake-for-faster-compilation/
https://crascit.com/2016/04/09/using-ccache-with-cmake/


killerwife:
Thx for the PR. I was busy so it took me some time to read up on it.

PR #243, opened 2018-08-09
Fix CMaNGOS non PCH build on High Sierra

Lillecarl:
PCH is still broken because unclear reasons.

### DIFF
--- a/src/shared/Database/DatabaseMysql.h
+++ b/src/shared/Database/DatabaseMysql.h
@@ -27,6 +27,8 @@
 
 #ifdef _WIN32
 #include <mysql/mysql.h>
+#elif __APPLE__
+#include <mysql/mysql.h>
 #else
 #include <mysql.h>
 #endif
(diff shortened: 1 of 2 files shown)


killerwife:
@cala any comments on this? I don't have apple development.

Lillecarl:
I should note that i'm building with gcc

src/shared/Database/DatabaseMysql.h
```diff
@@ -27,6 +27,8 @@
 
 #ifdef _WIN32
 #include <mysql/mysql.h>
+#elif __APPLE__
```
Laizerox: Why don't you just replace #ifdef and #elif with `#if defined(_WIN32) || defined(__APPLE__)` ?

Lillecarl: Why why is because i didn't know you could do that with preprocessor macros 😄

Lillecarl:
I'll have to revisit this, apple ecosystem doesn't seem to be very consistent across CI and actual OS.

PR #244, opened 2018-08-17
Fix NPC's Spawning with 1 hp

Baynar:
After commit @1388de1 creatures spawn with 1 hp and regenerate to full because table uses 1 hp as default where new value should be 0

If this doesn't get update many exploits with people killing 1 hp bosses will arrive 🗡


killerwife:
Well we made a blanket fix in the DB. This PR is nonetheless valid, but there is a butt. This won't fix anything for existing ones, since those do not fall back to default 0.

Baynar:
Well just did this quick fix cause after that updates my players reported mobs spaning with 1 hp

Gonna update the rest of the creatures too since this fixes only newly created onces

killerwife:
We already did that. https://github.com/cmangos/tbc-db/blob/master/Updates/0001_curhealth_curmana.sql

Baynar:
hmm then maybe the auto updater didn't update mine sorry :)

PR #245, opened 2018-09-22
Fixed drop_db_mysql.sql when executed in multiple transactions

joao-fidalgo:


cyberium:
Done another way, but please add a correct description next time.

PR #253, opened 2018-10-26
Fix: #1705 tbc sap

Pooch11:
Fixed [TBC] Rogue's SAP snares creatures when used followed by an spell
Had to resubmit this pull to build off the latest build changes instead of my own (Ops)

After countless hours of looking where this problem could lie and testing different configurations, I finally found a simple fix that allows mobs to move after being incapacitated. It seems as though there is no UNIT_FLAG enum that defines an incapacitated state. For the future I think that should be implemented to easier break this condition.

Related issue can be found [here](https://github.com/cmangos/issues/issues/1705)

### DIFF
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -1652,8 +1652,11 @@ void CreatureEventAI::UpdateAI(const uint32 diff)
             else
                 SetCombatMovement(true, true);
         }
-        else if (m_meleeEnabled && !m_currentRangedMode)
-            DoMeleeAttackIfReady();
+		else if (m_meleeEnabled && !m_currentRangedMode)
+		{
+			SetCombatMovement(true, true);
+			DoMeleeAttackIfReady();
+		}
     }
 }
 


killerwife:
I am afraid that this does not solve the underlying issue. The issue is not just specific for EAI. The real issue exists in SetIncapacitatedState and the renewal of movement. I will try to address the issue today and I will notify you to check that this PR is no longer needed.

Pooch11:
@killerwife  Thank you for your input. I had tried to start a discussion about this issue via the issue thread but it seemed pretty dead and I did not get any response. I really want to contribute, but I initially found navigating around for the problem was an even bigger hurdle for a new contributor. This was the only solution I found.

I will also look into a SetIncapacitatedState and see if I can come up with something more robust.

killerwife:
I am sorry, I didn't get to your thing, and now I got sick. I will try to push something and link it today.

killerwife:
@Pooch11 https://github.com/cmangos/mangos-tbc/commit/523941d3d1093e009bfb7b3da35b831784626091
Please check this out.

cyberium:
@SilvioDoMine confirmed that its working now.

PR #255, opened 2018-11-12
Implement null checking for all cstrings loaded from database and ret…

Lillecarl:
…urn empty string instead of null.

Fixes cmangos/issues#1722


Lillecarl:
Cheers mate

PR #258, opened 2018-12-10
Remove duplicated faction column on mangos.sql

joao-fidalgo:
Fixes issue [1799](https://github.com/cmangos/issues/issues/1799) by removing a duplicated column while doing an insert into `creature_template`. The origin of this problem is related to the removal of FactionHorde column commited in [0cb338](https://github.com/cmangos/mangos-tbc/commit/0cb338cd7b20732c541691b3729fb3fe7da5865d).


dvdvideo1234:
This needs to be merged for the sake of my TBC server :D
https://github.com/cmangos/mangos-tbc/blob/900051fb74749aa01cba95d04c4a5ef9dc05fbb3/sql/base/mangos.sql#L1356

PR #260, opened 2018-12-11
Earth Shield Aggro should go to the target being healed and not the caster

LevelOneX:
Earth shield in TBC shouldn't pass either triggeredByAura or originalCaster, so the aggro goes directly to the target.


killerwife:
I will deal with it once I finish exports and some small bugs. If not remind me in 2 weeks.

PR #263, opened 2019-01-25
Core/Script: Fix Shade of Aran - Blizzard

EricksOliveira:


EricksOliveira:
is this the right way @killerwife

killerwife:
No. You can see that SPELL_SUMMON_BLIZZARD triggers SPELL_CIRCULAR_BLIZZARD      in its C++ script in SpellEffects.cpp. The fix is to figure out why the code is not reached.

Phatcat:
```
case 29969:                                 // Summon Blizzard
{
    if (unitTarget)
    unitTarget->CastSpell(nullptr, 29952, TRIGGERED_OLD_TRIGGERED, nullptr, nullptr, m_caster->GetObjectGuid());

    return;
}
```

Spell looks kind of weird in spelleffects, is it supposed to be cast by the target? Is the spell not supposed to have a target?

killerwife:
Shade of Aran - Karazhan
No, it is correct and supposed to be cast by target. 29952 has TARGET_UNIT_CASTER which targets himself.

Phatcat:
```
case 29969:                                 // Summon Blizzard
{
    if (unitTarget)
        unitTarget->CastSpell(unitTarget, 29952, TRIGGERED_OLD_TRIGGERED, nullptr, nullptr, m_caster->GetObjectGuid());

    return;
}
```

Perhaps, then? I dunno, it's not working atm but works when using DoCastSpellIfCan(m_creature, SPELL_CIRCULAR_BLIZZARD, CAST_TRIGGERED); so I kind of wanna blame the nullptr target passing in spellEffects as that's the only real clear difference, then.

killerwife:
No, TARGET_UNIT_CASTER is never ever meant to have targetfill. No matter if it works one way or another, its not meant to be there. The issue is probably somewhere else. Debug it and see where it fails to go forward.

cyberium:
Not correct fix

PR #264, opened 2019-02-09
HashMapHolder: switch from mutex to shared_mutex

Phatcat:
Go from mutex to shared_mutex and utilize shared_lock for reads and unique_lock for writes. Allows for multiple owners of the mutex (in read cases).

Edit* Uses boost::shared_mutex indstead now for backwards compatibility
std::shared_mutex is part of c++17 so switch to that once we... eventually support c++17..

Fixes https://github.com/cmangos/issues/issues/787 and https://github.com/cmangos/issues/issues/1355

### DIFF
--- a/src/game/Globals/ObjectAccessor.cpp
+++ b/src/game/Globals/ObjectAccessor.cpp
@@ -73,8 +73,7 @@ ObjectAccessor::~ObjectAccessor()
     }
 }
 
-Unit*
-ObjectAccessor::GetUnit(WorldObject const& u, ObjectGuid guid)
+Unit* ObjectAccessor::GetUnit(WorldObject const& u, ObjectGuid guid)
 {
     if (!guid)
         return nullptr;
@@ -147,8 +146,7 @@ void ObjectAccessor::KickPlayer(ObjectGuid guid)
     }
 }
 
-Corpse*
-ObjectAccessor::GetCorpseForPlayerGUID(ObjectGuid guid)
+Corpse* ObjectAccessor::GetCorpseForPlayerGUID(ObjectGuid guid)
 {
     Guard guard(i_corpseGuard);
 
@@ -161,8 +159,7 @@ ObjectAccessor::GetCorpseForPlayerGUID(ObjectGuid guid)
     return iter->second;
 }
 
-void
-ObjectAccessor::RemoveCorpse(Corpse* corpse)
+void ObjectAccessor::RemoveCorpse(Corpse* corpse)
 {
     MANGOS_ASSERT(corpse && corpse->GetType() != CORPSE_BONES);
 
@@ -181,8 +178,7 @@ ObjectAccessor::RemoveCorpse(Corpse* corpse)
     i_player2corpse.erase(iter);
 }
 
-void
-ObjectAccessor::AddCorpse(Corpse* corpse)
+void ObjectAccessor::AddCorpse(Corpse* corpse)
 {
     MANGOS_ASSERT(corpse && corpse->GetType() != CORPSE_BONES);
 
@@ -197,8 +193,7 @@ ObjectAccessor::AddCorpse(Corpse* corpse)
     sObjectMgr.AddCorpseCellData(corpse->GetMapId(), cell_id, corpse->GetOwnerGuid().GetCounter(), corpse->GetInstanceId());
 }
 
-void
-ObjectAccessor::AddCorpsesToGrid(GridPair const& gridpair, GridType& grid, Map* map)
+void ObjectAccessor::AddCorpsesToGrid(GridPair const& gridpair, GridType& grid, Map* map)
 {
     Guard guard(i_corpseGuard);
     for (auto& iter : i_player2corpse)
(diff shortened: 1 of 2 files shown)


Levitanious:
In case it's not super clear: This is for cmangos/issues#787

Phatcat:
... How old is the travis compiler? shared_mutex is standard since gcc 6.0 (seems like travis is using gcc 4.8.4 https://api.travis-ci.org/v3/job/490819018/log.txt)... do I really need to use boost::shared_mutex? Can travis be updated, or...?

Laizerox:
Shared mutex was introduced in C++17. (See: **https://en.cppreference.com/w/cpp/thread/shared_mutex**)

Iirc cmangos doesn't support above C++11 yet.

Phatcat:
So boost:shared_mutex it is then?

Phatcat:
Cmangos could support c++17? Is there anything worth considering holding it back? I'd rather not introduce more usage of boost unless there's no way around it..

ulfgebhardt:
https://github.com/cmangos/mangos-tbc/blob/master/.travis.yml#L66

killerwife:
Yes. @cala is holding us back on C++11 and some people using VS2015 are holding us back on C++14.

Phatcat:
So boost it is then...

cala:
> Yes. @cala is holding us back on C++11 and some people using VS2015 are holding us back on C++14.

Huh ? I'm no more holding us back to C++11 than most of Mac users out there. 

And though I don't use VS, as far as I know this is a paying tool and we should not expect developers nor even users to pull money out of their pocket to upgrade their software (and sometimes hardware) if they wish to keep contributing or using CMaNGOS. Not everyone can have the latest version provided by his/her company nor is eligible to educational licences of the software. 
There may also be valid reasons to not always upgrade tools like VS that are generally used in a professional environment.

Should I give my personal opinion, I find that it would be paradoxal to provide an emulator for a fifteen years old game while requesting users to have top of the notch versions of today's softwares. Especially as the game clients are unlikely to receive any updates and therefore are doomed to slowly become incompatible with current computers. For example, Classic and TBC clients for Mac will stop working with latest mac OS and Macs starting as soon as September of this year. So upgrading versions is never a straightforward decision to make.

Phatcat:
I can fully appreciate what you're saying @cala, I'm not at all against backwards compatibility. I've changed it to boost::shared_mutex

Phatcat:
> And though I don't use VS, as far as I know this is a paying tool and we should not expect developers nor even users to pull money out of their pocket to upgrade their software
(...)
Not everyone can have the latest version provided by his/her company nor is eligible to educational licences of the software.

In the particular case of VS you can always get a free license for open source development, though, but that doesn't change things for mac, though, no.

>An unlimited number of users within an organization can use Visual Studio Community for the following scenarios: in a classroom learning environment, for academic research, or for contributing to open source projects.

https://visualstudio.microsoft.com/vs/community/

Phatcat:
What's with the hold-up? Not sure to merge this or not?

Cmangos has something called 'writeguard' and 'readguard' and treats them both as readguards. Already here it's a little strange, as there is no reason what so ever to differentiate them in name (read vs write) when they are functionality wise both read guards, locking out anything else from accessing it whatsoever (hence the current crash with .saveall)
If you are against shared mutexes and their added overhead, then there's 0% need for any differentiation in name at all.

So there are 3 options from here; first option is to keep it as it is and not care about the crashing (potentially just remove the command .saveall), second option is to merge this as it is and deal with the added overhead of shared vs. non shared mutexes. Third option is to rewrite the system so that a lock can never be called twice on an unresolved stack (like it is now with .saveall), but otherwise stick to non shared mutexes. Both the first and third option kind of call for the renaming and restructuring of locks as to not deceptively lead the user to believe there actually is a difference between them and that a writeguard won't lock you out of a read (which it will)

Third option is the best option, but also the least likely to happen in the foreseeable future.

cyberium:
Thank for this PR as i see you worked hard on the issue to find those possibilities.
3 is my prefered choice ofc because after some digging it sound like its more a design problem than a mutex problem.

Your point about read and write guards are valid. I rememeber guy that was trying to deal with crash caused by FindPlayer() and trying to optimise it so we can continue to read it even if the writeguard was locked. Well that was not a good idea and as i see the code was not fully discarded. So basically it need a couple of renaming and cleaning to get rid of that code.

So now in Player.cpp we have
```cpp
// save pet (hunter pet level and experience and all type pets health/mana except priest pet).
    if (Pet* pet = GetPet())
        pet->SavePetToDB(PET_SAVE_AS_CURRENT);
```
The question is can we safely add the owner to  Pet::SavePetToDB method?
Something like
```cpp
pet->SavePetToDB(PET_SAVE_AS_CURRENT, this);
```
Iam not 100% sure because in Pet::SavePetToDB we use Pet::GetOwner() wich is not trivial code. Pet class look simple at first view but it is not.

Phatcat:
Seems like pet->SavePetToDB() is indeed looking for it's owner, which it shouldn't have to, as pets, at least pets whenever they are saved to the db, are always checked if they have a player owner prior. the main reason SavePetToDB is looking for it's owner is to confirm that it's a player... it's totally redundant if passed along in the call.

Is this the only case where the mutex locks and there's an attempt to read? or was the guy you remember working on solving another, perhaps similar issue?

I suggest considering this pr here instead: https://github.com/cmangos/mangos-classic/pull/390 which is based on your suggestion @Cyberium and eliminates the issue as well, but before closing this PR so it's not forgotten, perhaps rename and slightly alter the mutex lock system as to not give the impression of a shared mutex.

cyberium:
Good job @Phatcat. I'll ask you however to test it in different case to be sure its ok and also if it not bother you much to port it to other core (i can do it if you dont want to/cant do it.)

PR #267, opened 2019-02-16
Add new config variable to Ignore key for heroic dungeons.

HyperionTBC:


killerwife:
Fix tabs and force push.

evil-at-wow:
Instead of adding another config variable, wouldn't it be easier to adjust your database to your liking? I believe the key requirements are in the database - I think that's heroic_key/heroic_key2 of areatrigger_teleport - so you could easily get rid of them with a simple query. You could even adjust the requirements per dungeon if you wanted.

killerwife:
I talked to him and explained that this is essentially duplicate behaviour anyone can customize for themselves.

PR #269, opened 2019-03-03
fixes phantom casting on some spell cast items

MrMarvin:
When using the [Netherweave Net](http://db.hellfire-tbc.com/?item=24269) item, **the item would be consumed, no matter if the related spell effect had a target or not.** It should have failed with 'You have no target' or 'Invalid target'.

In tests with similar items it was concluded that both Netherweave Net items lack the proper client-side validations and always invoke `Player::CastItemUseSpell`.

**Reproduce**
```
# Macro
/target player
.setskill 197 325
.additem 24269 1
.additem 4941 1
.cooldown clear
/cleartarget
/use Netherweave Net
```

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -4997,6 +4997,18 @@ SpellCastResult Spell::CheckCast(bool strict)
     {
         for (int j = 0; j < MAX_EFFECT_INDEX; ++j)
         {
+            // There are some items that do not reliably filter client-side if they are usable
+            // i.e. a proper target exists.
+            if (m_spellInfo->EffectImplicitTargetA[j] == TARGET_UNIT_ENEMY ||
+                m_spellInfo->EffectImplicitTargetB[j] == TARGET_UNIT_ENEMY)
+            {
+                if (m_caster->GetTarget())
+                    // Explicitly set implicitTarget so other checks have something to work with
+                    m_targets.setUnitTarget(m_caster->GetTarget());
+                else
+                    return SPELL_FAILED_BAD_IMPLICIT_TARGETS;
+            }
+
             if (m_spellInfo->EffectImplicitTargetA[j] == TARGET_UNIT_SCRIPT_NEAR_CASTER ||
                     m_spellInfo->EffectImplicitTargetB[j] == TARGET_UNIT_SCRIPT_NEAR_CASTER ||
                     m_spellInfo->EffectImplicitTargetA[j] == TARGET_LOCATION_SCRIPT_NEAR_CASTER ||


src/game/Spells/Spell.cpp
```diff
@@ -3064,6 +3064,17 @@ SpellCastResult Spell::SpellStart(SpellCastTargets const* targets, Aura* trigger
         return SPELL_FAILED_SPELL_IN_PROGRESS;
     }
 
+    // There are some items that do not reliably filter client-side if they are usable
```
killerwife: Checks like this belong to Spell::CheckCast and generally should use the "cycle" with MAX_EFFECT_INDEX paradigm instead of 3 hardcoded checks.

MrMarvin:
Thanks a lot for the review @killerwife! I moved the checking logic to `Spell::CheckCast`. There were already some other conditions based on a spells implicit targets, so its fitting quite well, I guess.

Also as we discussed in discord, while I agree that this is considered a client bug, I do not think that mitigating these packages on a per-opcode basis makes sense. For each `Player::CastItemUseSpell` call we would need to fully unwrap the Item->Spell{1,2,3}->Effect{1,2,3}/ImplicitTarget{1,2,3} tree. That's basically re-implementing a whole lot of core functionality. Therefore I decided to keep the checks as deep to the actual spell cast as possible, by checking each spell call for incomplete target information.

While testing locally I found some additional implications of the incomplete spell call. While we could easily filter out spell casts with no target, there were related issues with the rest of `Spell::CheckCast`. Such as casts with a friendly target never properly range checked (i.e. the range check for enemy target only happened client side -> cheat possibility). This has been worked around by explicitly setting the `unitTarget` to the casters current target.

The current state is as follows:

* Cast without target -> 'You have no target'
* Cast on enemy -> casts properly
* Cast on unit further than the spells max range -> 'You are too far away'
* Cast on friendly npc -> 'Invalid target'
* Cast on self -> player gets rooted...

I was not yet able to find any evidence for the later two points if this is "blizzlike" or not.

MrMarvin:
Hi there! Is there anything more for me to do with this PR to move it forward? 
Thanks!

killerwife:
Hm. Didn't I talk to you on discord about this, or was it someone else?

MrMarvin:
> Didn't I talk to you on discord about this

Yes, we discussed this issue one to one in Discord. As mentioned [above](https://github.com/cmangos/mangos-tbc/pull/269#issuecomment-469049874), I evaluated your suggestions to work around the client side quirks on the outer most layer, i.e. op code handling functions, but came to the conclusion that handling this in the actual spell casting logic seems to be the way to go. Please see my above comments.
I do understand that this might make it more difficult to merge my fix with your larger spell-casting overhaul. Please let me know if there is anything I can do to assists getting it all together!

killerwife:
I have adapted my recent targeting rework to also resolve this issue. I will close this PR myself when I export it. (more of a note to cyberium)

killerwife:
Fixed in https://github.com/cmangos/mangos-tbc/commit/b2761a1fa33932e20871d7de689e48f7bf09df9e

PR #271, opened 2019-03-08
1892 abolish poison effect

Almightie:
- added UNIT_TARGET_FRIEND to allow the effect to have a target


cala:
Makes sense to me.

I can't find right now of a spell where this should be false, so this would be harmless to merge. @killerwife ?

cala:
Merged, thank you!

PR #274, opened 2019-03-27
Sweeping Strikes stance change

Yatzii93:
Update to stop Sweeping Strikes from dropping on stance change

### DIFF
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -6240,7 +6240,7 @@ void Aura::HandleShapeshiftBoosts(bool apply)
         Unit::SpellAuraHolderMap& tAuras = target->GetSpellAuraHolderMap();
         for (Unit::SpellAuraHolderMap::iterator itr = tAuras.begin(); itr != tAuras.end();)
         {
-            if (itr->second->IsRemovedOnShapeLost())
+            if (itr->second->IsRemovedOnShapeLost() && itr->second->GetSpellProto()->Id != 12328)  // Stop Sweeping Strikes dropping on stance change
             {
                 target->RemoveAurasDueToSpell(itr->second->GetId());
                 itr = tAuras.begin();
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -6240,7 +6240,7 @@ void Aura::HandleShapeshiftBoosts(bool apply)
         Unit::SpellAuraHolderMap& tAuras = target->GetSpellAuraHolderMap();
         for (Unit::SpellAuraHolderMap::iterator itr = tAuras.begin(); itr != tAuras.end();)
         {
-            if (itr->second->IsRemovedOnShapeLost() && itr->second->GetSpellProto()->Id != 12328)  // Stop Sweeping Strikes dropping on stance change
+            if (itr->second->IsRemovedOnShapeLost())
             {
                 target->RemoveAurasDueToSpell(itr->second->GetId());
                 itr = tAuras.begin();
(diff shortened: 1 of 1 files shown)


robinsch:
Just add it here https://github.com/cmangos/mangos-tbc/blob/master/src/game/Spells/SpellAuras.cpp#L7525

5man999:
> Just add it here https://github.com/cmangos/mangos-tbc/blob/master/src/game/Spells/SpellAuras.cpp#L7525

So just turn false instead of true?

robinsch:
Yes

5man999:
Cool if its all good from my end I'll amend this pull request.

killerwife:
@Yatzii93 are you supposed to be able to have sweeping strikes in defensive stance after switching? because its tagged Stances: 0x00050000 (FORM_BATTLESTANCE, FORM_BERSERKERSTANCE)

Yatzii93:
@killerwife Good question! Not sure, will need to verify.

killerwife:
Ok, when you verify and provide me evidence, I will deal with this PR.

AnonXS:
wotlk - https://youtu.be/622rVgbM6sI?t=164
atleast the aura is persistant. - related trinitycore commit: https://github.com/TrinityCore/TrinityCore/commit/8c3d0e23689f8cbe4f4d7381dd9ebe9a905adfa8

killerwife:
The commit says what I said. Only allow moving to berserker, not defensive.

killerwife:
Due to noone providing evidence saying that it should be kept in defensive stance, closing this.

Phatcat:
https://www.wowhead.com/forums&topic=25679/sweeping-strike-retention

"No... the effect will remain even if you change stances... the same is true about un-equipping your shield after activating Shield Wall... but since Concussion Blow is rather an effect than an actual attack (because it causes no dmg), it won't proc twice on different targets..."

killerwife:
Can you please find video evidence too?

Phatcat:
Right here, the gool 'ol classic Pat PVP;

https://www.youtube.com/watch?v=RGBnjELkgok

Watch at 4:03 - 4:08 - he is in battle stance, drops sweeping strikes and immediately switches to berserker while retaining sweeping buff.

killerwife:
Berserker is confirmed, I was either looking for shield wall in battle/berserk or sweeping in defensive. Sorry if my question was not clear.

Phatcat:
Or are you looking for evidence about sweeping strikes and defensive stance specifically?

That seems like 2 different issues in 1.

First issue; Sweeping strikes should persist through stance change.

Second issue, is sweeping strikes allowed in defensive.

First issue seems independent of the second issue.

killerwife:
Yes, except first one is not an issue, because I tested that it already works.

Phatcat:
well, again, this guide also talks about the effects of defensive stance on sweeping strikes; https://hilbert.wordpress.com/sweeping-strikes-guide/

"At the dawn of time, Sweeping Strike was a simple but limited skill. It would literally act as a third arm hitting an additional target whenever you wanted to. Its biggest drawback was that it was removed when you changed stance which made it impossible to use together with whirlwind, something that was presumed to counter the possibility of SS + whirlwind being too powerful. This did not change until much later.

...

"What is not mentioned in the patch notes is that patch 1.7 also introduced the greatest change to SS, namely allowed it to remain while changing stances (just like the “big 3” which ironically were mentioned in the same patch notes)."

...

"Thus, the calculations are correct and SS does indeed get a second enrage bonus. Our assumption was correct.

What this also means is that defensive stance will incur the 10% penalty twice in SS as well, but luckily that’s not a big deal since you won’t combine those two very often."

- Coming them wouldn't be possible at all if the buff was removed by switching to defensive.
But no, I do not have video evidence, and good luck with finding it, if such evidence was ever recorded at all.

Perhaps turn to 1.13 for answers? Sweeping strikes was a thing in classic and if it worked like that in classic/tbc it should still work like that in 1.13

killerwife:
I do not have access to any of that stuff right now I'm afraid.

killerwife:
https://www.youtube.com/watch?v=uzmYAv5YEso&feature=youtu.be
https://www.youtube.com/watch?v=ilZywfh4NuQ&feature=youtu.be
Will implement blanket solution for all warrior spells.

PR #275, opened 2019-04-08
Alchemy Addition

Yatzii93:
Using AzerothCore as a reference.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/world/npc_professions.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/world/npc_professions.cpp
@@ -397,7 +397,7 @@ void SendActionMenu_npc_prof_Alchemy(Player* pPlayer, Creature* pCreature, uint3
 		break;
 		//Unlearn Alchemy
 	case GOSSIP_ACTION_INFO_DEF + 4:
-		if (pPlayer->HasSpell(S_POTION))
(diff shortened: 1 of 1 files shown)


5man999:
Whats missing is correct cost, right now it's using the cost for blacksmithing/leatherworking - I'll correct it at some point, if I don't forget :P

evil-at-wow:
I don't think this PR is needed. We used to have this in our scripts (which might be where AzerothCore got it in the first place?), but it was removed because that gossip is now handled in the database. See scriptdev2/scriptdev2-tbc@b25943682b41 for details (and note that scriptdev2 was integrated in the core somewhere early 2017).

Did you add this because the gossip was missing, or not working as intended?

5man999:
It wasn't working. It just took the gold cost so I figured it wasn't implemented.

evil-at-wow:
Starting an issue is a good idea either way I think. That makes it a known issue that can be investigated. 
As far as I'm concerned, this PR can then be closed because I think it's somewhat unlikely it will be merged, but if it remains open while the issue is being investigated, I don't mind either.

killerwife:
All of this exists in DBScripts. There are particular spells for unlearning all this. It was removed as a result of it being changed to dbscripts. Please go look up scripts that use 41563, 41564 and 41565 in dbscripts (_on_gossip) I believe to confirm my memory @Yatzii93 .

PR #276, opened 2019-04-12
Chat command .learn all_gm only teaches spells of catergory internal

schell244:
When entering .learn all_gm (with a security level 3 account) in the chat, the character should only learn spell from category "Internal" (-> good for testing etc). 
I just gathered the most common spell ids from this category - might not be all yet.


killerwife:
Dealt with even in greater detail in https://github.com/cmangos/mangos-tbc/commit/7f61fc2fd409e57c95c2743bdc811e78c94e26dd

PR #278, opened 2019-04-17
Warlord Najentus Needle Spell Effects + Targeting

Yatzii93:
boss_warlord_najentus.cpp Black Temple
-SPELL_NEEDLE_SPINE_TARGET(39992) targeting.
-SPELL_NEEDLE_SPINES(39835) spell effect.

According to info online the spell hits targets in a cone area so I think there needs to be more targeting logic to make this pure blizzlike, this is a start though.


killerwife:
This PR is incomplete, because it needs a whole lot more code. I will export this when I export Najentus on next export cycle. Closing it for now, should be done by me in next month. The critical part here is to have the boss rewrite coincide with fixing this.

PR #279, opened 2019-04-17
Totemic Mastery Fix (Skyshatter Tier 6 - 2 set)

Yatzii93:
Before changes the spell case couldn't be reached in the script. This meant Totemic Mastery(38443) did nothing in terms of applying mod increases and the effect Dummy Trigger(18350) would cause anything the player targeted to instantly aggro onto said player.

Let me know if this is a pepega approach, happy to learn.


killerwife:
@Yatzii93 pls, look at the commit I've made. You are doing good work, I am just pushing it up a notch.

Yatzii93:
> @Yatzii93 pls, look at the commit I've made. You are doing good work, I am just pushing it up a notch.

Not a problem at all, please feel free to change any of my contribution however you please. I'm not a programmer by trade so I expect there to always be better ways. Ty for the kind works too :)

PR #282, opened 2019-04-18
Can Not Interrupt Hurricane From Druid And NPC

Gelori:
Those two spells were totally missing InterruptFlags,
NPC could continue to cast even moving, added standard
15 flag
Issue:
	https://github.com/cmangos/issues/issues/1865


killerwife:
I hate to be THAT guy, but (and especially since I am late to the party) but I bet one of these: ChannelIF 0x00007C0C is gonna have something to do with that.... Cos thats a whole load of flags right there.

killerwife:
After further examination, tranquility has them. Will merge.

BenDol:
out of curiosity, should an update file be produced for changes like this for `sql\updates\mangos\`?

killerwife:
Nope, this is the correct workflow.

BenDol:
so these changes need to be manually applied? I normally just run `InstallFullDB.sh` for all my mangosd changes :p

killerwife:
No, they are automatically applied if you have CORE_DIR setup properly.

BenDol:
Ah okay, nice thanks

PR #286, opened 2019-05-26
Removed data blob in item_instance

henhouse:
Removed the data blob left in item_instance as [removed in TrinityCore](https://github.com/TrinityCore/TrinityCore/commit/77fc91bde85b44298ab6694236d1f6c9ece3cd34) back in 2010.

The `uint32toint32` SQL function for data migration does not appear to work as intended anymore, however, and I need some advice or help on how to go forward with it. I am assuming that in 2010 when this commit was done in TC it was possible to somehow get MySQL to legally overflow an unsigned integer and cast it to signed so that the end result was a -1, but in more modern versions nowadays you're met with an out of bounds error. This is just my guess as to why this was possible before. I am unsure how to proceed with the migration script due to this.

The charges as stored as max uint32 values in the data blob currently and should be -1 as an end result.

Application into the classic and WotLK branches should be relatively similar as WotLK needs 1 additional field and classic is the same, I think, minus the guild bank related code.

### DIFF
--- a/src/game/AuctionHouse/AuctionHouseMgr.cpp
+++ b/src/game/AuctionHouse/AuctionHouseMgr.cpp
@@ -280,8 +280,8 @@ void AuctionHouseMgr::SendAuctionExpiredMail(AuctionEntry* auction)
 
 void AuctionHouseMgr::LoadAuctionItems()
 {
-    // data needs to be at first place for Item::LoadFromDB 0  1        2
-    QueryResult* result = CharacterDatabase.Query("SELECT data,itemguid,item_template FROM auction JOIN item_instance ON itemguid = guid");
+    // data needs to be at first place for Item::LoadFromDB 0        1            2                3      4         5        6      7             8                 9           10          11        12
+    QueryResult* result = CharacterDatabase.Query("SELECT itemEntry, creatorGuid, giftCreatorGuid, count, duration, charges, flags, enchantments, randomPropertyId, durability, itemTextId, itemguid, item_template FROM auction JOIN item_instance ON itemguid = guid");
 
     if (!result)
     {
@@ -300,8 +300,8 @@ void AuctionHouseMgr::LoadAuctionItems()
         bar.step();
 
         Field* fields = result->Fetch();
-        uint32 item_guid        = fields[1].GetUInt32();
-        uint32 item_template    = fields[2].GetUInt32();
+        uint32 item_guid        = fields[11].GetUInt32();
+        uint32 item_template    = fields[12].GetUInt32();
 
         ItemPrototype const* proto = ObjectMgr::GetItemPrototype(item_template);
 
--- a/contrib/iteminstance_converter/README.md
+++ b/contrib/iteminstance_converter/README.md
@@ -1,0 +1,22 @@
+# Item Instance data converter
+Tool to convert the datablob in the item_instance table to extracted format.
+https://github.com/henhouse/ItemInstanceConverter
+
+## Instructions & Usage
+
+##### Warning:
+This script should be run **BEFORE** applying the character-db SQL update. This tool will extract
+the current data inside the `item_instance` table and prepare it in the new format. It will output `INSERT`
+statements for each item to be put back in after you apply the cmangos character-db update, which will `TRUNCATE`
+the table.
+
+BACKUP YOUR CHARACTER DATABASE!
+
+##### Usage:
+- Download and install the latest version of [Go from Google](https://golang.org/dl).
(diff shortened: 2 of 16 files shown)


killerwife:
Well. I do not know how much of a pain it would be to make a conversion program in a more sophisticated language like python or C++. It should be possible to do some shenanigans using casting in SQL, but let me tell you, its HORRID to work with.

henhouse:
That would be the best way to deal with this. But is there any precedence for using a separate program to help with data migration before? I don't know how we'd provide such an update or give instruction for everyone to follow it accordingly.

I am most proficient in Go and could easily write a nice program to do this migration, but then the question is... do we bundle a compiled one for each platform, instruct people to install Go and run or compile it?) Same issue would exist with Python, I guess.

Laizerox:
@henhouse for node there's pkg npm package that bundles javascript code, node and v8 engine iirc. It produces binaries for osx, linux and windows out of the box. There's lots of different options how to make migration app :)

killerwife:
There is not, but I doubt anyone will mind in the midst of this. I personally made a C++ tool when I was doing spell DBC shenanigans, but there is plenty of C# stuff and so on. The only real requirement is that it's easy to setup, which in the case of python is super easy.

ratkosrb:
Here is a simple tool that splits the data field into human readable columns.
https://github.com/vmangos/ItemInstance/blob/master/ItemInstance/ItemInstance.cpp

Laizerox:
I forgot... we're living "it probably has an app" era :D

+1 @ratkosrb

killerwife:
@henhouse adapt that, add it to the core, test it and finish the draft. If possible.

henhouse:
Didn't know that existed, thanks @ratkosrb.

Will take a look when I have some free time again and try to finish the draft.

henhouse:
Sorry for the delay with getting back to this. @killerwife @Laizerox 

I took ratkosrb's project and adapted it for TBC, however when it came to making it cross-platform compatible and the varying MySQL includes, incompatible typedefs, etc. I was getting frustrated with trying to make each environment happy in Cmake, and since I have less experience with that I opted to fall back on another language I felt would be easier to deliver cross-platform compatibility with Go. (Instead of Python simply because I'm way more proficient in it, and the steps to setup are insanely simple).

I've updated the PR but it is lacking the tool I speak of. Here is the version made to work universally for each platform (Go can build for nearly any platform): https://github.com/henhouse/ItemInstanceConverter

The tool is quite performant as I parsed 10 million entries within roughly 30-40 seconds on my local machine (i7 7700k), so it feels adequate enough for this task.

Can you please advise if this looks alright and if so, if I may drop it into `contrib`? At that point I'll be ready to wrap up this PR draft. I've included usage of the tool in its README file and a run script to run the script.

MantisLord (approved):

henhouse:
Fixed a few remaining cases in previous commit thanks to @MantisLord for noticing. No longer set to draft and is ready for review.

Please feel free to use [this test character DB](https://gist.github.com/henhouse/de028f2575a4e68de4a07f0104685890) to verify the conversion. It tests each extracted scenario to check before/after conversion with.

Specifically testing cases:
- Item creator
- Gift creator
- Stack count
- Duration of buff (see sharpening stone)
- Spell charges (see consumable)
- Flags (binding & wrapped)
- Enchantments
- Random properly
- Text ID from mailed message
- Durability

One can do more extensive testing, but I wanted to provide this template to make it easier.

src/game/AuctionHouse/AuctionHouseMgr.cpp
```diff
@@ -280,8 +280,8 @@ void AuctionHouseMgr::SendAuctionExpiredMail(AuctionEntry* auction)
 
 void AuctionHouseMgr::LoadAuctionItems()
 {
-    // data needs to be at first place for Item::LoadFromDB 0  1        2
-    QueryResult* result = CharacterDatabase.Query("SELECT data,itemguid,item_template FROM auction JOIN item_instance ON itemguid = guid");
+    // data needs to be at first place for Item::LoadFromDB 0        1            2                3      4         5        6      7             8                 9           10          11        12
+    QueryResult* result = CharacterDatabase.Query("SELECT itemEntry, creatorGuid, giftCreatorGuid, count, duration, charges, flags, enchantments, randomPropertyId, durability, itemTextId, itemguid, item_template FROM auction JOIN item_instance ON itemguid = guid");
```
MantisLord, line 284: @henhouse why change this? we only need two fields here. same goes for all the queries really, I don't see a need to add the new fields in where they aren't being used

henhouse:
Rebased and squashed all into 1 commit per @MantisLord's request.

PR #287, opened 2019-05-26
Fixes the crash in quest 10451

BenDol:
Resolves https://github.com/cmangos/issues/issues/1593)

This also removes a crash case in boss_supremus.cpp of this same problem.

Wasn't sure if `PetAI::UpdateAI(const uint32 diff)` requires the `Clear(false)` call still?

### DIFF
--- a/src/game/AI/BaseAI/PetAI.cpp
+++ b/src/game/AI/BaseAI/PetAI.cpp
@@ -392,7 +392,7 @@ void PetAI::UpdateAI(const uint32 diff)
             {
                 if (following)
                 {
-                    m_unit->GetMotionMaster()->Clear(false);
+                    m_unit->GetMotionMaster()->Clear(false);    // does this need to be removed since MoveFollow is calling Clear?
                     m_unit->GetMotionMaster()->MoveFollow(owner, m_followDist, m_followAngle);
                 }
             }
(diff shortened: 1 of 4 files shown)


Yatzii93:
Is..... m_unit->GetMotionMaster()->Clear(false); the cause of the crash? Apolgoies if I'm mistaken. I only ask because theres a lot more scripts that use this function. 52 from my search.

Yatzii93:
I applied the cleanup of PetAI and nothing amiss so far from my testing.

BenDol:
Only reason I didn't remove it was because it had the false flag provided, which I assumed could be necessary. I think the MoveFollow needs to be reworked to ignore cases where Clear has already been called. Perhaps before invoking the Clear() we check if it is already empty?

src/game/AI/BaseAI/PetAI.cpp
```diff
@@ -392,7 +392,7 @@ void PetAI::UpdateAI(const uint32 diff)
             {
                 if (following)
                 {
-                    m_unit->GetMotionMaster()->Clear(false);
+                    m_unit->GetMotionMaster()->Clear(false);    // does this need to be removed since MoveFollow is calling Clear?
```
killerwife: This one is a tad tricky, because currently PetAI is used also by players who are MCed and creatures who are MCed and they might have a loooot of different movement going on.

PR #288, opened 2019-05-26
Avoid using Soul Stones when you already have a Healthstone

BenDol:
Resolves https://github.com/cmangos/issues/issues/1910

Still fairly new to cmangos so let me know if there are any improvements.

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -3236,11 +3236,7 @@ void Spell::cast(bool skipCheck)
     SpellCastResult castResult = CheckPower();
     if (castResult != SPELL_CAST_OK)
     {
-        SendCastResult(castResult);
-        SendInterrupted(castResult);
-        finish(false);
-        m_caster->DecreaseCastCounter();
-        SetExecutedCurrently(false);
+        StopCast(castResult);
         return;
     }
 
@@ -3250,11 +3246,7 @@ void Spell::cast(bool skipCheck)
         castResult = CheckCast(false);
         if (castResult != SPELL_CAST_OK)
         {
-            SendCastResult(castResult);
-            SendInterrupted(castResult);
-            finish(false);
-            m_caster->DecreaseCastCounter();
-            SetExecutedCurrently(false);
+            StopCast(castResult);
             return;
         }
     }
@@ -4381,7 +4373,7 @@ void Spell::TakeAmmo() const
 
 void Spell::TakeReagents()
 {
-    if (m_caster->GetTypeId() != TYPEID_PLAYER)
+    if (!m_caster->IsPlayer())
         return;
 
     if (IgnoreItemRequirements())                           // reagents used in triggered spell removed by original spell or don't must be removed.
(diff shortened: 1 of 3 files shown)


killerwife:
Correct idea, bad execution.
Suggested course of action:
1. move the check to the switch (m_spellInfo->Id) inside spell::checkcast at the complete end. This is in line with other spells. look for the correct error which should be displayed, i assume its gonna be smth along the lines of "alrdy have too many of those"
2. the refactor inside EffectDummy needs to be reverted. Why? Soon (understand this year) I plan to add nicely hooked spell scripting, and that script will be then moved there. It will no longer be part of spell class, so this just throws logs under our feet for doing that.

BenDol:
@killerwife thanks for the help, Spell system was a little bit of a maze so I appreciate the direction. As for the `GetUsableHealthStoneItemType` method. I added this because I need to check the health stone type in the check so I either duplicate the same code or move it to a new method. You still want me to revert this?

killerwife:
Hm, aren't healthstones stacking globally? Or you can have rank 1 2 3 4 5 etc.

BenDol:
The original code didnt account for rank, but its possible.

killerwife:
Well then, go find out. I do not have a clue. But it seems important.

BenDol:
> However, the Echoes of Doom Patch nerfed this ability, and now Healthstones in General are unique items, no matter the Ranks of Improved Healthstone the Warlock had.
[Echoes of Doom (3.0.2)](https://wowwiki.fandom.com/wiki/Patch_3.0.2)

killerwife:
Well then. Keep the helper function, but add functionality for it to work like before the described patch notes. I will deal with the refactoring once the new spell scripting is added. After this is merged, I will ask you to make a wotlk version of this functionality.

BenDol:
What do you mean add functionality for before the patch notes?

killerwife:
Okay, that means that the PR is then ready for merging in TBC. Will do a final review later today/tomorrow and merge it.

src/game/Spells/Spell.cpp
```diff
@@ -5951,6 +5943,19 @@ SpellCastResult Spell::CheckCast(bool strict)
...
+        case 6202:
+        case 6201:
+        case 5699:
+        {
+            // check if we already have a healthstone
+            uint32 itemType = GetUsableHealthStoneItemType(m_caster);
+            if (itemType && m_caster->IsPlayer() && ((Player*)m_caster)->GetItemCount(itemType) > 0)
+                return SPELL_FAILED_TOO_MANY_OF_ITEM;
+            break;
+        }
```
Ve1vet: @killerwife @BenDol 
Maybe move this under this line? https://github.com/cmangos/mangos-tbc/blob/80d2abeb9aa7c5aedff744b8568231feb296bcdb/src/game/Spells/Spell.cpp#L4652
```c++
case SPELLFAMILY_WARLOCK:
  if( m_spellInfo->SpellFamilyFlags & uint64(0x0000000000100000) ){
    
    // check if we already have a healthstone
    uint32 itemType = GetUsableHealthStoneItemType(m_caster);
    if (itemType && m_caster->IsPlayer() && ((Player*)m_caster)->GetItemCount(itemType) > 0)
      return SPELL_FAILED_TOO_MANY_OF_ITEM;
  }
  break;
```

killerwife: I disagree with the mask approach, because then it wont be transferrable to spell scripting easily. It will have to be assigned to specific spell IDs.

Ve1vet: @killerwife i don't understand you at all... 😞 m_spellInfo already contains Id inside anyway. And instead 6 checks it will check only once. This is just a check optimization speed.
And how this check can be related to spell scripting? And more bigger question, about what spell scripting are you talking about? I feel myself dumb right now...

killerwife: The scripting that is not implemented yet. I am merely thinking 6 months ahead. I am sure stuff will be making more sense soon™.

PR #289, opened 2019-05-29
Fix a crash in SendGold for issue 1947

BenDol:
## 🍰 Pullrequest
Crashes when you open the quest reward (bag of silver). The issue is that its calling `ForceLootAnimationClientUpdate` after already calling `Release` which handles the client update when required. Instead we ignore `ForceLootAnimationClientUpdate` when releasing in `SendGold`.

### Proof
https://github.com/cmangos/issues/issues/1947

### Issues
https://github.com/cmangos/issues/issues/1947

### How2Test
create an alliance character
.quest add 116
.quest complete 116
.go creature "Barkeep Daniels"

### Todo / Checklist

### DIFF
--- a/src/game/Loot/LootMgr.cpp
+++ b/src/game/Loot/LootMgr.cpp
@@ -1147,7 +1147,7 @@ void Loot::SetPlayerIsNotLooting(Player* player)
     }
 }
 
-void Loot::Release(Player* player)
+bool Loot::Release(Player* player)
 {
     bool updateClients = false;
     if (player->GetObjectGuid() == m_currentLooterGuid)
@@ -1251,7 +1251,7 @@ void Loot::Release(Player* player)
         {
             Corpse* corpse = (Corpse*) m_lootTarget;
             if (!corpse || !corpse->IsWithinDistInMap(player, INTERACTION_DISTANCE))
-                return;
+                return false;
 
             if (IsLootedFor(player))
             {
@@ -1302,7 +1302,7 @@ void Loot::Release(Player* player)
                     break;
                 }
             }
-            return;                                         // item can be looted only single player
+            return false; // item can be looted only single player
         }
         case HIGHGUID_UNIT:
         {
@@ -1376,6 +1376,8 @@ void Loot::Release(Player* player)
 
     if (updateClients)
         ForceLootAnimationClientUpdate();
+
+    return updateClients;
 }
 
 // Popup windows with loot content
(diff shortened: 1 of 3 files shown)


Yatzii93:
Ok so it seems the problem right now with this PR is in dungeons party members are getting shown that mobs are lootable client side when infact they are not server side, I guess it's because we havent told their clients to update.

BenDol:
ah so I'll need to investigate what part of `ForceLootAnimationClientUpdate` is having the issue.

BenDol:
@Yatzii93 latest commit should resolve those issues, thanks for catching that!

src/game/Loot/LootMgr.cpp
```diff
@@ -2187,9 +2189,8 @@ void Loot::SendGold(Player* player)
     }
     m_gold = 0;
 
-    if (IsLootedFor(player))
-        Release(player);
-    ForceLootAnimationCLientUpdate();
+    if(!IsLootedFor(player) || Release(player))
```
killerwife: missing space between if and brackets

killerwife:
You did a good job solving this, but my spidey sense is tingling, because it reminds me of an issue I was trying to fix with aura holders and I ended up wasting months on it.
I am angry at two things:
1. item code uses "delete this" and thats like next level of design heresy
2. all other object deletions are delayed, but items are immediate. This is not the only place in the core we will have issues. For example when using a different allocator the issue will not even surface.

I am open for a discussion on how to solve this permanently, the PR is mergable, but I think we might want to avoid the code complexity this imposes on the loot code, since it is already fkin complex. If we do not reach a good conclusion, I will merge it anyhow.

BenDol:
You're talking about the `Release` code, right? I agree it's awful. I searched all the instances where `Release` and `ForceLootAnimationClientUpdate` are being called to make sure it wasn't happening else where. But yeah I likely needs to be reworked some what. This will at least band aid the crash instances until then.

Yatzii93:
Still seem to be having the same issue occur just not as frequent.

BenDol:
You're 100% sure that its caused by this change?
I tested it and wasn't able to reproduce it. Do you know a way to reproduce the issue?

BenDol:
@Yatzii93 okay I have tailored this to only affect the use case that crashes the server. You shouldn't see any issues with this now. The whole LootMgr needs to be reworked honestly, so this will likely just serve as a band-aid to stop the crashes.

PR #291, opened 2019-05-30
Resolve a spell stand issue (witnessed when eating)

BenDol:
## 🍰 Pullrequest
Resolves https://github.com/cmangos/issues/issues/1914
We want to initiate a stand no matter what the conditions of the spell (assuming successful).

### Proof
https://github.com/cmangos/issues/issues/1914

### Issues
https://github.com/cmangos/issues/issues/1914

### How2Test
1. login on two characters (one rogue)
2. stealth on the Rogue
2. /sit on each character (or eat food, same thing)
3. use Sap on the other character
4. each should be standing (eating will be interrupted).

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -5856,36 +5856,36 @@ void Unit::CasterHitTargetWithSpell(Unit* realCaster, Unit* target, SpellEntry c
         if (spellInfo->HasAttribute(SPELL_ATTR_EX3_NO_INITIAL_AGGRO) && !spellInfo->HasAttribute(SPELL_ATTR_EX3_OUT_OF_COMBAT_ATTACK))
             return;
 
+        // we want to change the stand state of each character if possible/required
+         // Since patch 1.5.0 sitting creature always stand up on attack (even if stunned)
+        if (success)
+        {
+            if (!spellInfo->HasAttribute(SPELL_ATTR_CASTABLE_WHILE_SITTING) && !realCaster->IsStandState())
+                realCaster->SetStandState(UNIT_STAND_STATE_STAND);
+
+            if (!target->IsStandState())
+                target->SetStandState(UNIT_STAND_STATE_STAND);
+        }
+
         // Hostile spell hits count as attack made against target (if detected), stealth removed at Spell::cast if spell break it
         const bool attack = (!IsPositiveSpell(spellInfo->Id, realCaster, target) && IsVisibleForOrDetect(target, target, false) && CanEnterCombat() && target->CanEnterCombat());
 
-        if (!spellInfo->HasAttribute(SPELL_ATTR_EX3_NO_INITIAL_AGGRO))
+        if (attack && !spellInfo->HasAttribute(SPELL_ATTR_EX3_NO_INITIAL_AGGRO) && !spellInfo->HasAttribute(SPELL_ATTR_EX_NO_THREAT))
         {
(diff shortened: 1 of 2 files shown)


berserkingyadis:
Tested and can confim that both chars are standing after the Sap.

src/game/Entities/Unit.cpp
```diff
@@ -5856,36 +5856,36 @@ void Unit::CasterHitTargetWithSpell(Unit* realCaster, Unit* target, SpellEntry c
         if (spellInfo->HasAttribute(SPELL_ATTR_EX3_NO_INITIAL_AGGRO) && !spellInfo->HasAttribute(SPELL_ATTR_EX3_OUT_OF_COMBAT_ATTACK))
             return;
 
+        // we want to change the stand state of each character if possible/required
+         // Since patch 1.5.0 sitting creature always stand up on attack (even if stunned)
+        if (success)
+        {
+            if (!spellInfo->HasAttribute(SPELL_ATTR_CASTABLE_WHILE_SITTING) && !realCaster->IsStandState())
+                realCaster->SetStandState(UNIT_STAND_STATE_STAND);
```
killerwife: I want to point out that in the old code we did not call this anywhere.

BenDol: yeah I know, I included it because the old code seemed like it had messed up thee logic. The SPELL_ATTR_CASTABLE_WHILE_SITTING was checked to see if it could make the target stand. Figured it was mean't for the caster given its not used anywhere else in the codebase.

killerwife: Actually, I have different evidence. In the classic beta, the spells without the attribute are not casted. You get a message "you need to be standing". Which leads me to believe this code is not intended.

BenDol: So rather than force standing, we want to perform a spell check for our standing state and reject it if this attribute doesn't exist?

killerwife: For caster, yes. For target keep the reworked code.

src/game/Entities/Unit.cpp
```diff
@@ -5856,36 +5856,31 @@ void Unit::CasterHitTargetWithSpell(Unit* realCaster, Unit* target, SpellEntry c
...
 
-        if (!spellInfo->HasAttribute(SPELL_ATTR_EX3_NO_INITIAL_AGGRO))
+        if (attack && !spellInfo->HasAttribute(SPELL_ATTR_EX3_NO_INITIAL_AGGRO) && !spellInfo->HasAttribute(SPELL_ATTR_EX_NO_THREAT))
         {
-            if (attack)
+            if (success)
             {
-                // Since patch 1.5.0 sitting characters always stand up on attack (even if stunned)
-                if (!spellInfo->HasAttribute(SPELL_ATTR_CASTABLE_WHILE_SITTING))
-                    if (success && !target->IsStandState() && target->GetTypeId() == TYPEID_PLAYER)
```
killerwife: Ok, I can change this myself when merging it, but you did not keep the attribute guard and the IsPlayer guard around the setting stand state. I think this can be dangerous to certain mobs which use stand state for animations. As a result we definitely need to keep it player only, AND the attribute is questionable. For the attribute I do not have evidence, I do not know who put it in the sources.

killerwife:
One last thing, please split the last commit into "refactor" and the actual change. When people are looking back at it, it will be easier to review.

PR #292, opened 2019-05-30
Convenience for VS property COMMAND_ARGUMENTS

BenDol:
## 🍰 Pullrequest
Sometimes VS debug fails to find the config files, this is a workaround.

### Issues
When you debug in Visual Studio there is an issue where sometimes the runner that launches the server doesnt allow us to load the conf files directly from the WorkDir, so this just explicitly specifies the conf files for debug purposes. Just a simple quality of life for anyone wanting to launch debug mode in VS.

### DIFF
--- a/src/mangosd/CMakeLists.txt
+++ b/src/mangosd/CMakeLists.txt
@@ -76,6 +76,7 @@ if(WIN32)
   set_target_properties(${EXECUTABLE_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY_DEBUG "${DEV_BIN_DIR}")
   set_target_properties(${EXECUTABLE_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY_RELEASE "${DEV_BIN_DIR}")
   set_target_properties(${EXECUTABLE_NAME} PROPERTIES VS_DEBUGGER_WORKING_DIRECTORY "$(OutDir)")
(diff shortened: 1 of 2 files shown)


killerwife:
This has never happened to me personally. I usually come, build solution with cmake and I can press F5 to debug. Hmmm, if you do a clean install this works out of the box?

BenDol:
It didn't happen to me initially, then just randomly started happening. I cleared my cmake directory and rebuilt it still happening. Not really too sure.

berserkingyadis:
Strange, after a fresh compile on a new system this is also happening to me.
image

These are my AuthServer and WorldServer settings
image
image

And this is my bin folder
image

I remember this definetly working on my other pc. Did I forget something?

killerwife:
I do not know, try to put printfing of current location of PATH before the error. Maybe it will give us a clue. Since I can't reproduce, I'd love to help in figuring out why it happens in the first place.

berserkingyadis:
Its working on one of my machines, but not on the other. I can try what you said this weekend when the other is available to me.

killerwife:
ok, @berserkingyadis pls do, we will deal with the PR after that

berserkingyadis:
I @cyberium 's commits fixed it for me.

@BenDol  can you try it on your end?

BenDol:
Yeah working for me too

PR #293, opened 2019-05-30
Fix stacking issues with cast item spells (i.e. Drums of War)

BenDol:
## 🍰 Pullrequest
Resolves https://github.com/cmangos/issues/issues/1943

Modified the RemoveNoStackAurasDueToAuraHolder check to also account for the cast items, if they ever match we need to avoid stacking when they are from another player. I also fixed some of the passive spell check logic to account for all the SPELL_EFFECT_APPLY_AURA_* spell effects. This will ensure friendly auras are always check.

### Proof
https://github.com/cmangos/issues/issues/1943

### Issues
https://github.com/cmangos/issues/issues/1943

### How2Test
* .learn all_crafts
* .setskill 165 350
* .additem 29528
* login to two character
* join party
* click drum
* .cooldown clear

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -4680,14 +4680,14 @@ bool Unit::AddSpellAuraHolder(SpellAuraHolder* holder)
     {
         if (!RemoveNoStackAurasDueToAuraHolder(holder))
         {
-            return false;                                   // couldn't remove conflicting aura with higher rank
+            return false; // couldn't remove conflicting aura with higher rank
         }
     }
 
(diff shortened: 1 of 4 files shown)


Warlockbugs:
Current state of stacking rewrite for Classic/TBC was introduced as a way to avoid extremely crude "same icon" rule with something more sensible and doing actual research stacking rules in the field on per-case basis. Currently it has depleted its capabilities as there is a layer of problematic cases related to partial effect multistacking, which requires another rewrite. It should be noted that this part of the core is currently being maintained only, while we gather more information as a basis for more complicated and granular system. Thus further major extensions are undesirable, every patch should be annotated one way or another to push research and further development further in the right direction.

PR #294, opened 2019-05-31
Resolve evading issues with units that have pets

BenDol:
## 🍰 Pullrequest
Resolves https://github.com/cmangos/issues/issues/1941

We need to make sure we are also evading the pets when calling EnterEvadeMode, otherwise the pet will keep attacking and reinvoke you as the target on the owner, causing strange behavior.

There were also other cases where stopping pets combat would also be appropriate that I have updated here too.

### Proof
https://github.com/cmangos/issues/issues/1941

### Issues
https://github.com/cmangos/issues/issues/1941

### How2Test
* Go to npc with pet (for example .go creature id 5617)
* Start combat.
* Try to go out of combat by run from them (try several times)

### Todo / Checklist

### DIFF
--- a/src/game/AI/BaseAI/GuardianAI.cpp
+++ b/src/game/AI/BaseAI/GuardianAI.cpp
@@ -84,7 +84,7 @@ void GuardianAI::CombatStop()
 void GuardianAI::EnterEvadeMode()
 {
     m_creature->RemoveAllAurasOnEvade();
-    m_creature->CombatStop(true);
+    m_creature->CombatStopWithPets(true);
 
     m_creature->TriggerEvadeEvents();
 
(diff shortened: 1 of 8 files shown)


killerwife:
Hmmm, shouldn't this be called in ANY combatstop? I mean, all entities who can be in combat can have pets, and this could potentially happen for any, not just the ones you've fixed. I would love to know what you think about that.

BenDol:
I thought so too, but I thought perhaps there are special edge cases where that wasn't the case, so I didn't move the code to the `StopCombat`. When I looked into it, it wasn't super obvious to me that it was the case. Except perhaps boss fights that have pets which remain in combat maybe? Not sure though.

cyberium:
Merged that way sound resonable and probably less harmfull. Thanks.

PR #295, opened 2019-06-01
Resolves issues with flag capping in BG's like AB

BenDol:
## 🍰 Pullrequest
Resolves https://github.com/cmangos/issues/issues/1250 

The solution here is that we need to send a player update for the objects state will sometimes not sync correctly with the client since the object is invisible when the state change takes place.
In order to maintain its object state on the clients we will keep track.

### Proof
https://github.com/cmangos/issues/issues/1250 

### Issues
https://github.com/cmangos/issues/issues/1250 

### How2Test
* login 2 chars of opposite factions
* .debug bg
* join Arathi Basin on each character
* cap the same flag post on each character
* do it forever because its fixed

### Todo / Checklist

### DIFF
--- a/src/game/BattleGround/BattleGround.h
+++ b/src/game/BattleGround/BattleGround.h
@@ -415,6 +415,9 @@ class BattleGround
         virtual void EndBattleGround(Team winner);
         static void BlockMovement(Player* plr);
 
+        void SendObjectUpdatesToPlayer(Player* player);
+        void AddPlayerUpdateObject(ObjectGuid guid);
+
         void SendMessageToAll(int32 entry, ChatMsg type, Player const* source = nullptr);
         void SendYellToAll(int32 entry, uint32 language, ObjectGuid guid);
         void PSendMessageToAll(int32 entry, ChatMsg type, Player const* source, ...);
@@ -616,6 +619,8 @@ class BattleGround
         float m_TeamStartLocZ[PVP_TEAM_COUNT];
         float m_TeamStartLocO[PVP_TEAM_COUNT];
         float m_startMaxDist;
+
+        GuidVector m_objUpdates;
 };
 
 // helper functions for world state list fill
--- a/src/game/BattleGround/BattleGround.h
+++ b/src/game/BattleGround/BattleGround.h
@@ -500,7 +500,7 @@ class BattleGround
 
         void HandleTriggerBuff(ObjectGuid go_guid);
 
-        void SpawnBGObject(ObjectGuid guid, uint32 respawntime);
+        void SpawnBGObject(ObjectGuid guid, uint32 respawntime, bool storeUpdate = false);
         void SpawnBGCreature(ObjectGuid guid, uint32 respawntime);
 
         void DoorOpen(ObjectGuid guid);
--- a/src/game/BattleGround/BattleGround.h
+++ b/src/game/BattleGround/BattleGround.h
@@ -620,7 +620,7 @@ class BattleGround
         float m_TeamStartLocO[PVP_TEAM_COUNT];
         float m_startMaxDist;
 
-        GuidVector m_objUpdates;
+        std::set<ObjectGuid> m_objUpdates;
 };
 
 // helper functions for world state list fill
--- a/src/game/BattleGround/BattleGround.cpp
+++ b/src/game/BattleGround/BattleGround.cpp
@@ -524,6 +524,31 @@ void BattleGround::SendPacketToTeam(Team teamId, WorldPacket const& packet, Play
     }
 }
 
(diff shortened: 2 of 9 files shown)


src/game/BattleGround/BattleGround.h
```diff
@@ -616,6 +619,8 @@ class BattleGround
         float m_TeamStartLocZ[PVP_TEAM_COUNT];
         float m_TeamStartLocO[PVP_TEAM_COUNT];
         float m_startMaxDist;
+
+        GuidVector m_objUpdates;
```
killerwife: Why not use std::set instead. You require uniqueness, so that solves it for you implicitly.

killerwife:
Damn, I would love to investigate this and look for it in a sniff, but I sure as hell do not have time for such a rabbit hole. So the issue is that we do not send some data, which causes the data to be stale in the client and then client thinks that a certain value is still set to the old one? The true fix to me would be to properly send that all on the object showing up again, but I digress. @Warlockbugs your opinion?

BenDol:
Yeah its a strange issue, if anything I can highlight it here if someone wants to find a more robust solution to the issue. I tried to dig in and see where it was going wrong with little success. But my time is spreading thin haha.

src/game/BattleGround/BattleGround.h
BenDol: Yeah good point will update that

killerwife:
@Shauren does this in any way ring a bell for anything in TrinityCore?

Shauren:
Yes, this is the same issue as .gobject move/turn commands had (had, but the fix for those is not applicable here)

Basically, client stores recently deleted (OutOfRange update) objects and brings them back using old state when new CreateObject is received, ignoring new state sent in that CreateObject packet

commands work around it by simply spawning a new object with new GUID

killerwife:
So the "true" fix is to have dynamic guids, and this is a bandaid until we have that like TC does, correct?

Shauren:
No, as I said at the beginning - the fix is not applicable to this BG flag case (guid is not supposed to change for the same object), treat what I said as simply explaining the cause

That client object cache behavior is not really well handled/reversed

BenDol:
Well the ball is rolling for this fix if someone wants to implement it correctly

Yatzii93:
I tried this out and for me it didn't resolve the issue....hmmmm

Lymphome:
Hi, don't know if it is the good place to post that (my first post on github) but following the comments from Kelno I modified SpawnBGObject() in BattleGround.cpp to solve this flag problem :

if (respawntime == 0)
{
        obj->Respawn(); // respawn => SetRespawnTime(0);
        if (obj->GetGoType() == GAMEOBJECT_TYPE_BUTTON)
                obj->SetInUse(false);   // GO_STATE_READY

        if (obj->GetLootState() == GO_JUST_DEACTIVATED)
                obj->SetLootState(GO_READY);

} else {
        // flags are buttons (except neutral flags) : need to force state update to clients
        if (obj->GetGoType() == GAMEOBJECT_TYPE_BUTTON)
        {
                obj->SetInUse(true);    // GO_STATE_ACTIVE
                obj->SendForcedObjectUpdate();
        }

        obj->SetLootState(GO_JUST_DEACTIVATED);
        obj->SetRespawnDelay(respawntime);
        obj->SetForcedDespawn();
}

I don't know if it is the good way but it works

killerwife:
Please, if you can, re-edit your code for it to be fully visible.

Yatzii93:
Interesting -
These are custom right ?
> obj->SetRespawnDelay(respawntime);
> obj->SetForcedDespawn();

Lymphome:
These parts come from Entities/[Creature|GameObject].cpp, there are already in. I add nothing else but this small part of code. I did a lots of tries to finally make it work. The order of the calls are important : you have to change states and respawntimes when the object is visible in the client (as Kelno notices it).

Lymphome:
AV quickly tested, it seems ok. Just have to change ally contested flag type for cim : put 'flags' field to 0 like other cim flags (DB Gob template id 179287). Tower flags are ok. May have other things to test...

killerwife:
Why would you need to change EY banner types?

Lymphome:
With type = 26, they are all displayed and don't despawn with this patch. With type 24 it works and this type seems more appropriate to the object (flag drop against flag stand)

Yatzii93:
@Lymphome Can you do a seperate pull request so I can see what you've done in an easy format to achieve the results?

cyberium:
Does not fix the issue. A proper way will be to give another guid to the object. Hopfully something is comming ;)

killerwife:
Soon™ 
image

PR #296, opened 2019-06-01
WSG flag worldstates & fix flag pickup bug

Yatzii93:
## 🍰 Pullrequest
-Improvement to the bug where the flag states switch on area change. https://github.com/cmangos/issues/issues/1872
-Stops a bug where flag pickup code is incorrectly triggered when both flag states have the status on the ground. (i.e Alliance player returns their flag but also gains horde flag.

### Proof
https://github.com/cmangos/issues/issues/1872

### Issues
-Faction/flag holder world states - https://github.com/cmangos/issues/issues/1872
-Flag pickup bug

### How2Test

World states:

- Pick up flag 
- Run into the middle of the map causing area change code to trigger, flag states will switch showing incorreclty which faction has which flag.

Flag pickup:

- Join WSG with 2 chars
- Pick up both flags and drop them at the same time (both flags on the ground)
- With 1 char return your faction flag, you'll gain the other factions flag in the process.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/BattleGround/BattleGroundWS.cpp
+++ b/src/game/BattleGround/BattleGroundWS.cpp
@@ -356,7 +356,7 @@ void BattleGroundWS::EventPlayerClickedOnFlag(Player* source, GameObject* target
         }
         else
         {
-			if(GetFlagState(HORDE) == BG_WS_FLAG_STATE_ON_GROUND)
+	        if(GetFlagState(HORDE) == BG_WS_FLAG_STATE_ON_GROUND)
                 return;
             message_id = LANG_BG_WS_PICKEDUP_AF;
             type = CHAT_MSG_BG_SYSTEM_HORDE;
--- a/src/game/BattleGround/BattleGroundWS.cpp
+++ b/src/game/BattleGround/BattleGroundWS.cpp
@@ -356,7 +356,7 @@ void BattleGroundWS::EventPlayerClickedOnFlag(Player* source, GameObject* target
         }
         else
         {
-	    if(GetFlagState(HORDE) == BG_WS_FLAG_STATE_ON_GROUND)
+	    if (GetFlagState(HORDE) == BG_WS_FLAG_STATE_ON_GROUND)
                 return;
             message_id = LANG_BG_WS_PICKEDUP_AF;
             type = CHAT_MSG_BG_SYSTEM_HORDE;
--- a/src/game/BattleGround/BattleGroundWS.cpp
+++ b/src/game/BattleGround/BattleGroundWS.cpp
@@ -356,7 +356,7 @@ void BattleGroundWS::EventPlayerClickedOnFlag(Player* source, GameObject* target
         }
         else
         {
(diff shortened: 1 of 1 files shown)


Yatzii93:
@killerwife

Yatzii93:
Seem to be missing pickup message and updates when flag is dropped and new member of the same faction picks it back up. Probably related to the world states hack fix.

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -356,6 +356,8 @@ void BattleGroundWS::EventPlayerClickedOnFlag(Player* source, GameObject* target
         }
         else
         {
+	    if(GetFlagState(HORDE) == BG_WS_FLAG_STATE_ON_GROUND)
```
killerwife: missing space between if and bracket

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -581,12 +585,12 @@ void BattleGroundWS::FillInitialWorldStates(WorldPacket& data, uint32& count)
...
     FillInitialWorldState(data, count, BG_WS_FLAG_CAPTURES_MAX, BG_WS_MAX_TEAM_SCORE);
 
-    if (m_FlagState[TEAM_INDEX_HORDE] == BG_WS_FLAG_STATE_ON_PLAYER)
+    if (m_FlagState[TEAM_INDEX_ALLIANCE] == BG_WS_FLAG_STATE_ON_PLAYER)
         FillInitialWorldState(data, count, BG_WS_FLAG_STATE_HORDE, 2);
     else
         FillInitialWorldState(data, count, BG_WS_FLAG_STATE_HORDE, 1);
-
-    if (m_FlagState[TEAM_INDEX_ALLIANCE] == BG_WS_FLAG_STATE_ON_PLAYER)
+
```
killerwife: this, whatever this is

BenDol:
I'm working on a proper fix for these issues, do you mind if I apply the change you made for the index switching there too? I can leave that here in this PR though 👍

src/game/BattleGround/BattleGroundWS.cpp
Yatzii93: oops sorry lol I'll correct

src/game/BattleGround/BattleGroundWS.cpp
Yatzii93: Does not having a space there stop that code from running. If so thats really odd because I havent noticed that.

AnonXS: codestyle

src/game/BattleGround/BattleGroundWS.cpp
Yatzii93: ye it's a werid one if I catch your drift, it seems the world states are getting mixed up somehow. I don't really know the inner workings on how the data bits are being applied, just a quick hack fix to get it working for the most part. Need to test BenDol's rework out though as anything is better than my work arounds :D

PR #297, opened 2019-06-02
Stop exploit involving the ability to respawn Hyjal mining nodes.

Yatzii93:
## 🍰 Pullrequest

-Stops mining nodes being continually respawned by player interaction.

A side note but related, the gems spawn for the whole raid after the first 2 bosses. As far as I am aware the mines should spawn in sections after defeating each faciton camp encounters. Might be something to add later.

### Proof

Don't really have a direct source to say a player couldn't respawn the nodes continually in retail, but it would be ridiculous if they could. Safe to assume that due to the potential to ruin the gem economy, this was not a thing back then, plus they vendor for a lot of gold.

### Issues

-Makes sure that a player can't keep triggering the retreat gossip on Jaina or Thrall (after completing each section bosses) to respawn the mining nodes.

### How2Test
- Go in Hyjal
- Kill first 2 bosses
- Do the retreat gossip on Jaina to spawn nodes.
- Mine a node
- Go back to Jaina and repeat the process.
- Easy repeatable gem collection with no restrictions.

### Todo / Checklist

- [X] None


killerwife:
As much as I hate to say this over and over, the way I nuked hyjal scripts and wrote them anew is gonna make his irrelevant. Hyjal will be out on next export round.

PR #298, opened 2019-06-02
Add a helper import script for the base database scripts

BenDol:
## 🍰 Pullrequest
Imports the base scripts automatically for you (with some basic configurations)

### How2Test
Run script.

### DIFF
--- a/sql/base/auto_import.sh
+++ b/sql/base/auto_import.sh
@@ -1,0 +1,25 @@
+USERNAME=""
+read -e -p "MySQL User: " -i "root" USERNAME
+PASSWORD=""
+read -s -p "MySQL Password: " PASSWORD
+export MYSQL_PWD="$PASSWORD"
+echo ""
+echo ""
+read -e -p "mangos database name?: " -i "tbcmangos" DB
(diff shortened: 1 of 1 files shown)


killerwife:
Suggestions:
Make it generate config on first use which you can fill up with your own data. Add the config to .gitignore. Same way as its done in all 3 DBs.

berserkingyadis:
If this is used at db import/creation, could the creation script in sql/create be included in this script?

BenDol:
@killerwife its just a once off import script when you need to reset your world data or w/e not sure we need to overdo it with configs personally. I use the prompts to stop the action when I use this script so I will likely leave it as is.

@berserkingyadis It could, I use it outside the creation of the actual create scripts. So prefer leaving it out or I could make it a configuration option.

berserkingyadis:
I was thinking this could maybe be the core side equivalent of the InstallfullDB.sh

cyberium:
@BenDol Please take in consideration @killerwife advise. The thing is separate data and script will permit to get script clean as we can add .config on gitignore list.
Also why not put that script directly in sql folder (not in base folder).

PR #300, opened 2019-06-02
Only mark auras as unstackable if the item types match

BenDol:
## 🍰 Pullrequest
Continuation of https://github.com/cmangos/mangos-tbc/pull/293

Needs more testing to confirm.

Modified the RemoveNoStackAurasDueToAuraHolder check to also account for the cast items, if they ever match we need to avoid stacking when they are from another player. I also fixed some of the passive spell check logic to account for all the SPELL_EFFECT_APPLY_AURA_* spell effects. This will ensure friendly auras are always check.

### Proof
https://github.com/cmangos/issues/issues/1943

### Issues
https://github.com/cmangos/issues/issues/1943

### How2Test
* .learn all_crafts
* .setskill 165 350
* .additem 29528
* login to two character
* join party
* click drum
* .cooldown clear

### DIFF
--- a/src/game/Spells/SpellMgr.cpp
+++ b/src/game/Spells/SpellMgr.cpp
@@ -1318,8 +1318,8 @@ bool SpellMgr::IsNoStackSpellDueToSpellAndCastItem(SpellAuraHolder const* spellH
         ObjectGuid castItem1 = spellHolder1->GetCastItemGuid();
         ObjectGuid castItem2 = spellHolder2->GetCastItemGuid();
         
-        // check that we have found an item, if we do then we are not stackable
-        stackable = !((castItem1 && !castItem1.IsEmpty()) || (castItem2 && !castItem2.IsEmpty()));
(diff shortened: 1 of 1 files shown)


BenDol:
@Warlockbugs we should get this into master for testing asap. I believe people have had issues with the previous changes.

InDebt:
I'm a bit confused. If you compare the Item GUIDs, how are they ever gonna be the same if we ensure !own in the first check. Each item, even if they have the same entry should have a different GUID so casters must trade their items mid-cast? 🤔 

Did you mean to check if they have an equal entry?

If so, i belive this to break spells with random effects like Darkmon Caard Madness or Mad Alchemist's Potion since they should all share the same cast item.

https://www.wowhead.com/item=34440/mad-alchemists-potion
https://www.wowhead.com/item=31859/darkmoon-card-madness

Furthermore this might even prevent Mongoose from double proccing.

BenDol:
Yeah that might be more appropriate actually

PR #301, opened 2019-06-02
Rework Warsong Gulch Battleground

BenDol:
## 🍰 Pullrequest
Reworked Warsong Gulch to be more efficient and resolve the strange sync issues.

Thanks to @Yatzii93 for testing and also providing the fix for the FillInitialWorldStates issue.

Deals with:
* The huge amount of code duplication
* Uses team index for all common data configs (FlagIds, MessageIds, MessageTypes, Sounds)
* Centralize flag actions to `PickUpFlagFromBase` & `GroundFlagInteraction`
* Properly checks the flags display id for the ground pick up actions

### Proof
Will compile a video of all the results, but it seems to be working flawlessly form all my testing.

### Issues
https://github.com/cmangos/issues/issues/1872
also addresses https://github.com/cmangos/mangos-tbc/pull/296

### How2Test
.debug bg
enter warsong

### Todo / Checklist
- [X] Internal Tests
- [x] Production Tests

### DIFF
--- a/src/game/BattleGround/BattleGround.cpp
+++ b/src/game/BattleGround/BattleGround.cpp
@@ -583,6 +583,9 @@ void BattleGround::CastSpellOnTeam(uint32 SpellID, Team teamId)
 
 void BattleGround::RewardHonorToTeam(uint32 Honor, Team teamId)
 {
+    if (teamId == TEAM_NONE)
+        return;
+
     for (BattleGroundPlayerMap::const_iterator itr = m_Players.begin(); itr != m_Players.end(); ++itr)
     {
         if (itr->second.OfflineRemoveTime)
--- a/src/game/BattleGround/BattleGroundWS.cpp
+++ b/src/game/BattleGround/BattleGroundWS.cpp
@@ -222,10 +222,10 @@ void BattleGroundWS::RespawnDroppedFlag(Team team)
     PvpTeamIndex teamIdx = GetTeamIndexByTeamId(team);
     RespawnFlag(team, false);
 
-    SendMessageToAll(wsMessageIds[teamIdx][BG_WS_FLAG_ACTION_CAPTURED],
-        wsChatMessageTypes[teamIdx][BG_WS_FLAG_ACTION_CAPTURED]);
+    SendMessageToAll(wsMessageIds[teamIdx][BG_WS_FLAG_ACTION_RESPAWN],
+        wsChatMessageTypes[teamIdx][BG_WS_FLAG_ACTION_RESPAWN]);
 
-    PlaySoundToAll(wsSounds[teamIdx][BG_WS_FLAG_ACTION_CAPTURED]);
+    PlaySoundToAll(wsSounds[teamIdx][BG_WS_FLAG_ACTION_RESPAWN]);
 
     GameObject* obj = GetBgMap()->GetGameObject(GetDroppedFlagGuid(team));
     if (obj)
@@ -325,7 +325,7 @@ void BattleGroundWS::EventPlayerDroppedFlag(Player* player)
         UpdateFlagState(team, 1);
 
         SendMessageToAll(wsMessageIds[otherTeamIdx][BG_WS_FLAG_ACTION_DROPPED],
-                         wsChatMessageTypes[TEAM_INDEX_HORDE][BG_WS_FLAG_ACTION_DROPPED], player);
+                         wsChatMessageTypes[otherTeamIdx][BG_WS_FLAG_ACTION_DROPPED], player);
         UpdateWorldState(wsStateUpdateId[otherTeamIdx], uint32(-1));
         m_FlagsDropTimer[otherTeamIdx] = BG_WS_FLAG_DROP_TIME;
     }
--- a/src/game/BattleGround/BattleGroundWS.cpp
+++ b/src/game/BattleGround/BattleGroundWS.cpp
@@ -243,7 +243,8 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* player)
 
     player->RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_ENTER_PVP_COMBAT);
 
-    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(player->GetTeam());
+    Team team = player->GetTeam();
+    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(team);
     uint8 otherTeamIdx = GetOtherTeamIndex(teamIdx);
 
     if (!IsFlagPickedUp(otherTeamIdx))
@@ -260,10 +261,10 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* player)
         m_TeamScores[teamIdx] += 1;
     
     PlaySoundToAll(wsSounds[teamIdx][BG_WS_FLAG_ACTION_CAPTURED]);
-    RewardReputationToTeam(890, m_ReputationCapture, player->GetTeam());
+    RewardReputationToTeam(890, m_ReputationCapture, team);
 
     // for flag capture is reward 2 honorable kills
-    RewardHonorToTeam(GetBonusHonorFromKill(2), player->GetTeam());
+    RewardHonorToTeam(GetBonusHonorFromKill(2), team);
 
     // despawn flags
     SpawnEvent(WS_EVENT_FLAG_A, 0, false);
@@ -272,8 +273,8 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* player)
     SendMessageToAll(wsMessageIds[otherTeamIdx][BG_WS_FLAG_ACTION_CAPTURED],
                      wsChatMessageTypes[teamIdx][BG_WS_FLAG_ACTION_CAPTURED], player);
 
-    UpdateFlagState(player->GetTeam(), 1);                  // flag state none
-    UpdateTeamScore(player->GetTeam());
+    UpdateFlagState(team, 1);                  // flag state none
+    UpdateTeamScore(team);
     // only flag capture should be updated
     UpdatePlayerScore(player, SCORE_FLAG_CAPTURES, 1);      // +1 flag captures
 
@@ -298,7 +299,8 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* player)
 
 void BattleGroundWS::EventPlayerDroppedFlag(Player* player)
 {
-    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(player->GetTeam());
+    Team team = player->GetTeam();
+    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(team);
     PvpTeamIndex otherTeamIdx = GetOtherTeamIndex(teamIdx);
 
     if (GetStatus() != STATUS_IN_PROGRESS)
@@ -320,7 +322,7 @@ void BattleGroundWS::EventPlayerDroppedFlag(Player* player)
         }
 
         player->CastSpell(player, SPELL_RECENTLY_DROPPED_FLAG, TRIGGERED_OLD_TRIGGERED);
-        UpdateFlagState(player->GetTeam(), 1);
+        UpdateFlagState(team, 1);
 
         SendMessageToAll(wsMessageIds[otherTeamIdx][BG_WS_FLAG_ACTION_DROPPED],
                          wsChatMessageTypes[TEAM_INDEX_HORDE][BG_WS_FLAG_ACTION_DROPPED], player);
@@ -331,7 +333,8 @@ void BattleGroundWS::EventPlayerDroppedFlag(Player* player)
 
 void BattleGroundWS::PickUpFlagFromBase(Player* player)
 {
-    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(player->GetTeam());
+    Team team = player->GetTeam();
+    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(team);
     PvpTeamIndex otherTeamIdx = GetOtherTeamIndex(teamIdx);
 
     PlaySoundToAll(wsSounds[otherTeamIdx][BG_WS_FLAG_ACTION_PICKEDUP]);
@@ -340,7 +343,7 @@ void BattleGroundWS::PickUpFlagFromBase(Player* player)
     m_FlagState[otherTeamIdx] = BG_WS_FLAG_STATE_ON_PLAYER;
 
     // update world state to show correct flag carrier
-    UpdateFlagState(player->GetTeam(), BG_WS_FLAG_STATE_ON_PLAYER);
+    UpdateFlagState(team, BG_WS_FLAG_STATE_ON_PLAYER);
     UpdateWorldState(wsStateUpdateId[otherTeamIdx], 1);
 
     player->CastSpell(player, wsSpellTypes[otherTeamIdx][BG_WS_FLAG_ACTION_PICKEDUP], TRIGGERED_OLD_TRIGGERED);
@@ -351,7 +354,8 @@ void BattleGroundWS::PickUpFlagFromBase(Player* player)
 
 uint32 BattleGroundWS::GroundFlagInteraction(Player* player, GameObject* target)
 {
-    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(player->GetTeam());
+    Team team = player->GetTeam();
+    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(team);
     PvpTeamIndex otherTeamIdx = GetOtherTeamIndex(teamIdx);
 
     int32 actionId = -1;
@@ -361,7 +365,6 @@ uint32 BattleGroundWS::GroundFlagInteraction(Player* player, GameObject* target)
     if (wsFlagIds[teamIdx] == displayId)
     {
         actionId = BG_WS_FLAG_ACTION_RETURNED;
-        Team team = player->GetTeam();
         UpdateFlagState(GetOtherTeam(team), BG_WS_FLAG_STATE_WAIT_RESPAWN);
         RespawnFlag(team, false);
         UpdatePlayerScore(player, SCORE_FLAG_RETURNS, 1);
@@ -374,9 +377,9 @@ uint32 BattleGroundWS::GroundFlagInteraction(Player* player, GameObject* target)
         actionId = BG_WS_FLAG_ACTION_PICKEDUP;
         SpawnEvent(otherTeamIdx, 0, false);
         SetFlagCarrier(otherTeamIdx, player->GetObjectGuid());
-        player->CastSpell(player, wsSpellTypes[otherTeamIdx][BG_WS_FLAG_ACTION_PICKEDUP], TRIGGERED_OLD_TRIGGERED);
+        player->CastSpell(player, wsSpellTypes[otherTeamIdx][actionId], TRIGGERED_OLD_TRIGGERED);
         m_FlagState[otherTeamIdx] = BG_WS_FLAG_STATE_ON_PLAYER;
-        UpdateFlagState(player->GetTeam(), BG_WS_FLAG_STATE_ON_PLAYER);
+        UpdateFlagState(team, BG_WS_FLAG_STATE_ON_PLAYER);
         UpdateWorldState(wsStateUpdateId[otherTeamIdx], 1);
         SendMessageToAll(wsMessageIds[otherTeamIdx][actionId], wsChatMessageTypes[teamIdx][actionId], player);
         PlaySoundToAll(wsSounds[otherTeamIdx][actionId]);
--- a/src/game/BattleGround/BattleGroundWS.cpp
+++ b/src/game/BattleGround/BattleGroundWS.cpp
@@ -358,7 +358,7 @@ uint32 BattleGroundWS::GroundFlagInteraction(Player* player, GameObject* target)
     PvpTeamIndex teamIdx = GetTeamIndexByTeamId(team);
     PvpTeamIndex otherTeamIdx = GetOtherTeamIndex(teamIdx);
 
-    int32 actionId = -1;
+    int32 actionId = BG_WS_FLAG_ACTION_NONE;
     uint32 displayId = target->GetDisplayId();
 
     // check if we are returning our flag
@@ -385,7 +385,7 @@ uint32 BattleGroundWS::GroundFlagInteraction(Player* player, GameObject* target)
         PlaySoundToAll(wsSounds[otherTeamIdx][actionId]);
     }
 
-    if (actionId > -1)
+    if (actionId != BG_WS_FLAG_ACTION_NONE)
         player->RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_ENTER_PVP_COMBAT);
     return actionId;
 }
@@ -404,7 +404,7 @@ void BattleGroundWS::EventPlayerClickedOnFlag(Player* player, GameObject* target
         PickUpFlagFromBase(player);
     }
     // Check if we are trying to pick up or return a flag from the ground
-    else if (player->IsWithinDistInMap(target, 10) && !GroundFlagInteraction(player, target))
+    else if (player->IsWithinDistInMap(target, 10) && GroundFlagInteraction(player, target) == BG_WS_FLAG_ACTION_NONE)
     {
         sLog.outError("Failed to action the WS flag from event '%d'.", event);
     }
--- a/src/game/BattleGround/BattleGroundWS.cpp
+++ b/src/game/BattleGround/BattleGroundWS.cpp
@@ -26,6 +26,94 @@
 #include "WorldPacket.h"
 #include "Tools/Language.h"
 
+static const uint32 wsFlagIds[PVP_TEAM_COUNT] = {
+    5912, // TEAM_INDEX_ALLIANCE
+    5913  // TEAM_INDEX_HORDE
+};
+
+static const uint32 wsMessageIds[PVP_TEAM_COUNT][WS_FLAG_ACTIONS_TOTAL] = {
+    // TEAM_INDEX_ALLIANCE
+    {
+        LANG_BG_WS_PICKEDUP_AF,            // BG_WS_FLAG_ACTION_PICKEDUP
+        LANG_BG_WS_RETURNED_AF,            // BG_WS_FLAG_ACTION_RETURNED
+        LANG_BG_WS_DROPPED_AF,             // BG_WS_FLAG_ACTION_DROPPED
+        LANG_BG_WS_CAPTURED_AF,            // BG_WS_FLAG_ACTION_CAPTURED
+        LANG_BG_WS_ALLIANCE_FLAG_RESPAWNED // BG_WS_FLAG_ACTION_RESPAWN
+    },
+    // TEAM_INDEX_HORDE
+    {
+        LANG_BG_WS_PICKEDUP_HF,            // BG_WS_FLAG_ACTION_PICKEDUP
+        LANG_BG_WS_RETURNED_HF,            // BG_WS_FLAG_ACTION_RETURNED
+        LANG_BG_WS_DROPPED_HF,             // BG_WS_FLAG_ACTION_DROPPED
+        LANG_BG_WS_CAPTURED_HF,            // BG_WS_FLAG_ACTION_CAPTURED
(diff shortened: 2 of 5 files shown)


Yatzii93:
Will test very soon and report back.

killerwife:
Okay, let me say this outright. I will focus on your other PRs, and when you are done here, we will talk design.

BenDol:
There issues with the design I've used here?

killerwife:
Nope, or at least not as of right now. But until you are done, I'd rather focus on other things. Please do provide and extensive list of issues it addresses when you are done.

BenDol:
Ah okay I see, sure thing, I'm done now; extensively tested as much as I could by myself.

It mainly addresses the issues I posted but I just couldn't stand how poorly written it was with all the code duplication etc. So I just rewrote it to be a lot cleaner.

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -553,10 +474,12 @@ WorldSafeLocsEntry const* BattleGroundWS::GetClosestGraveYard(Player* player)
     {
         if (GetStatus() == STATUS_IN_PROGRESS)
             return sWorldSafeLocsStore.LookupEntry(WS_GRAVEYARD_MAIN_ALLIANCE);
+
         return sWorldSafeLocsStore.LookupEntry(WS_GRAVEYARD_FLAGROOM_ALLIANCE);
     }
-    if (GetStatus() == STATUS_IN_PROGRESS)
+    else if (GetStatus() == STATUS_IN_PROGRESS)
```
domgho: Redundant else, please don't undo my work: https://github.com/cmangos/mangos-tbc/commit/ea4256273ba5b82e85fd88d9ba155d75b98ea5fb#diff-9f4e0b9638611b3f7a54b42f2c8fa91e

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -50,7 +50,7 @@ void BattleGroundWS::Update(uint32 diff)
                 RespawnFlag(ALLIANCE, true);
             }
         }
-        if (m_FlagState[TEAM_INDEX_ALLIANCE] == BG_WS_FLAG_STATE_ON_GROUND)
+        else if (m_FlagState[TEAM_INDEX_ALLIANCE] == BG_WS_FLAG_STATE_ON_GROUND)
```
domgho: Good else

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -133,13 +131,13 @@ void BattleGroundWS::RespawnDroppedFlag(Team team)
...
+    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(team);
     RespawnFlag(team, false);
-    if (team == ALLIANCE)
-        SendMessageToAll(LANG_BG_WS_ALLIANCE_FLAG_RESPAWNED, CHAT_MSG_BG_SYSTEM_NEUTRAL);
-    else
-        SendMessageToAll(LANG_BG_WS_HORDE_FLAG_RESPAWNED, CHAT_MSG_BG_SYSTEM_NEUTRAL);
 
-    PlaySoundToAll(BG_WS_SOUND_FLAGS_RESPAWNED);
+    SendMessageToAll(wsMessageIds[teamIdx][BG_WS_FLAG_ACTION_CAPTURED],
+        wsChatMessageTypes[teamIdx][BG_WS_FLAG_ACTION_CAPTURED]);
```
domgho: I’ll take this as an example for all your similar “optimizations”.. all you are doing is adding array lookups since the branch (maybe even two branches) still exists in GetTeamIndexByTeamId. Yes there is less code duplication but at the price of performance and i don’t think BG code needs to be touched often if it’s working

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -501,10 +424,8 @@ void BattleGroundWS::Reset()
 void BattleGroundWS::EndBattleGround(Team winner)
 {
     // win reward
-    if (winner == ALLIANCE)
-        RewardHonorToTeam(GetBonusHonorFromKill(m_HonorWinKills), ALLIANCE);
-    if (winner == HORDE)
-        RewardHonorToTeam(GetBonusHonorFromKill(m_HonorWinKills), HORDE);
+    RewardHonorToTeam(GetBonusHonorFromKill(m_HonorWinKills), winner);
```
domgho: What if there is no winner? (45m time limit)

domgho:
Since I was the last one who cleaned up WS code I decided to review it for you. Personally I would not take the design changes and code duplication reduction because of inefficiency and I would ask you to only add the fixes that you apparently unearthed but I left the project some time ago to pursue new things. Hope it helps.

BenDol:
Nothing personal dude, was just not the most efficiently written class, all the logic and design was spot on (minus a few bugs). If you want to strain progress because you're unable to accept constructive criticism I you aren't the right person to be reviewing this PR. No harm was intended.

src/game/BattleGround/BattleGroundWS.cpp
BenDol: Good point, thanks.

domgho:
I wanted to help you see it from a different perspective and gave you an example of why I think it’s slower and agreed that it reduces code duplication which I however personally value less than performance. You of course don’t need to take my feedback into account. No offense taken or intended

BenDol:
My only goal here is to help out this project. I don't care if this PR is merged or not. If its deemed to be too much of a performance hit for the sake of cleaner code, then I'll accept that from one of the core dev team. The reason I rewrote it in the first place was because debugging the issues was hell since everything was so fragmented and the code was duping all over the place.

So agree to disagree I guess.

src/game/BattleGround/BattleGroundWS.cpp
Yatzii93: Well to be fair, there was game breaking problems that needed to be addressed.

src/game/BattleGround/BattleGroundWS.h
```diff
@@ -101,9 +111,92 @@ enum BG_WS_Events
     WS_EVENT_SPIRITGUIDES_SPAWN   = 2
 };
 
+static const uint32 wsFlagIds[PVP_TEAM_COUNT] = {
```
evil-at-wow: It looks like none of the statics in this new block are used in the header, so please move all of this to the source file (for example to the start of the source file, in an anonymous namespace). I know this project isn't always the best example, but it's generally good practice to keep header files clean and minimal, so everything that's not part of "the interface" or "the API" and that can reasonably be kept out of headers, should probably not be in the header.

src/game/BattleGround/BattleGroundWS.cpp
evil-at-wow: No one said we can't touch the code if there are issues (game breaking or otherwise). I think he added "if it's working" as a condition for a reason :smile:

src/game/BattleGround/BattleGroundWS.h
```diff
@@ -115,21 +208,21 @@ class BattleGroundWS : public BattleGround
...
-        bool IsAllianceFlagPickedUp() const { return !m_flagCarrierAlliance.IsEmpty(); }
-        bool IsHordeFlagPickedUp() const { return !m_flagCarrierHorde.IsEmpty(); }
+        bool IsFlagPickedUp(uint8 teamIdx) const { return !m_FlagCarrier[teamIdx].IsEmpty(); }
+        bool IsAllianceFlagPickedUp() const { return !m_FlagCarrier[TEAM_INDEX_ALLIANCE].IsEmpty(); }
+        bool IsHordeFlagPickedUp() const { return !m_FlagCarrier[TEAM_INDEX_HORDE].IsEmpty(); }
 
         void RespawnFlag(Team team, bool captured);
         void RespawnDroppedFlag(Team team);
         uint8 GetFlagState(Team team) { return m_FlagState[GetTeamIndexByTeamId(team)]; }
+        uint32 GetWorldStateUpdateId(uint8 teamIdx);
```
evil-at-wow: This is not modifiying state, so please make it a const function.

src/game/BattleGround/BattleGroundWS.h
```diff
@@ -143,6 +236,10 @@ class BattleGroundWS : public BattleGround
         void EndBattleGround(Team winner) override;
         virtual WorldSafeLocsEntry const* GetClosestGraveYard(Player* player) override;
 
+        // Flag interactions
+        void PickUpFlagFromBase(Player* source, GameObject* target);
```
evil-at-wow: Unless I missed something, the GameObject parameter isn't used in the implementation. So that parameter can be removed.

Also, you *always* need a valid Player object to be passed to this function, because you're unconditionally dereferencing the parameter in the implementation. That's fine, but that fact is not clear with this API \(*\). A suggestion would be to make it a reference instead, because that makes the intention "a valid object is required" (aka "nullptr is not a valid value") clear even if you only look at the header.

\(*\) I admit that API in the project is often doing it like this though. We should change/improve that IMHO.

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,221 +205,156 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
...
 
-    if (!message_id)
+void BattleGroundWS::EventPlayerClickedOnFlag(Player* player, GameObject* target)
+{
+    if (GetStatus() != STATUS_IN_PROGRESS)
         return;
 
-    SendMessageToAll(message_id, type, source);
-    source->RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_ENTER_PVP_COMBAT);
+    uint8 event = (sBattleGroundMgr.GetGameObjectEventIndex(target->GetGUIDLow())).event1;
```
evil-at-wow: I think you can remove the outer pair of () here.

src/game/BattleGround/BattleGroundWS.cpp
evil-at-wow: I second that.

src/game/BattleGround/BattleGroundWS.h
```diff
@@ -20,11 +20,19 @@
 #define __BATTLEGROUNDWS_H
 
 #include "BattleGround.h"
+#include "Tools/Language.h"
```
evil-at-wow: This can be moved to the source as well (see remark below about the block of statics). Fewer dependencies in headers is always good :wink:

src/game/BattleGround/BattleGroundWS.cpp
BenDol: Which it wasn't & even this rework might have more issues. The point is that cleaner code makes for easier debugging, that was my whole reason to rework this class.

src/game/BattleGround/BattleGroundWS.cpp
BenDol: strange coding practices to remove an else  where even though logically it is still fine. The `else` from my experience also  makes it easier to read the code to understand the logical trail of the `if` statement. Removing them seems so pointless and deters away from the if statements logic trail. Doesn't make sense to remove them at all, unless I've missed something.

src/game/BattleGround/BattleGroundWS.cpp
Yatzii93: I probably took it to literally :P I have a habit of doing that.

src/game/BattleGround/BattleGround.cpp
```diff
@@ -583,6 +583,9 @@ void BattleGround::CastSpellOnTeam(uint32 SpellID, Team teamId)
 
 void BattleGround::RewardHonorToTeam(uint32 Honor, Team teamId)
 {
+    if (teamId == TEAM_NONE)
+        return;
```
domgho: Instead of adding this check in the one required place this now adds unnecessary checks for a team in cases where we already knew that the team is not none and makes it harder harder to reason about the code

src/game/BattleGround/BattleGroundWS.h
```diff
@@ -115,21 +124,21 @@ class BattleGroundWS : public BattleGround
...
-        void ClearHordeFlagCarrier() { m_flagCarrierHorde.Clear(); }
+        void ClearFlagCarrier(uint8 teamIdx) { m_FlagCarrier[teamIdx].Clear(); }
+        void ClearAllianceFlagCarrier() { m_FlagCarrier[TEAM_INDEX_ALLIANCE].Clear(); }
+        void ClearHordeFlagCarrier() { m_FlagCarrier[TEAM_INDEX_HORDE].Clear(); }
 
-        bool IsAllianceFlagPickedUp() const { return !m_flagCarrierAlliance.IsEmpty(); }
-        bool IsHordeFlagPickedUp() const { return !m_flagCarrierHorde.IsEmpty(); }
+        bool IsFlagPickedUp(uint8 teamIdx) const { return !m_FlagCarrier[teamIdx].IsEmpty(); }
+        bool IsAllianceFlagPickedUp() const { return !m_FlagCarrier[TEAM_INDEX_ALLIANCE].IsEmpty(); }
+        bool IsHordeFlagPickedUp() const { return !m_FlagCarrier[TEAM_INDEX_HORDE].IsEmpty(); }
```
domgho: The generic version that you added should be called instead of the team versions now

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,221 +288,156 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
         EndBattleGround(winner);
     }
     else
-    {
-        m_FlagsTimer[GetOtherTeamIndex(GetTeamIndexByTeamId(source->GetTeam()))] = BG_WS_FLAG_RESPAWN_TIME;
-    }
+        m_FlagsTimer[GetOtherTeamIndex(GetTeamIndexByTeamId(player->GetTeam()))] = BG_WS_FLAG_RESPAWN_TIME;
```
domgho: Use `otherTeamIdx`

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,221 +288,156 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
...
         }
         else
-            EventPlayerDroppedFlag(plr);
+            EventPlayerDroppedFlag(player);
     }
 }
 
+uint32 BattleGroundWS::GetWorldStateUpdateId(uint8 teamIdx) const
+{
+    return teamIdx == TEAM_INDEX_ALLIANCE ? BG_WS_FLAG_UNK_ALLIANCE : BG_WS_FLAG_UNK_HORDE;
```
domgho: This should be an array lookup too instead of a function of a branch

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,221 +288,156 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
...
-        PlaySoundToAll(BG_WS_SOUND_HORDE_FLAG_PICKED_UP);
-        SpawnEvent(WS_EVENT_FLAG_H, 0, false);
-        SetHordeFlagCarrier(source->GetObjectGuid());
-        m_FlagState[TEAM_INDEX_HORDE] = BG_WS_FLAG_STATE_ON_PLAYER;
-        // update world state to show correct flag carrier
-        UpdateFlagState(ALLIANCE, BG_WS_FLAG_STATE_ON_PLAYER);
-        UpdateWorldState(BG_WS_FLAG_UNK_HORDE, 1);
-        source->CastSpell(source, BG_WS_SPELL_WARSONG_FLAG, TRIGGERED_OLD_TRIGGERED);
+        UpdateFlagState(GetOtherTeam(player->GetTeam()), BG_WS_FLAG_STATE_WAIT_RESPAWN);
+        RespawnFlag(player->GetTeam(), false);
```
domgho: `player->GetTeam()` should be cached in variable if its used multiple times in a function

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,221 +288,156 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
...
 
-    int32 message_id = 0;
-    ChatMsg type = CHAT_MSG_BG_SYSTEM_NEUTRAL;
+uint32 BattleGroundWS::GroundFlagInteraction(Player* player, GameObject* target)
+{
+    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(player->GetTeam());
+    PvpTeamIndex otherTeamIdx = GetOtherTeamIndex(teamIdx);
 
-    uint8 event = (sBattleGroundMgr.GetGameObjectEventIndex(target_obj->GetGUIDLow())).event1;
+    int32 actionId = -1, messageId = 0, soundId = 0;
```
domgho: We use a single line per variable declaration: https://wiki.sei.cmu.edu/confluence/display/c/DCL04-C.+Do+not+declare+more+than+one+variable+per+declaration

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,221 +288,156 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
...
+        SetFlagCarrier(otherTeamIdx, player->GetObjectGuid());
+        player->CastSpell(player, wsSpellTypes[otherTeamIdx][actionId], TRIGGERED_OLD_TRIGGERED);
+        m_FlagState[otherTeamIdx] = BG_WS_FLAG_STATE_ON_PLAYER;
+        UpdateFlagState(player->GetTeam(), BG_WS_FLAG_STATE_ON_PLAYER);
+        UpdateWorldState(GetWorldStateUpdateId(otherTeamIdx), 1);
     }
 
-    // Horde flag on ground(not in base) (returned or picked up again)
-    if (GetFlagState(HORDE) == BG_WS_FLAG_STATE_ON_GROUND && source->IsWithinDistInMap(target_obj, 10))
+    if (actionId > -1)
```
domgho: You would get less lines and code complexity by simply calling those three functions in the places where you set the variables

src/game/BattleGround/BattleGroundWS.cpp
domgho: 1. Resharper and other code cleanup tools suggest it: https://www.jetbrains.com/help/resharper/RedundantIfElseBlock.html 2. It reduces nested code which you already noticed as you didn't add an else a few lines above/below either 3. It allows programmers to recognize this pattern quicker 4. It is currently consistent in the whole code base

BenDol: Seems to be semantics at this point, I'll change it back for the sake of point `4` which is a valid point.

src/game/BattleGround/BattleGround.cpp
BenDol: no team gets no RewardHonorToTeam, given this call is made from EndBattleGround and isn't very intensive it doesn't seem like a reasonable optimization.

domgho: Obviously the performance is negligible in all touched code but why intentionally write more inefficient and harder to understand code than my suggestion

domgho: Obviously the performance is negligible in all touched code but why intentionally write more inefficient and harder to understand code than my suggestion of moving it to the single needed place in EndBattleGround

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -150,56 +236,46 @@ void BattleGroundWS::RespawnDroppedFlag(Team team)
...
-        SendMessageToAll(LANG_BG_WS_CAPTURED_HF, CHAT_MSG_BG_SYSTEM_ALLIANCE, source);
-    else
-        SendMessageToAll(LANG_BG_WS_CAPTURED_AF, CHAT_MSG_BG_SYSTEM_HORDE, source);
+    SendMessageToAll(wsMessageIds[otherTeamIdx][BG_WS_FLAG_ACTION_CAPTURED],
+                     wsChatMessageTypes[teamIdx][BG_WS_FLAG_ACTION_CAPTURED], player);
 
-    UpdateFlagState(source->GetTeam(), 1);                  // flag state none
-    UpdateTeamScore(source->GetTeam());
+    UpdateFlagState(player->GetTeam(), 1);                  // flag state none
+    UpdateTeamScore(player->GetTeam());
```
domgho: GetTeam() should be cached at start of function

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,218 +293,143 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
...
-        PlaySoundToAll(BG_WS_SOUND_ALLIANCE_FLAG_PICKED_UP);
-        SpawnEvent(WS_EVENT_FLAG_A, 0, false);
-        SetAllianceFlagCarrier(source->GetObjectGuid());
-        m_FlagState[TEAM_INDEX_ALLIANCE] = BG_WS_FLAG_STATE_ON_PLAYER;
-        // update world state to show correct flag carrier
-        UpdateFlagState(HORDE, BG_WS_FLAG_STATE_ON_PLAYER);
-        UpdateWorldState(BG_WS_FLAG_UNK_ALLIANCE, 1);
-        source->CastSpell(source, BG_WS_SPELL_SILVERWING_FLAG, TRIGGERED_OLD_TRIGGERED);
+        actionId = BG_WS_FLAG_ACTION_RETURNED;
+        Team team = player->GetTeam();
```
domgho: GetTeam() should be cached at start of function and used in all cases instead of calling the function

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,218 +293,143 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
...
+        UpdateFlagState(player->GetTeam(), BG_WS_FLAG_STATE_ON_PLAYER);
+        UpdateWorldState(wsStateUpdateId[otherTeamIdx], 1);
+        SendMessageToAll(wsMessageIds[otherTeamIdx][actionId], wsChatMessageTypes[teamIdx][actionId], player);
+        PlaySoundToAll(wsSounds[otherTeamIdx][actionId]);
     }
 
-    // Alliance flag on ground(not in base) (returned or picked up again from ground!)
-    if (GetFlagState(ALLIANCE) == BG_WS_FLAG_STATE_ON_GROUND && source->IsWithinDistInMap(target_obj, 10))
+    if (actionId > -1)
+        player->RemoveAurasWithInterruptFlags(AURA_INTERRUPT_FLAG_ENTER_PVP_COMBAT);
```
domgho: You could still remove this branch by moving this into each section of the previous branch which is more efficient and easier to understand

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,218 +293,143 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
...
-            SendMessageToAll(LANG_BG_WS_DROPPED_HF, CHAT_MSG_BG_SYSTEM_HORDE, source);
-            UpdateWorldState(BG_WS_FLAG_UNK_HORDE, uint32(-1));
-        }
-        else
-        {
-            SendMessageToAll(LANG_BG_WS_DROPPED_AF, CHAT_MSG_BG_SYSTEM_ALLIANCE, source);
-            UpdateWorldState(BG_WS_FLAG_UNK_ALLIANCE, uint32(-1));
-        }
+        player->CastSpell(player, SPELL_RECENTLY_DROPPED_FLAG, TRIGGERED_OLD_TRIGGERED);
+        UpdateFlagState(player->GetTeam(), 1);
```
domgho: GetTeam() should be cached at start of function and used in all cases where GetTeam is called

src/game/BattleGround/BattleGroundWS.cpp
```diff
@@ -217,218 +293,143 @@ void BattleGroundWS::EventPlayerCapturedFlag(Player* source)
...
+    PvpTeamIndex teamIdx = GetTeamIndexByTeamId(player->GetTeam());
+    PvpTeamIndex otherTeamIdx = GetOtherTeamIndex(teamIdx);
+
+    PlaySoundToAll(wsSounds[otherTeamIdx][BG_WS_FLAG_ACTION_PICKEDUP]);
+    SpawnEvent(otherTeamIdx, 0, false);
+    SetFlagCarrier(otherTeamIdx, player->GetObjectGuid());
+    m_FlagState[otherTeamIdx] = BG_WS_FLAG_STATE_ON_PLAYER;
+
+    // update world state to show correct flag carrier
+    UpdateFlagState(player->GetTeam(), BG_WS_FLAG_STATE_ON_PLAYER);
```
domgho: GetTeam() should be cached at start of function and used in all cases where GetTeam is called

domgho:
I think the code is looking better than before and if you take my remaining and still unresolved suggestions into account I would revert my stance of not taking this cleanup since it will probably have no measurable effect on actual live performance. Thanks for your work on this project and your willingness to be open to changes.

src/game/BattleGround/BattleGround.cpp
BenDol: This is to make sure this call is future proof for other use cases and fits the design of the function which is `Reward Honor To Team` that implies a valid team which `TEAM_NONE` does not fit. This is not something that would be considered harder to understand.

src/game/BattleGround/BattleGroundWS.cpp
BenDol: I think since I'm returning what action was executed, I might as well leave this as it is.

berserkingyadis:
After testing the flag/capture/point interactions, these are my results:

# Display
- horde flag placement a bit off?
image
The alliance one appears to be centered well:
image
- top display of who has the flag: ok
- scoring points: ok
- bg happenings in chat, ok, with 1 thing not correct
The ingame output does not appear to be right. This text appears when the flag gets returned automatically
image
# Flag interactions
- getting flag from enemy base: ok
- dropping flag by rightclicking: ok
- dropping flag by death: ok
- picking flag back up: ok
- not picking up flag and it gets returned to base automatically: ok
- flags respawining after score: ok
- not being able to score because own flag is gone: ok
# Bg 
- win when 3 flags were captured: ok
- Scoreboard number of flag captures and flag returns are correct: ok

BenDol:
@killerwife do we have sniffs for the flag spawn positions?

AnonXS:
```
[0] GUID: Full: 0xF11002BE76000B34 Entry: 179830 UInt64: 17370386779814759220
[0] Object Type: GameObject (5)
[0] Update Flags: Unknown1, LowGuid, StationaryObject (88)
[0] Stationary Position: X: 1540.423 Y: 1481.325 Z: 351.8284, O: 3.089183

[3] GUID: Full: 0xF11002BE77000B35 Entry: 179831 UInt64: 17370386779831536437
[3] Object Type: GameObject (5)
[3] Update Flags: Unknown1, LowGuid, StationaryObject (88)
[3] Stationary Position: X: 916.5073 Y: 1433.826 Z: 346.3796, O: 0.01745246
```

BenDol:
@berserkingyadis fixed the positioning here https://github.com/cmangos/tbc-db/pull/655/files and also resolved the message issues in the latest commit. As for the spells dropping the flag, I have the feeling those spells aren't supposed to drop the flag, but could be wrong.

berserkingyadis:
These are the new Flag positions. Alliance flag looks good as ever and Horde flag still appears not to be in the middle of these skulls but can't argue with sniffs, I guess.
flags
Message issues are now fixed. The new strings are:
"The Horde flag is now placed at its base."
"The Alliance flag is now placed at its base."

Regarding the spells dropping the flag, can we get some BG pros in here? Alternatively just feedback how it should work with these spells (and other spells).

AnonXS:
The Spells you mentioned (SS/CoS/Blink) should not drop flag afaik.

src/game/BattleGround/BattleGroundWS.cpp
Lillecarl: > `player->GetTeam()` should be cached in variable if its used multiple times in a function. Same in other places

What's the benefit of this?

killerwife:
Closed by https://github.com/cmangos/mangos-tbc/commit/f4fd912e2f3253bd67625361224faa0fa31a1fc3
Thx for the rework.

henhouse:
Hey @BenDol, sorry to bump a closed PR. Was looking over this and saw that the debuff logic is apparently missing from cMaNGOS entirely. Did you intentionally leave it out in this rework, or were unaware of it by chance?

Spells are `46392` and `46393`.

See comment here for explanation on how they work: https://www.wowhead.com/spell=46393/brutal-assault#comments:id=290484

PR #302, opened 2019-06-03
Add Battleground config Raid Warning announcements 

BenDol:
## 🍰 Pullrequest
Adds the config Battleground.RaidWarningAnnouncer (default: 0)
Enable the use of raid warning message type for announcements (i.e starting messages)

Not sure if customization like this are acceptable with configurations, but this one seemed nice to have as an option.

### Proof
image

### How2Test
Start a Battleground

### DIFF
--- a/src/game/BattleGround/BattleGround.cpp
+++ b/src/game/BattleGround/BattleGround.cpp
@@ -32,7 +32,7 @@
 #include "WorldPacket.h"
 #include "Tools/Formulas.h"
 #include "Grids/GridNotifiersImpl.h"
-#include "Chat/Chat.h"
+#include "World/World.h"
 
 #include <cstdarg>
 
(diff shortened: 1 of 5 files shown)


killerwife:
May I ask how it works in oldschool TBC BGs? I do not mind a setting for this at all, but we need to have the correct default.

BenDol:
By default it should show in the battleground channel which just shows in the chat box (which is currently what it does).

cyberium:
Is there a demand for such a custom feature?

BenDol:
I submitted it because I wanted to use it in my server, figure people will want something like this?

Yatzii93:
> Is there a demand for such a custom feature?

Who wouldn't want "( . Y . )" in their battleground raid warning! High demand ;)

cyberium:
Custom change. Not bad but i dont see the need to add it to core base. Ping me there and if enough people want it we will see.

PR #304, opened 2019-06-05
eliminate warning about comparison of signed and unsigned types

berserkingyadis:
## 🍰 Pullrequest

This little change removes 3 warnings I had while compiling cmangos core.

### How2Test

1. Checkout this PR and compile
2. There should be no more warnings about type conflicts with comparisons

### TODO
- Do this for wotlk and classic if needed

@cala This uint32() casting was used a few lines before for similar comparisons, I wasnt sure why It wasnt done for the 3 places changed in this PR.

### DIFF
--- a/src/game/PlayerBot/Base/PlayerbotClassAI.cpp
+++ b/src/game/PlayerBot/Base/PlayerbotClassAI.cpp
@@ -342,7 +342,7 @@ Player* PlayerbotClassAI::GetHealTarget(JOB_TYPE type, bool onlyPickFromSameGrou
     // Try to find a tank in need of healing (if multiple, the lowest health one)
     while (true)
     {
-        if ((uCount + i) >= uint32(targets.size()) || !(targets.at(uCount).type & JOB_TANK)) break;
+        if (uint32(uCount + i) >= uint32(targets.size()) || !(targets.at(uCount).type & JOB_TANK)) break;
         uCount++;
     }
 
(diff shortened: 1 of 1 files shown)


cala:
@berserkingyadis thanks for this PR. The casting was probably lost by myself when I extended the functions to add new roles a few weeks ago.

evil-at-wow:
Out of curiosity, what compiler are you using?

I'm asking because a signed vs unsigned warning in this case doesn't make much sense to me. Both ```uCount``` and ```i``` are unsigned (uint8), so the result of the addition is unsigned as well, and the uint32 we're comparing with is obviously also unsigned... so that makes me wonder where the compiler sees a signed value. :confused:

cala:
@evil-at-wow I did not had warnings on my side either (gcc/clang here) but the casting was harmful and indeed consistent with the code above, so I merged this.

cala:
Strangely, Classic and WotLK versions of the exact same function already had these castings though I do PlayerBot porting from Classic to Wotlk... I'm puzzled on how they disappeared in TBC and reappeared in WotLK...

berserkingyadis:
I am using Visual Studio 2017 Community and the compiler seems to be MSVC 19.16.27027.1 .

evil-at-wow:
Your compiler is correct (they usually are because they know better than us developers). I was apparently not paying attention last night, because I got caught again by the C++ rules (despite explicitly typing uint8).
> Both ```uCount``` and ```i``` are unsigned (uint8), so the result of the addition is unsigned as well (...)

This is not true because basically arithmetic operators do not accept types smaller than int:
> If the operand passed to an arithmetic operator is integral or unscoped enumeration type, then before any other action (but after lvalue-to-rvalue conversion, if applicable), the operand undergoes integral promotion.

In this case, ```(uCount + i)``` will have type ```int```, which is signed, hence the warning when comparing it with an uint32.

I'm not sure why GCC/Clang don't produce a warning here.

PR #305, opened 2019-06-06
Hunter 3:2 Macro Support!

BenDol:
## 🍰 Pullrequest
All hunters can sleep in peace? Support for the Hunter 3:2 Macro.

Long shot that this will make it into `master` but at least provides a solution and also gets the ball rolling on the Hunter macro issues.

Give tips to improve this, it is somewhat hacky? But its not too bad overall given the nuance of this mechanic is very complicated in the current spell system.

### Proof
https://www.youtube.com/watch?v=cObqUH-zKwg
https://www.youtube.com/watch?v=lnoOnmW-0Bs

### Issues
- None

### How2Test
Add macro:
```
#showtooltip Steady Shot
/cast !Auto shot
/cast [target=pettarget, exists] Kill command
/cast Steady shot
/script UIErrorsFrame:Clear() 
```
Use it as demonstrated in the video.

### Todo / Checklist

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -588,6 +588,7 @@ Player::Player(WorldSession* session): Unit(), m_taxiTracker(*this), m_mover(thi
     m_ammoDPS = 0.0f;
 
     m_temporaryUnsummonedPetNumber = 0;
+    m_pendingSteadyShot = false;
 
     //////////////////// Rest System/////////////////////
     time_inn_enter = 0;
--- a/src/game/Entities/Player.h
+++ b/src/game/Entities/Player.h
@@ -1184,6 +1184,8 @@ class Player : public Unit
         void RemoveAmmo();
         float GetAmmoDPS() const { return m_ammoDPS; }
         bool CheckAmmoCompatibility(const ItemPrototype* ammo_proto) const;
+        void SetPendingSteadyShot(bool pendingSteadyShot) { m_pendingSteadyShot = pendingSteadyShot; }
+        float IsPendingSteadyShot() const { return m_pendingSteadyShot; }
         void QuickEquipItem(uint16 pos, Item* pItem);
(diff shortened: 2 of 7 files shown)


Yatzii93:
Tested and working great, aaaaaaaaaaaaaaaamazing!

Zarant:
I tested this change earlier and I noticed that the auto shot wind up time is not resetting properly. When you move or stop casting a spell the auto shot wind up timer should reset.

Zarant:
I'm gonna copy paste the research I did a while ago:

cyberium:
@killerwife check that implemention and use of "TRIGGERED_IGNORE_GCD" please.

Zarant:
Another thing that I should mention, testing auto shot mechanics on the classic beta client I found out that whenever the auto shot wind up timer is reset the client fires the "UNIT_SPELLCAST_FAILED" event for auto shot.

BenDol:
Hmm interesting, I actually don't think those issues you mentioned are related to this PR specifically. I think this needs to be investigated.

DramatimeWoW:
I agree BenDol. To speak a bit on what Zarant was talking about with the double steady w/ an auto inbetween this is what I am observing:
> Auto Shot started
> Steady Shot Started
> Steady Shot Fires
> Steady Shot started
> Auto Shot Timer finishes, but is delayed for Steady Shot
> Steady Shot cast finishes, but doesn't go off
> Delayed Auto Shot fires
> Without pressing anything further, a steady shot is started and finishes. This appears to happen when you start a steady in the later portion of the auto shot timer so that the auto will finish at the same time or before the steady shot finishes.

Zarant:
I found out earlier that the SetPendingSteadyShot(true) function is being called when steady shot is mid cast, causing the weird issues Dramatime described.

cyberium:
Too much specific case is used in this PR. There must be another way to do it. Maybe spellscript can be used now?
Anyway not acceptable as it is.

PR #307, opened 2019-06-07
Add RelWithDebInfo & MinSizeRel build types

BenDol:
## 🍰 Pullrequest
Add RelWithDebInfo & MinSizeRel build types to CMAKE build.

### How2Test
Run CMake


cyberium:
Merged Thank you.

CMakeLists.txt
```diff
@@ -318,6 +318,8 @@ configure_file(
 )
 
 set(DEFINITIONS_RELEASE NDEBUG)
+set(DEFINITIONS_RELWITHDEBINFO NDEBUG _RELWITHDEBINFO MANGOS_RELWITHDEBINFO)
+set(DEFINITIONS_MINSIZEREL NDEBUG _MINSIZEREL MANGOS_MINSIZEREL)
```
evil-at-wow: The NDEBUG here is useful, but is there any point in adding MANGOS_RELWITHDEBINFO, MANGOS_MINSIZEREL,  _RELWITHDEBINFO and _MINSIZEREL? These are not used in code - and hopefully will never be - so why would we want to pass those to the compiler?

cyberium: For me those might be used in perfect world to add/remove some code needed in those scenario. We will probably never use them but well we can not say this is not a good code.

evil-at-wow: Actually we can say it's not good code (well, partly). Names starting with an underscore (_) followed by an uppercase letter are **reserved** for the implementation (read: your compiler). So using _RELWITHDEBINFO and _MINSIZEREL in code is a very big no-go for us. 

I also don't see what code you might ever want to add or remove in these cases (what would you possibly not want in debug or release, yet want in release with debug symbols?), but OK, if you really want it, maybe we can just keep the MANGOS_ ones and only those then.

PR #308, opened 2019-06-08
Fixes auto shot wind up time issues

Zarant:
## 🍰 Pullrequest

Currently, auto shot has no wind up time, 
This pull request makes the auto shot wind up time reset whenever you are not able to fire an auto shot.

### Proof

If you pay close attention to the auto shot timer, you can see that auto shot has a 0.5 wind up time 
https://www.youtube.com/watch?v=FAz6E6W4T-I

This old wowwiki page also references the 0.5s wind up time: 
https://wowwiki.fandom.com/wiki/Auto_Shot?oldid=1153137

### Issues

- None

### How2Test

- Shoot something, move around long enough so the auto shot timer is fully reset and stop, you should only fire your next auto 0.5s later.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -4128,8 +4128,8 @@ void Unit::_UpdateAutoRepeatSpell()
         // cancel wand shoot
         if (m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_spellInfo->Category == 351)
             InterruptSpell(CURRENT_AUTOREPEAT_SPELL);
-        else if (getAttackTimer(RANGED_ATTACK) < 500)
-            setAttackTimer(RANGED_ATTACK, 500);
+        // set 0.5 second wind-up time.
+        m_AutoRepeatFirstCast = true;
         return;
     }
 
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -4128,6 +4128,8 @@ void Unit::_UpdateAutoRepeatSpell()
         // cancel wand shoot
         if (m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_spellInfo->Category == 351)
             InterruptSpell(CURRENT_AUTOREPEAT_SPELL);
+        else if (getAttackTimer(RANGED_ATTACK) < 500)
+			setAttackTimer(RANGED_ATTACK, 500);
         return;
     }
 
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -4129,7 +4129,7 @@ void Unit::_UpdateAutoRepeatSpell()
         if (m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_spellInfo->Category == 351)
             InterruptSpell(CURRENT_AUTOREPEAT_SPELL);
         else if (getAttackTimer(RANGED_ATTACK) < 500)
-			setAttackTimer(RANGED_ATTACK, 500);
+            setAttackTimer(RANGED_ATTACK, 500);
         return;
     }
 


DramatimeWoW:
I pulled this locally and tested it, fix works as described.

Phatcat:
You could essentially have the same effect by doing this instead, taking advantage of the fact that there's a 'check if less than 0.5 seconds' right below as well as a 'AutoRepeatFirstCast bool to take advantage of;
```
    // check "real time" interrupts
    if (IsMoving() || IsNonMeleeSpellCasted(false, false, true))
    {
        // cancel wand shoot
        if (m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_spellInfo->Category == 351)
            InterruptSpell(CURRENT_AUTOREPEAT_SPELL);
        // set 0.5 second wind-up time.
        m_AutoRepeatFirstCast = true;
        return;
    }

    // apply delay
    if (m_AutoRepeatFirstCast && getAttackTimer(RANGED_ATTACK) < 500)
        setAttackTimer(RANGED_ATTACK, 500);
    m_AutoRepeatFirstCast = false;
```

src/game/Entities/Unit.cpp
```diff
@@ -4128,6 +4128,8 @@ void Unit::_UpdateAutoRepeatSpell()
         // cancel wand shoot
         if (m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_spellInfo->Category == 351)
             InterruptSpell(CURRENT_AUTOREPEAT_SPELL);
+        else if (getAttackTimer(RANGED_ATTACK) < 500)
+			setAttackTimer(RANGED_ATTACK, 500);
```
Phatcat: This line here is also indented with tabs, this project uses 4 white-space characters in place of a tab, like you've correctly done on the line above it.

Phatcat:
In fact, to me, it looks like the 0.5s delay to hunter auto shot is already implemented in the core (look under the comment // apply delay), it's just not utilized currently since m_AutoRepeatFirstCast is not set to true anywhere at all... I'm pretty sure it used to (I've been poking around in this function before)

Zarant:
If you look closely, when updating the autorepeat spell state you first check if you're moving or casting a spell, if one of those conditions are true means that you can't cast an auto repeat spell.
The second block of code is not run if the above 2 conditions are met, it only executes on the very first time you turn on your auto repeat spell, meaning that once you turn your auto repeat spell on, only the very first attack will have a wind up time, every other attack will not.

Phatcat:
I don't have to 'look closely', I see exactly what's going on.
Unit::_UpdateAutoRepeatSpell() is called however often I don't know, but a bunch of times each second. Whenever it runs, it checks for moving and/or casting of another spell, and in case of wand users, they stop wanding, in case of hunters, nothing happens. On your pr, here you check if the time to auto shoot is less than 0.5 seconds, if it is, you set it to 0.5 seconds, then you return out of the function.

Next time the function runs, if the player is still moving or casting another spell, this will keep repeating until the player stands still and stops casting other things, at which point it will now by-pass the if statement as well as the 'apply delay' if statement stuff and continue towards shooting.

The difference I suggest is that instead of checking if it's 0.5 seconds in the moving and casting if statement, rather set it as 'first cast', in which case it will still return out of the function, and only when the player stops moving or stops shooting other things will it bypass both of the if statements and allow for the player to start shooting.

It is 100% the same outcome, just using existing checks and bools instead of basically introducing code duplication.

"The second block of code is not run if one of 2 conditions above are met, it only executes on the very first time you turn on your auto repeat spell, meaning that once you turn your auto repeat spell on, only the very first attack will have a wind up time, every other attack will not."

Wait a second... did you not see the;
m_AutoRepeatFirstCast = true; line ? 
did you not look at the suggestion at all or read what I wrote?

killerwife:
I will merge this eventually because it conflicts with my nuking of ranged attack timer and using CD system for it. It is valid, but it will need splicing together.

ulfgebhardt:
@killerwife i see no conflict here - what do you mean by
> it conflicts with my nuking of ranged attack timer and using CD system for it
?

evil-at-wow:
He means it's conflicting with the currently still hidden stash of changes he has (done in the context of a server). You've been around here for a while, so you might remember that like twice a year or so a ton of commits ("exports") are done here. My interpretation is that the changes here conflict with what's going to be in that batch of commits.

killerwife:
What  @evil-at-wow said.

PR #309, opened 2019-06-08
Resolves players running on the spot

Yatzii93:
## 🍰 Pullrequest

-It seemed the old code was applying data packets from the source to other players in close proximity. This was causing a lot of visual issues like players suddenly running on the spot.
-This fix also improves the de-flying/re-flying animation client side. Much reduce delay to update other flying player states.

Fixes - https://github.com/cmangos/issues/issues/1933

(FYI: I may be describing whats happening here wrong, this area of the software isn't something I'm well versed in)

### Proof
https://github.com/cmangos/issues/issues/1933
https://youtu.be/KIDYfksq5W4
This is a custom anti-cheat, but it exposed the data packet problem this PR fixes + also resolves the issue with the anti-cheat processing these unintended duplicate data packets. 

Just so theres no confusion, without the anti-cheat the player wouldn't be teleporting around, this is just the anti-cheat responding to fact that it's receiving the druids movement information from the non-druid. The anti-cheat thinks the non-druid is hacking and ultimlaty is trying to send the non-druid back to a previously stored waypoint.

### Issues

-players running on the spot
-unintended duplicate data packets being applied to players in close proximity 

### How2Test

-leave druid flight form and re-enter in close proximity with another player
(I needed a custom anti-cheat to fully expose the issue)

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -4277,20 +4277,10 @@ void Player::SetLevitate(bool enable)
 
 void Player::SetCanFly(bool enable)
 {
-    WorldPacket data;
-    if (enable)
-        data.Initialize(SMSG_MOVE_SET_CAN_FLY, 12);
-    else
-        data.Initialize(SMSG_MOVE_UNSET_CAN_FLY, 12);
-
+    WorldPacket data(enable ? SMSG_MOVE_SET_CAN_FLY : SMSG_MOVE_UNSET_CAN_FLY, GetPackGUID().size() + 4);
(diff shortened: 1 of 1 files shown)


Yatzii93:
Ok I've had this change in live my end for most of the day and there hasen't been an instance of a player running on the spot or spinning. No issues from the change have been brought up by the playercase, I've done extensive testing myself, it all seems good.

BenDol:
Nice catch @Yatzii93 !

PR #313, opened 2019-06-12
Resolve issues with spell Shadowform failure reports

BenDol:
## 🍰 Pullrequest
There are cases where passive abilities should still be reported or you get client sync issues.

There may be a valid argument to say Shadowform should simply not be passive, but I fear that will have side effects that aren't intended. Thoughts?

### Proof
https://github.com/cmangos/issues/issues/1879

### Issues
Resolves https://github.com/cmangos/issues/issues/1879

### How2Test
Spam Shadowform spell until it fails.

### Todo / Checklist


berserkingyadis:
Can confirm that Shadowform isnt blocked anymore after 5 minutes of button mashing.

PR #314, opened 2019-06-12
Update Formulas.h

Phatcat:
The player level * 5 is not being calculated in with the base which is expected with the new formula


killerwife:
I'd love to merge this but I do not understand the context at all. Please explain more thoroughly, so I don't have to wait for cyberium.

cyberium:
You can merge it if you are in hurry it was on my list.

PR #320, opened 2019-07-01
add optional output path to extractor scripts

esno:
## 🍰 Pullrequest
this PR introduces an optional output path used by the extractor scripts.
It should prevent the requirement of writing into the client directory which could be in fact readonly.

### Issues
- https://github.com/cmangos/issues/issues/1965
- https://github.com/cmangos/mangos-classic/pull/399

### How2Test
```
./ExtractResources.sh a <clientdir> <outputdir>
```

### DIFF
--- a/contrib/mmap/src/IntermediateValues.cpp
+++ b/contrib/mmap/src/IntermediateValues.cpp
@@ -267,7 +267,7 @@ namespace MMAP
         fwrite(&b, sizeof(char), 1, objFile);
         fclose(objFile);
 
-        sprintf(objFileName, "%s/meshes/%03u%02u%02u.mesh", mapID, tileY, tileX);
+        sprintf(objFileName, "%s/meshes/%03u%02u%02u.mesh", m_workdir, mapID, tileY, tileX);
         objFile = fopen(objFileName, "wb");
         if (!objFile)
         {
--- a/contrib/mmap/src/IntermediateValues.cpp
+++ b/contrib/mmap/src/IntermediateValues.cpp
@@ -29,7 +29,7 @@ namespace MMAP
         rcFreePolyMeshDetail(polyMeshDetail);
     }
 
-    void IntermediateValues::writeIV(uint32 mapID, uint32 tileX, uint32 tileY)
+    void IntermediateValues::writeIV(const char* workdir, uint32 mapID, uint32 tileX, uint32 tileY)
     {
         char fileName[255];
         char tileString[25];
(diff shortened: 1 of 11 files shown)


contrib/mmap/src/IntermediateValues.cpp
```diff
@@ -254,7 +254,7 @@ namespace MMAP
         fwrite(&b, sizeof(char), 1, objFile);
         fclose(objFile);
 
-        sprintf(objFileName, "meshes/%03u%02u%02u.mesh", mapID, tileY, tileX);
+        sprintf(objFileName, "%s/meshes/%03u%02u%02u.mesh", mapID, tileY, tileX);
```
evil-at-wow: You're lacking a string argument here, so this call is not going to end well.

contrib/mmap/src/generator.cpp
```diff
@@ -220,6 +232,14 @@ bool handleArgs(int argc, char** argv,
 
             offMeshInputPath = param;
         }
+        else if (strcmp(argv[i], "--workdir") == 0)
+        {
+            param = argv[++i];
```
evil-at-wow: You're not checking if there's an argument after i. If the command ends with ```... --workdir``` that's an out of bounds access.

contrib/mmap/src/IntermediateValues.cpp
esno: yeah - saw that too already fixed locally ;)

contrib/mmap/src/generator.cpp
esno: oh you're right - looks like that counts to all multivalue params. I'll fix that

esno: @evil-at-wow meanwhile I had a look once again into that and I assume the loop is ending one before end of `argv` array. that's why other arguments hadn't that before. I would keep the current state and rework that later if necessary are you agree with that?

evil-at-wow: I checked again, and assuming I've been looking at the correct commit, it looks like it's just skipping the first argument (which makes sense to skip when looking at options, because that's the program name):
```
for (int i = 1; i < argc; ++i)
...
```
So when ```i``` gets its maximum value (being one less than argc) then ```argv[i]``` (=```argv[argc-1]```) is the last element of argv (remember that in C/C++ array indexes start at 0) and ```argv[++i]]``` is one past that, so an out of bounds access.

So I think they all have it. Try running the program with a missing value for an option at the end (say ```... --tile``` or ```... --maxAngle```) and see what happens.

cyberium:
@esno when you get some time please merge or close this. (on other core too)

al3xc1985:
@esno do u have time to merge this?

cyberium:
rebased in https://github.com/cmangos/mangos-tbc/pull/487

PR #321, opened 2019-07-05
Core/Unit: Fix a bug Stop Combat In Instances.

keyshuwen:
## 🍰 Pullrequest

creature Stop Combat In Instances.

### Proof

- None

### Issues

- Avoid escaping combat.

### How2Test

- In Instances, the attack creature, character moves a distance to escape the battle.

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -8933,8 +8933,12 @@ bool Unit::SelectHostileTarget()
     {
         if (IsLeashingTarget(target))
         {
-            AI()->EnterEvadeMode();
-            return false;
+            const MapEntry* entry = sMapStore.LookupEntry(GetMapId());
+            if (entry->map_type != MAP_RAID && entry->map_type != MAP_INSTANCE)
+            {
+                AI()->EnterEvadeMode();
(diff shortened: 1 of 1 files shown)


killerwife:
Please put it inside IsLeashingTarget. That's what this method is supposed to emulate.

cala:
I agree with @killerwife this should be in `IsLeashingTarget()`. This is a good start but will only fix issue for NPCs in dungeons, not for outer world NPCs. This will also prevent any leashed NPC in instance maps to evade on leashing because a complete leash code is still missing. 

I've been evaluating a small fix on my side for the past weeks using leash field in `creature_template` to select which NPCs should ignore leashing and which ones are eligible to it. This allows to have the code working to whatever NPC I want. However, as long as proper leash code is missing, I'm uneasy with forcing the use of a DB field as a selector (though I have the gut this is a good approach).

We can put your code or mine as a temporary solution.

keyshuwen:
Thank you for your advice.

killerwife:
You don't need to close the PR. Force push into the PR branch.

keyshuwen:
emmm......Sorry

PR #324, opened 2019-07-07
Fix a bug Stop Combat In Instances.

keyshuwen:
## 🍰 Pullrequest

creature Stop Combat In Instances.
### Proof

- None

### Issues

- Avoid escaping combat.

### How2Test

- In Instances, the attack creature, character moves a distance to escape the battle.

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -9289,7 +9289,7 @@ bool Unit::IsLeashingTarget(Unit* victim) const
 
     // Use AttackDistance in distance check if threat radius is lower. This prevents creature bounce in and out of combat every update tick.
     // TODO: Implement proper leashing
-    return !victim->IsWithinDist3d(x, y, z, ThreatRadius > AttackDist ? ThreatRadius : AttackDist);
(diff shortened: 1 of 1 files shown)


BenDol:
Does `IsDungeon` also return true for raids? I assume its the same behavior.

keyshuwen:
yes, 

> bool IsDungeon() const { return map_type == MAP_INSTANCE || map_type == MAP_RAID; }

cala:
Hi, I will merge this PR as it is with the addition of a few comments because it is an improvement over the current state of the core where many dungeon encounters or event are simply broken.

However, this fix assume there is zero leash in dungeons which is false. Some indoor bosses reset if pulled too far from their original position (and some outdoor bosses do the opposite).

So we need a real leash management that can handle all these cases. In the meantime, we will use this fix.

cala:
And because I'm a dumbass, I did not even thank you for this PR.

Thanks ! 😉

cala:
Merged in https://github.com/cmangos/mangos-tbc/commit/6bc15c5acf8224a9da1245bc08c54ecb89a1335c
Thank you!

PR #330, opened 2019-07-28
Despawn Wrath-Scryer's Felfire npcs to avoid being stuck in combat

Yatzii93:
## 🍰 Pullrequest

Wrath-Scryer's Felfire npcs do not despawn on fail or completion of the boss. This results in players staying in combat with these invisible creatures.
### Proof

- None

### Issues

- Stop players getting stuck in combat with invisible spell trigger npcs

### How2Test

- Go into The Arcatraz and engage Wrath-Scryer Soccothrates.
- Let him do his charge spell then proceed to kill the boss.
- You are now stuck in combat.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/arcatraz/arcatraz.h
+++ b/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/arcatraz/arcatraz.h
@@ -111,6 +111,7 @@ class instance_arcatraz : public ScriptedInstance, private DialogueHelper
         uint8 m_uiKilledDefenders;
 
         GuidList m_lSkyrissEventMobsGuidList;
+        GuidList m_ChargeFelfireMobsGuidList;
 };
 
 #endif
--- a/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/arcatraz/instance_arcatraz.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/arcatraz/instance_arcatraz.cpp
@@ -159,6 +159,7 @@ void instance_arcatraz::OnCreatureCreate(Creature* pCreature)
             break;
         case NPC_WRATH_SCRYER_FELFIRE:
             m_npcEntryGuidCollection[pCreature->GetEntry()].push_back(pCreature->GetObjectGuid());
+            m_ChargetargetMobsGuidList.push_back(pCreature->GetObjectGuid());
             break;
         case NPC_BLAZING_TRICKSTER:
         case NPC_PHASE_HUNTER:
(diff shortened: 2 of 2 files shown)


Yatzii93:
I changed the respawn of this npc in the db to basically nothing considering they need to be present before a group starts the encounter. This could be handled in the core though.

AnonXS:
Baseline strange thing is that triggers normally never enter combat and we do not have reports about combat bugs on this encounter so something is missing in cmangos-tbc.

Yatzii93:
Hmmm is there a unitflag or extra flag that stops a trigger getting an aggro table? Seems to be a common thing I have to account for my end where if a trigger does a spell and a player gets hit by said spell. The trigger is put in combat with the player.

AnonXS:
20978	Wrath-Scryer's Felfire	UnitFlags 33554432

Yatzii93:
I wonder if this is happening in the public core or if its just an issue my end.

AnonXS:
Also happens on my end, not sure how this wasnt reported. but i guess the real fix is in spell_template.

AnonXS:
Need to be despawned as they enter combat
```
NumObjUpdates: 2
MapID: 552 (552)
HasDestroyObjects: False
Data size: 183
[0] UpdateType: Values
[0] Object Guid: Full: 0x2016F44500147C8000787700004DD779 Creature/0 R1469/S30839 Map: 552 Entry: 20978 Low: 5101433
[0] UNIT_FIELD_FLAGS: 34078720/9.991702E-38 - UNIT_FLAG_IN_COMBAT + UNIT_FLAG_NOT_SELECTABLE 
```

killerwife:
On encounter end, they are supposed to leave combat forcefully. That's all, no despawning or any other shenanigans.

Also @Yatzii93 please rewrite the PR accordingly and also:
m_npcEntryGuidCollection[pCreature->GetEntry()].push_back(pCreature->GetObjectGuid());
we already have everything we need here. Just do:
```cpp
GuidVector felfireVector; // at aggro felfire mobs always teleport to respawn location
m_pInstance->GetCreatureGuidVectorFromStorage(NPC_WRATH_SCRYER_FELFIRE, felfireVector);
```
inside the boss script.

Yatzii93:
Sounds good I'll change it up. Thanks for the input.

Yatzii93:
I can get them to evade but not sure how to send them home, surely despawn/respawn is the easiest way?

AnonXS:
no shenanigans, only forcefully evade which will make them drop the aura, hence not being able to aggro onto anything anymore.

Yatzii93:
shenanigans isnt really the right word, I can find u numerious examples where despawning is the preferred method to reset mods that won't return home on evade, but ye I guess if u want to follow the sniffs strictly and its not a despawn then I get it.

PR #332, opened 2019-08-05
Fix ATTUMEN_MOUNTED not registering in JustSummon

Yatzii93:
## 🍰 Pullrequest

In the current script it's trying to pick up Attun mounted in Midnights JustSummon function, however it's actually Attun that summons the mounted version of himself. Moving the code to the proper npc script resolves the issue of the code not triggering. Fixing this also leave the oppurtunity to resolve the quick corse delay. Going by the code in Creature.cpp right now boss npcs are 3 hours corpse decay, thus the reason for the three hours set in this corretion. 

### Proof

- https://github.com/cmangos/mangos-tbc/blob/22d87fc513e2aec47c41a3a3e19f57718c818a84/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/boss_midnight.cpp#L183
pAttun is the one casting the summon spell not Midnight

### Issues

- Fast corspe decay
- JustSummon code not triggering

### How2Test

- .tele kara
- Fight midnight until the mounted version is summoned
- Notice how the Justsummon function for the summon npc entry isnt triggered

### Todo / Checklist

- [X] None


killerwife:
Hm, why does the corpse decay need to be set if in Creature.cpp we set it to 3 hours. It should work for this mob as well. Care to explain more?

PR #333, opened 2019-08-11
Multi-threading and data race fixes - part 1.

evil-at-wow:
## 🍰 Pullrequest
This is the first part of a series of changes I'm working on to get some multi-threading issues and data races fixed. Because some of these changes aren't exactly trivial code changes, I'll be opening pull requests for them first, so others can review, test and discuss these changes before I commit them. All details are described in the commit messages.

### Issues
The Listener changes here in particular have been causing sporadic crashes on my system when shutting down the realm and/or world daemon, but I'm not aware of opened issues that seem to be caused by this.

### How to test
Testing can be done by checking normal server operation works as before. In particular users should still be able to log on and play for extended periods of time, and you should still be able to shut down the realm and world daemon as before.

### DIFF
--- a/src/shared/Database/Database.h
+++ b/src/shared/Database/Database.h
@@ -218,12 +218,12 @@ class Database
         // set this to allow async transactions
         // you should call it explicitly after your server successfully started up
         // NO ASYNC TRANSACTIONS DURING SERVER STARTUP - ONLY DURING RUNTIME!!!
-        void AllowAsyncTransactions() { m_bAllowAsyncTransactions = true; }
+        void AllowAsyncTransactions() { mAllowAsyncTransactions = true; }
 
     protected:
         Database() :
             m_nQueryConnPoolSize(1), m_pAsyncConn(nullptr), m_pResultQueue(nullptr),
-            m_threadBody(nullptr), m_delayThread(nullptr), m_bAllowAsyncTransactions(false),
+            m_threadBody(nullptr), m_delayThread(nullptr), mAllowAsyncTransactions(false),
             m_iStmtIndex(-1), m_logSQL(false), m_pingIntervallms(0)
         {
             m_nQueryCounter = -1;
@@ -267,7 +267,7 @@ class Database
         SqlDelayThread*     m_threadBody;                   ///< Pointer to delay sql executer (owned by m_delayThread)
         MaNGOS::Thread*     m_delayThread;                  ///< Pointer to executer thread
 
-        bool m_bAllowAsyncTransactions;                     ///< flag which specifies if async transactions are enabled
+        std::atomic<bool> mAllowAsyncTransactions;          ///< flag which specifies if async transactions are enabled
 
         // PREPARED STATEMENT REGISTRY
         typedef std::mutex LOCK_TYPE;
--- a/src/shared/Network/Listener.hpp
+++ b/src/shared/Network/Listener.hpp
@@ -33,8 +33,8 @@ namespace MaNGOS
     class Listener
     {
         private:
-            std::unique_ptr<boost::asio::io_service> m_service;
-            std::unique_ptr<boost::asio::ip::tcp::acceptor> m_acceptor;
+            boost::asio::io_service m_service;
+            boost::asio::ip::tcp::acceptor m_acceptor;
 
             std::thread m_acceptorThread;
             std::vector<std::unique_ptr<NetworkThread<SocketType>>> m_workerThreads;
@@ -60,7 +60,7 @@ namespace MaNGOS
 
                 return m_workerThreads[minIndex].get();
             }
-            
+
             void BeginAccept();
             void OnAccept(NetworkThread<SocketType> *worker, std::shared_ptr<SocketType> const& socket, const boost::system::error_code &ec);
 
@@ -71,7 +71,7 @@ namespace MaNGOS
 
     template <typename SocketType>
     Listener<SocketType>::Listener(std::string const& address, int port, int workerThreads)
-        : m_service(new boost::asio::io_service()), m_acceptor(new boost::asio::ip::tcp::acceptor(*m_service, boost::asio::ip::tcp::endpoint(boost::asio::ip::address::from_string(address), port)))
+    : m_service(), m_acceptor(m_service, boost::asio::ip::tcp::endpoint(boost::asio::ip::address::from_string(address), port))
     {
         m_workerThreads.reserve(workerThreads);
         for (auto i = 0; i < workerThreads; ++i)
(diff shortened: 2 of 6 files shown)


src/shared/Database/Database.h
```diff
@@ -267,7 +267,7 @@ class Database
         SqlDelayThread*     m_threadBody;                   ///< Pointer to delay sql executer (owned by m_delayThread)
         MaNGOS::Thread*     m_delayThread;                  ///< Pointer to executer thread
 
-        bool m_bAllowAsyncTransactions;                     ///< flag which specifies if async transactions are enabled
+        std::atomic<bool> mAllowAsyncTransactions;          ///< flag which specifies if async transactions are enabled
```
killerwife: I think the current codestyle is "m_allowAsyncTransactions.

evil-at-wow: Ah, good. I was in doubt about that after seeing various styles being used - though often in old(er) code. I checked the wiki page with the coding standard and it wasn't entirely clear about that, in the sense that the example class shows a member ```m_field``` but the text never mentions anything about it.

I'll change this, also in future changes, no problem.

cyberium:
Volatile to atomic changes can be merged without problem.
For Listener change it goes to the right direction for sure. I dont see anything that can cause trouble.
Major change is the the acceptor.close that will be called in its thread. Iam not sure how it will be working in the end. Closing will be delayed a bit? Well even in that case nothing should go wrong as the thread join later.

src/shared/Network/Listener.hpp
```diff
@@ -71,26 +71,26 @@ namespace MaNGOS
...
     template <typename SocketType>
     Listener<SocketType>::~Listener()
     {
-        m_acceptor->close();
-        m_service->stop();
+        // Close the acceptor. This will cancel any asynchronous accept
+        // operation and should stop the acceptor thread. Note that closing
+        // the acceptor needs to be done in the acceptor thread, because
+        // using the m_acceptor object from multiple threads is unsafe!
+        m_service.post( [this]() { m_acceptor.close(); } );
```
cyberium: Why not use
 m_acceptorThread.post([this]() { m_acceptor.close(); });
Should be more aligned with your comment

Laizerox: I have a bit of doubt here. I see you removed `m_service->stop()` which handles signaling to `io_service` / `io_context` to stop actually doing it.

Since we run `m_service` using `run()` in separate thread our `io_service` is never actually told to stop working even if we close `m_acceptor`.

While most likely program will exit. It will produce a "crash" due to thread was never joined.

evil-at-wow: @cyberium 
Remember that ```m_acceptorThread``` is an ```std::thread```, which has no post method, or any other way to let that ```std::thread``` run some code directly - you'll have to communicate with that thread by other means. So I'm using the ```io_service```'s post method to do that. Per the asio documentation:

>  This function is used to ask the io_service to execute the given handler, but without allowing the io_service to call the handler from inside this function.
> The io_service guarantees that the handler will only be called in a thread in which the run(), run_one(), poll() or poll_one() member functions is currently being invoked.

So that's exactly what we want here: a small piece of code that's guaranteed to be executed in the thread running ```m_service.run()```.

@Laizerox 
Indeed, that stop() signaled that the ```io_service``` should stop the event processing loop. But remember what the ```run()``` does:

> The run() function blocks until all work has finished and there are no more handlers to be dispatched, or until the io_service has been stopped.

And there you have it: a ```stop()``` is not the only way to stop the event processing. If there are no more handlers to be dispatched, event loop processing also stops. This is why ```io_service::work``` is sometimes used to avoid that ```run()``` returns right away (say when you're using the io_service in a thread pool to handle tasks that will be posted later).

The only reason why ```run()``` didn't return in the original code is because there's _always_ an ```async_accept``` ongoing. If its completion handler is dispatched, we just started another one via ```BeginAccept()```. Rinse and repeat. In that case you have to stop the event processing loop explicitly, as we did. But that had a small drawback: calling ```stop()``` did not guarantee that the ```io_service``` event processing still processed anything afterwards. In particular it did not guarantee the ```async_accept```'s completion handler was still dispatched (with a cancellation 'error'). And as you can see in ```OnAccept()``` that means you were not guaranteed to call ```worker->RemoveSocket(socket.get());``` when stopping. Oops (because that leads to other issues).

In the new code there's also an ```async_accept``` ongoing at any time in normal operation, so again the ```run()``` does not return. That is because, similar as before, we just start a new one in the completion handler. But as you can see in ```OnAccept()``` we no longer call ```BeginAccept()``` and start a new ```async_accept``` if ```m_acceptor``` has been closed! As a result, there's nothing more to be dispatched, and we fall in reason 1 why the ```run()``` stops the event processing loop. We don't have to explicitly tell it to stop, we just stop feeding it work so it stops by itself, but this time after it has dispatched all completion handlers, including the 'cancelled' one that calls ```worker->RemoveSocket(socket.get())```! Profit!

As for joining that thread, that still happens on the line below (```m_acceptorThread.join()```) so the Listener is not destroyed before that accept thread has fully finished and was joined, as intended. If the ```run()``` would never finish, you'll just hang forever in that ```join()``` waiting for a thread that doesn't exit.

evil-at-wow: I forgot to mention that calling ```close()``` on ```m_acceptor``` cancels the outstanding ```async_accept```. The asio documentation says:

> This function is used to close the acceptor. Any asynchronous accept operations will be cancelled immediately. 

So what you have here in the new code is the following sequence when the Listener gets destroyed:
* We call ```m_acceptor.close();``` (in the correct thread, via the ```post()```).
* This cancels the outstanding ```async_accept``` immediately.
* The event processing loop will call the completion handler, ```OnAccept```, with an 'error' to signal it was actually cancelled.
* We'll clean up by doing ```worker->RemoveSocket(socket.get());```. Then ```BeginAccept()``` is _not_ called because the acceptor has been closed.
* There is now nothing more to be dispatched, so the event processing loop stops, and ```run()``` "naturally" returns, and the ```m_acceptorThread``` finishes and becomes joinable.

cyberium: Nice explanation. A pleasure to read ;)
I have nothing to add. Iam sure you already tested it so feel free to push that whenever you want. Maybe @Yatzii93 or @SilvioDoMine can test this on their realm?

evil-at-wow: Thank you, and you're welcome.

And yes, needless to say I've tested a change like this. The change has also been included in my local builds since then, for almost 3 months now, without any related problems or crashes. But to be honest, that's not saying much, because my local development builds are only used by me to play and test. Let's say the opposite would have more value: if I'd run into issues with my local builds, then it's safe to say there's going to be a problem :laughing:

I like your idea of testing this on a more populated realm if that's possible, so even though you already agree I'll keep this open for a bit longer, at least until the weekend (I don't think we're in a real hurry here). Also to give other people still a chance to comment; after all this PR is only two days old.

evil-at-wow:
Six weeks have passed since the last comments, so I've pushed this to the repository in commits cmangos/mangos-tbc@f945fae3b4594c90d956c350976c6fa15ad15110, cmangos/mangos-tbc@42e86bb5574d2f2288eb8101d4640b4f94809d7c, cmangos/mangos-tbc@cd06e98b19420f2120a3be7daaae42c46aae9bb9 and cmangos/mangos-tbc@abfc369ce31014454e4ffddb49af185a0a2ba2e9.

PR #334, opened 2019-08-14
Implement CanLoot for creature lootable flag changes

Yatzii93:
## 🍰 Pullrequest

Add logic for circumstances where we want to set a creature to not lootable flag prior to the Loot check scripts being triggered.
### Proof

- https://youtu.be/akFAdhZnlNs
First is without my changes.
Second is with.

### Issues

- Although previous code was setting the flag successfully, the loot check scripts/code would just reapply the lootable flag, resulting in the mob still being lootable to players. I figure the reason for this is due to the fact that the creature is still infact alive while the JustDied function is being triggered. As soon as it does truly die, loot conditions are checked and ultimatly the core reapplies the flag we are trying to remove. (disclaimer here, I could well be wrong :P)

### How2Test

- .tele mgt
-.go c Priestess Delrissa
-kill Priestess before you kill her surrounding npcs
-Can you loot her? 

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/magisters_terrace/boss_priestess_delrissa.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/magisters_terrace/boss_priestess_delrissa.cpp
@@ -173,11 +173,10 @@ struct boss_priestess_delrissaAI : public ScriptedAI
         if (!m_pInstance)
             return;
 
-        // Remove lootable flag if the lackeys are not killed
         if (m_pInstance->GetData(TYPE_DELRISSA) == SPECIAL)
             m_pInstance->SetData(TYPE_DELRISSA, DONE);
         else
-            m_creature->RemoveFlag(UNIT_DYNAMIC_FLAGS, UNIT_DYNFLAG_LOOTABLE);
+            m_creature->SetCanLoot(false);
     }
 
     void UpdateAI(const uint32 uiDiff) override
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/magisters_terrace/boss_priestess_delrissa.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/magisters_terrace/boss_priestess_delrissa.cpp
@@ -173,6 +173,7 @@ struct boss_priestess_delrissaAI : public ScriptedAI
         if (!m_pInstance)
             return;
 
+        // Remove lootable flag if the lackeys are not killed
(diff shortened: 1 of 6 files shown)


Kittnz:
I can't remember seeing this on retail tho, custom feature?

Yatzii93:
This PR is to made it blizzlike. These mobs should not be lootable before the requirements are met otherwise the boss can be cheesed, either by the way of resetting it over and over for continues loot or take Sunwell twins as an example, without this fix you will get double the loot considering you can loot from both mobs. This is 100% not blizzlike.

AnonXS:
should use death prevention +
```
ServerToClient: SMSG_SPELL_INSTAKILL_LOG (0x2C33) Length: 32 ConnIdx: 1 Time: 08/13/2017 14:46:00.521 Number: 48400
Target: Full: 0x2030A0492017FC000057C7000010469E Creature/0 R3112/S22471 Map: 585 Entry: 24560 Low: 1066654
Caster: Full: 0x2030A0492017FC000057C7000010469E Creature/0 R3112/S22471 Map: 585 Entry: 24560 Low: 1066654
SpellID: 3617 (3617)

for Text: Not what I had... planned.
```

Yatzii93:
I guess if Blizzard handled this differently I'll just take down the PR, the boss scripts need to be rewritten. I'll keep the loot lock functionality my end either way considering personally I think it's a better way to do it than having a dead mob do it's death text longer after it was slain.

AnonXS:
Maybe they did it so you can find the loot after the prolonged fight? :x

Yatzii93:
It also wouldn't address Sunwell twins, even with a death prevention you would run into the same issue considering at some point that mob will need to die and ultimatly you end up with two lootable mobs, double the loot.

Do you have sniffs for this boss? https://wowwiki.fandom.com/wiki/Eredar_Twins

AnonXS:
```
Target: Full: 0x2016F448801893400039B0000011DCD4 Creature/0 R1469/S14768 Map: 580 Entry: 25165 Low: 1170644
Caster: Full: 0x2016F448801893400039B0000011DCD4 Creature/0 R1469/S14768 Map: 580 Entry: 25165 Low: 1170644
SpellID: 29878 (29878)
```
prob used to prevent loot from the first mob dying in this encounter, the second one dies normally.

ulfgebhardt:
@AnonXS can you provide the sniff of the whole fight. It might be useful to spot other differences, determine the order of spells and have the general context?
<3

AnonXS:
Contact @killerwife , if you want to have more insight.

PR #337, opened 2019-08-18
Dirty deeds Rogue Talent Fix

Yatzii93:
## 🍰 Pullrequest

Rogue talent isnt working at all, the current code fails to pick up the dirty deeds aura on the player.

This personal fix is pretty disgusting and could be much better but just bringing awareness to the problem and showing my personal fix. I did this change a little while ago, it works but ye it could use a revisit, I've got some unnesscary stuff in there thats for sure. At the time I was being a bit over-zelous with the notion of avoiding crashes lol.
### Proof

- None

### Issues

- Fixes rogue dirty deeds talent.

### How2Test

- Take the talent in the Sub tree on a rogue, get a mob to less than 35% and check the damage output. Currently it does not increase.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -7682,7 +7682,7 @@ uint32 Unit::MeleeDamageBonusDone(Unit* pVictim, uint32 pdamage, WeaponAttackTyp
         AuraList const& mOverrideClassScript = GetAurasByType(SPELL_AURA_OVERRIDE_CLASS_SCRIPTS);
         for (auto i : mOverrideClassScript)
         {
-            if (!(i->GetSpellProto() || i->GetSpellProto()->SpellIconID || spellProto->SpellFamilyName || i->GetMiscValue()))
+            if (!(i->GetSpellProto() && i->GetSpellProto()->SpellIconID && spellProto->SpellFamilyName && i->GetMiscValue()))
                 continue;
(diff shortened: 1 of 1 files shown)


killerwife:
Ehm. It does, DB is missing correct data. Give me result of following query:
select * from spell_affect where entry IN(14082,14083);

Yatzii93:
entry | effectId | SpellFamilyMask
-- | -- | --
14082 | 0 | 1280

AnonXS:
|entry | effectId | SpellFamilyMask|
|-- | -- | -- |
|14082|	0|	1280|
|14082|	1|	1164582191646|
|14082|	2|	1048832|

Yatzii93:
I'll give this a try ty.

AnonXS:
You should have this too by default...?

Yatzii93:
Don't seem to which is odd, I could of deleted the entries when testing but if these are the fix for the problem my end then they might of not been there either way. No idea why though my db is kept up to date and there hasent been changes to this table since I've been around or atleast to my knowledge.

killerwife:
Some time ago, SPELL_AURA_OVERRIDE_CLASS_SCRIPTS did not have the capability of having spell_affect entries and was producing errors. It is not a too old of a change in TBC core, maybe last year.

Yatzii93:
I see ty for the details Anon/Killer :)

Yatzii93:
I'll test today and if working I'll close the PR

PR #340, opened 2019-08-19
Fix incorrectly applied stealth detection increase

Yatzii93:
## 🍰 Pullrequest

Removes the increased stealth detection from Detect Traps(2836) when Unit visibility is being checked.
### Proof

- https://imgur.com/E83Zb6t with the Detect Trap spell learnt

### Issues

- Removes non-blizzlike stealth detection caused by Detect Traps bonus being used for unit calculations rather than just the trap object.

### How2Test

- 2 players, both stealth, check stealth visibility for both.
- .learn 2836 on 1
- Test stealth visability again.

### Todo / Checklist

- [X] None


killerwife:
Please notice that the aura 2836 has misc = 1 (1), and that is only the case for this particular aura for stealth detection. Please add a helper func which checks for this Misc and then uses it properly in proper places ^^ It is probably differentiation between Unit stealth hand GO stealth.

killerwife:
Fixed in https://github.com/cmangos/mangos-tbc/commit/1c56dd36a98c58163e8b8110a20d1439b1c1ac33

PR #342, opened 2019-09-15
Add LOS check to Bestrial Wrath

Yatzii93:
## 🍰 Pullrequest

Currently even if the hunters pet is out of LOS the cast will still go off but nothing will happen.

### Proof

https://youtu.be/LamuL4e4Y0E?t=48 - hunter spam casts Bestrial wrath when his pet is on the other side of the pillar. Whilst the pet is out of los the spell fails (I cannot make out the exact error) asoon as the pet comes into LOS it casts as normal.
### Issues

- Adds LOS cast check on hunters pet

### How2Test

- Be hunter. Cast 19574 whilst your pet is out of los. With change do the oppersite.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -7634,6 +7634,13 @@ SpellCastResult Spell::OnCheckCast(bool /*strict*/)
 {
     switch (m_spellInfo->Id)
     {
+        case 19574: //Hunter Bestrial Wrath pet LOS check
+        {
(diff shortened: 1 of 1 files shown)


killerwife:
The code here: https://github.com/cmangos/mangos-tbc/blob/master/src/game/Spells/Spell.cpp#L4700 should be doing what you propose, but it doesnt cos its in the wrong place. I will add a small refactor soonish.

PR #343, opened 2019-10-01
Fixes hunter's auto shooting

yukozh:
🍰 Pullrequest
This is fixing hunter's auto shooting, without this change, the auto shooting will be stucked at 0.0s if they click steady shot very fast.

### How2Test
Without this change, click the steady shot quickly, the auto shooting will be interrupted, displayed 0 second at client side. Within this change, after cast a steady shot, the auto shoot will cast successfully.

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -4345,7 +4345,8 @@ bool Unit::IsNonMeleeSpellCasted(bool withDelayed, bool skipChanneled, bool skip
             bool specialResult = true;
             if (forMovement) // mobs can move during spells without this flag
                 specialResult = genericSpell->m_spellInfo->InterruptFlags & SPELL_INTERRUPT_FLAG_MOVEMENT;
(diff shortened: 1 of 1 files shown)


killerwife:
1. What does this impact in the core? What call of this function?
2. Never use direct comparison with &, but use HasAttribute func instead

yukozh:
1. Without this, the following code will be triggered, and the auto shoot will be interrupted

```
void Unit::_UpdateAutoRepeatSpell()
{
    // check "real time" interrupts
    if (IsMoving() || IsNonMeleeSpellCasted(false, false, true))
    {
        // cancel wand shoot
        if (m_currentSpells[CURRENT_AUTOREPEAT_SPELL]->m_spellInfo->Category == 351)
            InterruptSpell(CURRENT_AUTOREPEAT_SPELL);
        // set 0.5 second wind-up time.
        m_AutoRepeatFirstCast = true;
        return;
    }
```

2. Updated.

PR #344, opened 2019-10-02
Adding null check for VisitGridObjects

yukozh:
## 🍰 Pullrequest
Adding null check for VisitGridObjects, to avoid somewhere pass in the null value of center_obj

### Issues
https://github.com/cmangos/issues/issues/2044

### DIFF
--- a/src/game/Grids/CellImpl.h
+++ b/src/game/Grids/CellImpl.h
@@ -196,6 +196,8 @@ inline void Cell::VisitWorldObjects(const WorldObject* center_obj, T& visitor, f
 template<class T>
 inline void Cell::VisitAllObjects(const WorldObject* center_obj, T& visitor, float radius, bool dont_load)
 {
+    if (!center_obj)
+        return;
     CellPair p(MaNGOS::ComputeCellPair(center_obj->GetPositionX(), center_obj->GetPositionY()));
     Cell cell(p);
     if (dont_load)


yukozh:
## Origin issue:

## 🐛 Bugreport


### Expected behavior
Should add a global catch to catch all of the script AI exceptions. If error happens, close the instance and log erros, not crash the whole process.

### Version & Environment
Client Version: 2.4.3" (TBC)
CMaNGOS Repo & Commit Hash:  TBC
Operating System: WIN10

### Steps to reproduce
1. attack the boss shirrak
2. after the boss focuses on someone, the mangosd get crashed

### Others
Got the dump from the crash process, it crashed at 

```c++
template<class T>
inline void Cell::VisitGridObjects(const WorldObject* center_obj, T& visitor, float radius, bool dont_load)
{
    CellPair p(MaNGOS::ComputeCellPair(center_obj->GetPositionX(), center_obj->GetPositionY()));
    Cell cell(p);
    if (dont_load)
        cell.SetNoCreate();
    TypeContainerVisitor<T, GridTypeMapContainer > gnotifier(visitor);
    cell.Visit(p, gnotifier, *center_obj->GetMap(), *center_obj, radius);
}
```

because the center_obj is null, a workaround is adding a null check in this method:

```c++
if (!center_obj)
    return;
```

The boss is casting spell #32301, but I did not find this ID in script boss_shirrak. Also, dunno after adding this workaround, which phase will be impacted.

cala:
Merged, thank you!

PR #345, opened 2019-10-10
Fixes wrong erase operation in CreatureLinkingMgr

yukozh:
Repro:
- Remove any one creature instance from Magtheridon
- Aggro, sometimes causes crash


Yatzii93:
I did this a while ago and can confirm haven't had a problem since.

killerwife:
Same. I will merge this later this week.

PR #351, opened 2019-11-19
Add dist and angle to creature linking.

miraco:
# 🍰 Pullrequest
This is basic handling for creature groups that gets angle and dist for follow movement from database.
Its needed for Groups like the waves at Dark portal that spawn mostly on the same point and use bigger distances/other angle.

### Proof
Here you can see that they get spawned with an dist from ~1
68291549-62b90180-008a-11ea-8be0-523bd96a0f0c

But moving with a distance from ~5
68291581-73697780-008a-11ea-9b48-42996cc6d81a

### Issues
Should maybe use some more checks | not use FLAG_FOLLOW_WITH_ANGLE 

# #https://github.com/cmangos/issues/issues/2062


src/game/Entities/CreatureLinkingMgr.h
```diff
@@ -71,6 +71,7 @@ enum CreatureLinkingFlags
 
     // Dynamic behaviour, out of combat
     FLAG_FOLLOW                     = 0x0200,
+    FLAG_FOLLOW_WITH_ANGLE			= 0x0201,
```
lukaasm: FLAG_FOLLOW_WITH_ANGLE = 0x0201 is wrong enum mask in this case, because 0x0201 is a concatenation of FLAG_AGGRO_ON_AGGRO( 0x0001 ) | FLAG_FOLLOW( 0x0200 )

You should use FLAG_FOLLOW_WITH_ANGLE = 0x0400

cala:
Poke @cyberium on this.

miraco:
I think the whole group movement need a rework with its own FollowMovementGenerator. i am currently not rly able to do this, to much work irl.

cyberium:
Yes more work is needed for this to work in all over different cases in standard way and fully driven by db.
Thank anyway for your work and pr.

PR #354, opened 2019-12-15
Set DualWield & Damage On Creature Equipment Updates.

Yatzii93:
## 🍰 Pullrequest
When a creature changes weapons or goes from not dual wielding to dual wielding. I notice that the function SetCandualWield is not being applied unless individually coded into a creatures script. This would ensure any equipment changes correctly setup a mob to dual wield or undo dual wielding whilst also ensuring correct damage ouput.

Could well be the wrong way to go about this, if anyone has any input feel free.

### Proof

- None

### Issues

- One notable bosses where this comes into play is Prince in karazhan. In the second phase he equips his axes and starts dual wielding, the result is an offhand attack that does between 1-5 damage. Then when he unequips the axes in the last phase he continues to be set to dual wield which is incorrect. You can set it manually in the script but I thought a blanket check would be ideal rather than doing it individually.

### How2Test

- Fight Prince Malchezaar in Karazhan. Without the change in P2 his offhand attack will be really low and in P3 he will still continue to do dual attacks with his fists.

### Todo / Checklist

- See if theres a blanket way to apply 
SetCanDualWield(true); - If dualwielding.
&
UpdateDamagePhysical(BASE_ATTACK);
on creature spawn if they have equipment/dualW. Princes axe being a good example, right now without doing the above code on spawn it to will do both a 1-4 damage offhand and mainhand attack. Also occurs with ranged equipment, one example being the Ambushers during the Lurker Below fight in SSC.

### DIFF
--- a/src/game/AI/ScriptDevAI/include/sc_creature.cpp
+++ b/src/game/AI/ScriptDevAI/include/sc_creature.cpp
@@ -363,13 +363,24 @@ void ScriptedAI::SetEquipmentSlots(bool loadDefault, int32 mainHand, int32 offHa
     }
 
     if (mainHand >= 0)
+    { 
         m_creature->SetVirtualItem(VIRTUAL_ITEM_SLOT_0, mainHand);
+        m_creature->UpdateDamagePhysical(BASE_ATTACK);            
+    }
 
     if (offHand >= 0)
+    { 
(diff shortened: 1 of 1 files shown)


killerwife:
Hm, I added handling for this to dual weild spells. Are you absolutely sure these mobs dont use a dual weild aura instead?

killerwife:
In a typical willerkaifu fashion I have checked it out myself and I find your PR correct.

PR #357, opened 2019-12-28
Fix failure to update GUARDIAN_PET damage stats.

Yatzii93:
## 🍰 Pullrequest

GUARDIAN_PET's are doing 1-3 damage right now due to a return where a break should be.
### Proof

- None

### Issues

- Fixes all GUARDIAN_PET doing between 1-3 damage.

### How2Test

- [X] None

### DIFF
--- a/src/game/Entities/Pet.cpp
+++ b/src/game/Entities/Pet.cpp
@@ -1417,7 +1417,7 @@ void Pet::InitStatsForLevel(uint32 petlevel)
         case GUARDIAN_PET:
         {
             SelectLevel(petlevel);  // guardians reuse CLS function SelectLevel, so we stop here
-            return;
+            break;
         }
         default:
             sLog.outError("Pet have incorrect type (%u) for level handling.", getPetType());


Yatzii93:
@cala concerning my recent post about damage not being correct for G_pets.

Yatzii93:
@killerwife is this not the fix?

killerwife:
Turns out the issue is deep. This fixes some guardians but there is also a case where guardians have twice as much hp. I will be addressing it soonish in my export.

cala:
@killerwife is right. I added this `return` as part of a change because guardian pets had their stats calculated twice resulting in overpowered pets, especially at low levels.

From my tests, the HP/mana issue was fixed but I did not look at the damage side however. So issues can still exist and seem to.

There are many changes occurring right now in the whole stats area, especially in Classic DB but you can count on @killerwife and myself to square things down.

Yatzii93:
sounds good thanks for the info guys. I seems that putting the break back in doesn't cause the double hp/power issue atleast from looking at the guardians I was testing with. I'd need to break down the code though to see whats being applied additionally now theres no return. But ye from my quick 4am testing lastnight, it seems ok.

cala:
Moreover, summoning mechanics were changed between Classic and TBC thus it is possible that I overlooked something were porting from Classic (where the initial issue was) to the other cores.

killerwife:
@Yatzii93 please retest mangos-tbc as I've pushed the fix to all of this.

killerwife:
Fixed by
https://github.com/cmangos/mangos-tbc/commit/5ece1f5c746e6cee16eb90a9eda1374a861fc9c8
and
https://github.com/cmangos/mangos-tbc/commit/d7937881925bbbefa3fdc441c69cdad5521e6c0e

PR #364, opened 2020-01-01
Implement per-Map multithreading

killerwife:
Initial idea - TC
Main author - Laizerox

This commit implements per-map scheduled multithreading. Cross map operations are safeguarded after my last year efforts. This includes OutdoorPvP or WorldState scripts.

Please review for any potential improvements as this code is from 2016.

@cyberium and @evil-at-wow


src/game/Maps/MapUpdater.cpp
```diff
@@ -0,0 +1,101 @@
...
+void MapUpdater::WorkerThread()
+{
+    while (true)
+    {
+        Worker* request = nullptr;
+
+        _queue.WaitAndPop(request);
+
+        if (_cancelationToken)
+            return;
```
domgho: Delete request in return case?

killerwife: Nice catch. Will definitely fix it.

src/game/Maps/MapUpdater.h
```diff
@@ -0,0 +1,58 @@
...
+        void activate(size_t num_threads);
+        void deactivate();
+        void wait();
+        void join();
+        bool activated();
+        void update_finished();
+        void schedule_update(Worker* worker);
+
+    private:
+        ProducerConsumerQueue<Worker *> _queue;
```
evil-at-wow: I would rename the members to use our code style, with m_ prefix. It's only six easy search & replaces after all.

src/game/Maps/MapUpdater.h
```diff
@@ -0,0 +1,58 @@
...
+
+#ifndef _MAP_UPDATER_H_INCLUDED
+#define _MAP_UPDATER_H_INCLUDED
+
+#include "Platform/Define.h"
+#include "ProducerConsumerQueue.h"
+
+#include <mutex>
+#include <thread>
+#include <condition_variable>
```
evil-at-wow: I spot missing ```<atomic>``` and ```<vector>``` includes here. They probably get included indirectly, but let's not rely on that.

src/game/Maps/MapUpdater.h
```diff
@@ -0,0 +1,58 @@
...
+
+class Worker;
+
+class MapUpdater
+{
+    public:
+        MapUpdater() : _cancelationToken(false), pending_requests(0) {}
+        MapUpdater(size_t num_threads);
```
evil-at-wow: This constructor doesn't seem to be used anywhere?

It's also probably a good idea to make this class non-copyable, so add a deleted copy-constructor and assignment operator.

src/game/Maps/MapUpdater.h
```diff
@@ -0,0 +1,58 @@
...
+        
+        void activate(size_t num_threads);
+        void deactivate();
+        void wait();
+        void join();
+        bool activated();
+        void update_finished();
```
evil-at-wow: This one also doesn't seem to be used anywhere?

src/game/Maps/MapUpdater.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+void MapUpdater::wait()
+{
+    std::unique_lock<std::mutex> lock(_lock);
+
+    while (pending_requests > 0)
+        _condition.wait(lock);
+
+    lock.unlock();
```
evil-at-wow: Not needed, because unique_lock's destructor does that for you.

src/game/Maps/MapUpdater.cpp
```diff
@@ -0,0 +1,101 @@
...
+        return;
+
+    for (size_t i = 0; i < num_threads; ++i)
+        _workerThreads.push_back(std::thread(&MapUpdater::WorkerThread, this));
+}
+
+void MapUpdater::deactivate()
+{
+    _cancelationToken = true;
+
```
evil-at-wow: By setting the cancelationToken, your worker threads won't be executing any new updates any more and will exit after having finished any ongoing update, if any, so the ```wait()``` below seems unnecessary. I think you might as well just _queue.Cancel() and then join the worker threads, which will wait for them to exit anyway.

This also happens to avoid a double ```wait()``` in ```MapUpdater::join()``` below (on directly, one via ```deactivate()```.

src/shared/ProducerConsumerQueue.h
```diff
@@ -0,0 +1,113 @@
...
+        std::atomic<bool> _shutdown;
+
+    public:
+
+        ProducerConsumerQueue<T>() : _shutdown(false) { }
+
+        void Push(const T& value)
+        {
+            std::lock_guard<std::mutex> lock(_queueLock);
+            _queue.push(std::move(value));
```
evil-at-wow: The std::move() is deceiving here because value is a const T&, so you're unlikely to call the push you think you're calling. Just remove the std::move (remember, that's just a cast, it's not actually moving anything), or enable real move semantics by having a ```Push( T&& value )``` as well.

Also, this class is perhaps worth making non-copyable.

killerwife:
Nice comments @evil-at-wow I will adjust accordingly. I do not think there is anything I do not agree with. When it comes to thread safety, I have been adding thread safety to this for the past 2-3 years. I agree, that I still might've misses something though. (note how I solved worldstate and outdoorpvp)

src/game/Maps/MapUpdater.h
killerwife: GetWorker().update_finished(); uses this in MapWorkers.h

src/game/Maps/MapUpdater.h
killerwife: It will be used in a future update. Keeping it synced for simplicity of merging.

evil-at-wow:
I'm not sure why you were in such a hurry to get this merged. I personally think changes like this deserve a lot more time (and hopefully attention) than a single week before getting commited. I mean, I barely had the chance to do a first quick pass of code review (I know I don't have a lot of spare time to spend here, but still). But it is what it is, and I'll comment on the actual submit about all the things you didn't adjust :grin: 

The one thing I still wanted to comment on here is the thread-safety. My remark about the core likely not being ready for a change like this wasn't directed at you, or this particular change. I know you've changed a few things like you mentioned, which is obviously a good thing, and there have been some other efforts too (see also my first part of fixing threading issues last summer in cmangos/mangos-tbc#333). But there's still a lot of code that is used from multiple threads while it's not thread-safe at all, and even more so now with this. Sooner or later one gets unlucky, and we both know those issues are hard to debug. Except for fixing those issues eventually there's not really much else we can do (we could argue about changes like this forever, but I don't think that's very useful). One thing I'd like to suggest to keep in mind is that if 'weird' issues start showing up in the issue tracker - so far I seem to be the only one continuously running into them for some reason - then one of the first things we should ask is to try again with just 1 map update thread. That will avoid at least some of the issues until they get fixed.

killerwife:
Well, this has been in my pipeline for 4 years and @Yatzii93 is using a similar version of this. Without this code, the core won't handle over 500 players. I merely wanted to have a level plane field for everyone using cmangos. The code used from "multiple threads" is not used from multiple threads because only world context performs changes on them, so during map execution they do not change. I am aware there is a very fine line I am walking here, and many systems need review, but from experience people do not test PRs on a large scale, ever. I will incorporate any feedback you give me though.

evil-at-wow:
> from experience people do not test PRs on a large scale, ever

I fully agree with that statement. My threading fixes PR also didn't get any testing feedback, for example.

But people often comment and/or do some (superficial) reviews on PRs. Perhaps less than we'd want, but still. So allowing a bit more time for that seems a reasonable request - my PR for example was around for 6 weeks before I merged it to give everyone a fair chance to comment and/or test. If this has been in your pipeline for 4 years, a week or two more or less wouldn't have made a difference anyway. And in this case specifically, I even announced I'd do a more in-depth check later and probably some testing as well. I'm sorry that was not done the day after, but I have a busy (professional) life as well.

I do understand and even agree you eventually want to get it done. All I'm suggesting is to leave it open a bit longer. Anyway, it's a done deal, so let's move on!

> the core won't handle over 500 players

Hmm. One wonders where things went wrong then. I have no relevant experience because I've never run a server that even gets close to that amount of players, but I remember seeing posts on forums that ~2k players was possible like 10 years ago, maybe a bit higher with some tweaking/high end hardware. And that was on hardware from 10 years ago.

As limited as my own experience is, it does seem to match with my own observations. A server I installed in 2013 for a few friends (< 10 players) was barely using any CPU, ever. If I run today's core with the same small population, on a much faster machine, it's using a lot more CPU. Of course, with less than 10 players that's still not going to be a problem, but I can imagine it's a different story with hundreds.

Sure, there's a lot of things that weren't working in the core a decade ago, and for sure there's a lot we're doing nowadays we weren't doing back then, and for good reasons. But we also got hardware that's probably three times more powerful, so you'd think that levels things out a bit. An extra factor of 4 (500 vs 2k) is a lot, and more than I expected.

It could be worth checking out where we're losing so much performance. Not that I expect one to easily handle 20k players with a single thread, but 500 players sounds really low on modern hardware. I might give it a try one day... but no promises.

killerwife:
Everyone is running this (i.e. trinitycore| multithreading. That's why people run 2k+ people. Also we keep adding and adding functionality to  this day. That does come at an expense despite our best efforts.

The population performance issue is bigger in TBC and WOTLK. This also enables all grids loaded for the first time to run without insane lag. (and thats how I used valgrind to determine most of this stuff is safe).

What I meant that this was in my pipeline, is that this code ran on a 50-1000 people server on which we eventually fixed 99% of crashes (in august last year). It has been thoroughly battle tested.

All of that being said, I 100% understand your concerns.

evil-at-wow:
> Everyone is running this (i.e. trinitycore| multithreading. That's why people run 2k+ people.

Nowadays, sure, but there was no TC back when people were already running 2k servers (disclaimer: if we can believe the internet about the 2k). And the only thing MaNGOS had multi-threaded back then was socket processing I think (aka the network threads)? Anyway, as we both mentioned, a lot has been added since, and that's not free indeed.

>  This also enables all grids loaded for the first time to run without insane lag.

This is an interesting statement. I didn't check the consequences of this PR yet, but I assume this means we're loading grids from multiple threads too now?

The reason why I bring this up is that I even got trouble with this code (that was changed a good while ago, supposedly to start up faster):
```
void MapManager::CreateContinents()
{
    std::vector<std::future<void>> futures;
    uint32 continents[] = { 0, 1, 530 };
    for (auto id : continents)
    {
        Map* m = new WorldMap(id, i_gridCleanUpDelay);
        // add map into container
        i_maps[MapID(id)] = m;

        // non-instanceable maps always expected have saved state
        futures.push_back(std::async(std::launch::async, std::bind(&Map::Initialize, m, true)));
    }

    for (auto& futurItr : futures)
        futurItr.wait();
}
```
I found out running Initialize on multiple maps in parallel is not thread-safe (hence my trouble). There's a few things I found causing trouble, but mostly manipulating things in VMapManager2 and MMapManager calls (initializing a maps loads creatures, and things like GetZoneId() get to TerrainInfo and to vmap/mmap stuff eventually). I should check my notes for details though.

I'm sure we'll eventually get all this sorted out and fixed though, but it's good to hear it has been battle tested on a busy server! It looks like my machine makes a good sparring partner for that last 1% battle testing :grin:

cyberium:
I did add that code and if you found the commit in question you'll find that i had to handle vmap and mmap loading by adding mutexes. Its in the end needed as we can load them same time by different player even without that speedup code.
I didin't encountered nor see any complain about that even if load all grids. Ofc that doesn't mean i didn't forget something somewhere. I'll be happy to know what you have find.

PR #365, opened 2020-01-04
Tree of life passive bonus to work with Swiftmend

Yatzii93:
## 🍰 Pullrequest

Heals from swiftmend are currently not taking into account the tree of life aura bonus where 25% of the tree formed healer's total spirit should be included onto the swiftmend heal amount as a bonus.
### Proof

- https://imgur.com/a/wb3zjYz **Without** the fix. First not in tree form, second in tree form.
- https://imgur.com/a/zX3Dvm5 **With** the fix. First is in tree form second without. (differnet char to first - base spirit count is 133)

### Issues

- Swiftmend heal not taking into account tree of life healing bonus

### How2Test

- Go druid, learn treeform
- Cast tree form, pop a rejuv and use swiftmend instantly. See heal.
- Repeat steps without going into treeform.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Spells/SpellEffects.cpp
+++ b/src/game/Spells/SpellEffects.cpp
@@ -3988,7 +3988,10 @@ void Spell::EffectHeal(SpellEffectIndex eff_idx)
             int32 tickcount = GetSpellDuration(targetAura->GetSpellProto()) / targetAura->GetSpellProto()->EffectAmplitude[idx];
 
             unitTarget->RemoveAurasByCasterSpell(targetAura->GetId(), targetAura->GetCasterGuid());
-
+            
+            if(unitTarget->HasAura(34123)) //Tree Of Life Bonus Heal Passive
+                addhealth += int32(m_caster->GetStat(STAT_SPIRIT) / 4);
+            
             addhealth += tickheal * tickcount;
         }
         else if (m_spellInfo->SpellFamilyName == SPELLFAMILY_POTION)


killerwife:
Id: 35 (SPELL_EFFECT_APPLY_AREA_AURA_PARTY)
BasePoints = 1
Targets (1, 0) (TARGET_UNIT_CASTER, TARGET_NONE)
Aura Id 115 (SPELL_AURA_MOD_HEALING), value = 1, misc = 127 (127), miscB = 0, periodic = 0
Radius (Id 11) 45.00

Try to find out why this aura type is not pulling the correct spirit enhanced value. Sadly this PR is not a real fix.

Yatzii93:
Hmm I was told it works for ticks. I'll double check.

Yatzii93:
ok so the way to fix the MOD_HEALING being calculated is to at a 1 to direct bonus in the db.

"entry"	"direct_bonus"	"dot_bonus"	"ap_bonus"	"ap_dot_bonus"	"comments"
"18562"	"1"	"0"	"0"	"0"	"Druid - Swiftmend"

not doing it makes

> TakenTotal = pCaster->SpellBonusWithCoeffs(spellProto, TakenTotal, TakenAdvertisedBenefit, 0, damagetype, false); 

....
final calculation -  

> total += int32(benefit * coeff * LvlPenalty);

coeff variable 0 which just removes the bonus. 

Here it implies https://wowwiki.fandom.com/wiki/Spell_power_coefficient?oldid=1492745
the direct_bonus should be 1.

It does make me wonder, if a spell has 0 direct_bonus in the db does that mean any increased healing taken by this aura type, essentially gets removed? If so I can't see how that's intended.

Let me know if I'm off base here.

killerwife:
Check here: https://wowwiki.fandom.com/wiki/Spell_power_coefficient
The bonus for swiftmend is drawn from the consumed spell, which already benefited from SP, and this would result in double dipping. As a result this is non-issue.

Yatzii93:
That's fair, makes sense, but ultimately that doesn't mean the issue is resolved. As I said with no direct bonus in the db the SPELL_AURA_MOD_HEALING part of the calculated gets essentially removed. 

So the only way I see to fix swiftmend to take SPELL_AURA_MOD_HEALING into account is to just seperately add it into the swiftmend code. 

> int32 tickheal = targetAura->GetModifier()->m_amount;
> int32 tickcount = GetSpellDuration(targetAura->GetSpellProto()) / targetAura->GetSpellProto()->EffectAmplitude[idx];
> **int32 HealingTakenBonus = caster->SpellBaseHealingBonusTaken(GetSpellSchoolMask(targetAura->GetSpellProto()));**
> unitTarget->RemoveAurasByCasterSpell(targetAura->GetId(), targetAura->GetCasterGuid());
> **addhealth += HealingTakenBonus;**
> addhealth += tickheal * tickcount;

Yatzii93:
Error there should be 
 int32 HealingTakenBonus = unitTarget->SpellBaseHealingBonusTaken(GetSpellSchoolMask(targetAura->GetSpellProto()));

killerwife:
No, if you want that, just put it in db. But its a custom change so it doesnt belong to cmangos.

Yatzii93:
So you are confident that swiftmend doesnt take into account any healing mods on the player it's healing. Surely that cant be right?

AnonXS:
Just some gist:

`Consumes a Rejuvenation or Regrowth effect on a friendly target to instantly heal them an amount equal to 12 sec. of Rejuvenation or 18 sec. of Regrowth.`

`Cataclysm Patch 4.0.1 (2010-10-12): Changed from selectable talent to a specialization perk. Heals for a fixed amount rather than based on the HoT it consumes`

Based on the HoT it consumes means to me that the effect consumed is responsible for how big of a heal is set off by it, unrelated to currently active mods.

https://wowwiki.fandom.com/wiki/Swiftmend?oldid=1562661
https://wow.gamepedia.com/Swiftmend

killerwife:
https://wowwiki.fandom.com/wiki/Spell_power_coefficient

Yatzii93:
I mean I understand where you are coming from but I havent read anything that definitly says swiftmend does a fixed amount completely ignoring any healing bonus on the target. Isn't spell power coeff refering to spell power on the caster. Not the healing bonus on the target.

killerwife:
Yes, it wont increase, because it already increased the healing of the HoT spell on him.

Yatzii93:
ok, aslong as you're saying the increased amount added to the hot does nothing to the swiftmend amount then it's agreed with or without healing mods bonuses on the target the swiftmend heal is a fixed amount unaffected by healing mods on the target or any increase on the hot by healing mods.

PR #368, opened 2020-01-27
Fix Arena rating difference force queue update.

Yatzii93:
## 🍰 Pullrequest

Ensures the force update checks the right bracket. Right now the bottom bracket is being checked which doesn't have the queued players in, this results in the very first check shown below, returning.

>  if (m_QueuedGroups[bracket_id][BG_QUEUE_PREMADE_ALLIANCE].empty() &&
>             m_QueuedGroups[bracket_id][BG_QUEUE_PREMADE_HORDE].empty() &&
>             m_QueuedGroups[bracket_id][BG_QUEUE_NORMAL_ALLIANCE].empty() &&
>             m_QueuedGroups[bracket_id][BG_QUEUE_NORMAL_HORDE].empty())
>         return;

### Proof

- See below explanation in comments thread.

### Issues

- Force update has the wrong bracket id set. Queued players right now are sitting in the first not last.

### How2Test

- enable rating difference in configs. Have two teams queue with team rating outside the set rating difference. After the discard timer is met by both teams, they won't be queued into eachother because the force queue update is failing. 

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/BattleGround/BattleGroundMgr.cpp
+++ b/src/game/BattleGround/BattleGroundMgr.cpp
@@ -1168,9 +1168,9 @@ void BattleGroundMgr::Update(uint32 diff)
         {
             // forced update for level 70 rated arenas
             DEBUG_LOG("BattleGroundMgr: UPDATING ARENA QUEUES");
-            m_BattleGroundQueues[BATTLEGROUND_QUEUE_2v2].Update(BATTLEGROUND_AA, BG_BRACKET_ID_LAST, ARENA_TYPE_2v2, true, 0);
-            m_BattleGroundQueues[BATTLEGROUND_QUEUE_3v3].Update(BATTLEGROUND_AA, BG_BRACKET_ID_LAST, ARENA_TYPE_3v3, true, 0);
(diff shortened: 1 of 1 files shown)


killerwife:
My head hurts just from looking at this PR. I have 0 inner arena workings knowledge on this :D

Yatzii93:
No worries, it took me a good part of a day to get my head around how it all worked. I can show some proof if you'd like that queued players sit in the first bracket, expect it tomorrow at some point.

Yatzii93:
So if you look at the imagine below this is a group queued into arena.
image

They are sitting in the first bracket id. Not the last. So when the force update does the check right now with the main repo code, it looks in the wrong bracket and just decides there's no one queued. 

(Image below has my fix but just showing where the force update takes place)
image

Ultimately what this causes is a broken queue system if match making is active and the teams in queue are outside of the rating difference set. 
So to give an example if the arena rating differnet is set to 200 in the config. If both a 1500 rated team and a 1900 team queue, even if the discard timer is met (also set in config), it will never queue them into eachother because the force update essentially updates the wrong bracket which will be empty. 

What does still work though is if one requeues after the other has been in the queue long enough, and this only works because it uses a differnet set of code shown below when someone joins the queue.
image

killerwife:
Thank you for the explanation, it helped quite a lot. Small trivia note: The issue originated because someone fixed in it wotlk and never ported it. But wotlk has two arena brackets with rating (tbc has one) and that system is not 1:1 with the tbc one.

PR #372, opened 2020-02-02
Fix distance checks to reach all caged orcs

Yatzii93:
## 🍰 Pullrequest

Right now the checks are to short, only the orcs up against the gate are being selected. The result is back orcs that can be aggro'd which causes some hilarious buggy interaction whereas all the orcs in that cage will aggro, including the one that isn't selectable, a new friend some would say, a combat bug others. Either way this resolves the issue.
### Proof

- None

### Issues

- as described above

### How2Test

- Go to third boss in Blood Furnace, notice how you can target the orcs furtherest in the back of the cages.

### Todo / Checklist

- [X] None


AnonXS:
This worked befor.

killerwife:
Have you tested it on a clean cmangos?

MantisLord:
Confirmed the issue on current core. This is a valid fix.

killerwife:
This issue occured due to me deleting wrongful override of GO bounding radius calculation during proper interact distance of GOs being added. Good find @Yatzii93

PR #373, opened 2020-02-09
Allow CLion to properly configure Cmake in Windows

henhouse:
## 🍰 Pullrequest
JetBrain's CLion IDE works really well out of the box with any project using CMake. This can be used as an alternative to Microsoft Visual Studio for those familiar with the JetBrains environment. In order to work properly, it just needs to read the `${CMAKE_BUILD_TYPE}` variable instead of the `$(Configuration)` which is only present when in MSVS, I believe.

I have not tested this on Linux or Mac's CLion as this fix is intended for building with the MS VS compilers (which is where the problem therein lies). 

Another note: this isn't really adding support for another IDE since CLion only utilizes CMake and is agnostic to which build tools you use. This just provides it the right config variable in order to build using MS VS compilers.

### Proof

- Just build normally without this branch in CLion and it will fail saying it could not link the .exes to `"x64_/blabla`" missing the build type variable.

### Issues

- None

### How2Test

- Download the latest CLion from JetBrains. If you're a student, you can get the educational pack for free (I think there are additional options for open source projects). Else, you can evaluate it for 30 days or so.
- Load the project root folder in and let Cmake configure
- Build
- Build should succeed with executables in the appropriate Debug or Release folder.


cyberium:
Merged Thank you

PR #375, opened 2020-02-21
Add SelectFlag forr Interupts (check for casting players)

miraco:
## 🍰 Pullrequest

Adding new flag for SelectAttackingTarget. This is getting used from mobs that use interupt spells like pummel/silence on Players that are casting spells.
### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -1700,6 +1700,11 @@ inline Unit* CreatureEventAI::GetTargetByType(uint32 target, Unit* actionInvoker
             if (!resTarget)
                 isError = true;
             return resTarget;
+        case TARGET_T_HOSTILE_RANDOM_CASTING:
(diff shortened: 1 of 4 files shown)


al3xc1985:
with this patch the mobs search for chars in range and pummel them?
for example he is fighting with a warrior and a priest is healling the warrior, does the mob search in radius for priest and pummel/silence him?

AnonXS:
Adding smart targetflags is better then adding additonal events :+1

killerwife:
I suggest adding it to enum CastFlags instead and then to case ACTION_T_CAST: the same way I added CAST_PLAYER_ONLY
It is less code and easier to setup in the future. Also, which mobs utilize this. @AnonXS any idea?

AnonXS:
Yea CAST_TARGET_CASTING to select a Target from the TargetFlags that is actually currently casting, even better.

miraco:
https://github.com/cmangos/mangos-tbc/pull/415
Thats the better version then i guess

PR #378, opened 2020-02-28
Fix dangling pointers in debugcmds

RyadaProductions:
## 🍰 Pullrequest
Fixes some dangling pointers in the debugcommands

### Proof
- None

### Issues
- fixes #2029

### How2Test
- Ask @Xazax-hun If they can run the new compiler warnings on the code again.

### Todo / Checklist
- [ ] The `GetGuidStr().c_str()` calls might also have the same issue, as it just wraps `GetGuidObject().GetString()`. 
But would like confirmation that those calls exhibit the same problem, before i proceed

### DIFF
--- a/src/game/Chat/debugcmds.cpp
+++ b/src/game/Chat/debugcmds.cpp
@@ -384,10 +384,13 @@ bool ChatHandler::HandleDebugGetLootRecipientCommand(char* /*args*/)
     if (!target->HasLootRecipient())
         SendSysMessage("loot recipient: no loot recipient");
     else if (Player* recipient = target->GetLootRecipient())
+    {
+        std::string targetLootRecipientGuidString = target->GetLootRecipientGuid().GetString();
         PSendSysMessage("loot recipient: %s with raw data %s from group %u",
-                        recipient->GetGuidStr().c_str(),
-                        target->GetLootRecipientGuid().GetString().c_str(),
-                        target->GetLootGroupRecipientId());
+            recipient->GetGuidStr().c_str(),
+            targetLootRecipientGuidString.c_str(),
+            target->GetLootGroupRecipientId());
+    }
     else
         SendSysMessage("loot recipient: offline ");
 
@@ -454,9 +457,12 @@ bool ChatHandler::HandleDebugGetItemStateCommand(char* args)
             if (!item) continue;
             if (!item->IsBag())
             {
-                if (item->GetState() == state)
+                if (item->GetState() == state) 
+                {
+                    std::string itemOwnerGuidString = item->GetOwnerGuid().GetString();
                     PSendSysMessage("%s bag: 255 slot: %u owner: %s",
-                                    item->GetGuidStr().c_str(),  uint32(item->GetSlot()), item->GetOwnerGuid().GetString().c_str());
+                        item->GetGuidStr().c_str(), uint32(item->GetSlot()), itemOwnerGuidString.c_str());
+                }
             }
             else
             {
@@ -464,10 +470,13 @@ bool ChatHandler::HandleDebugGetItemStateCommand(char* args)
                 for (uint8 j = 0; j < bag->GetBagSize(); ++j)
                 {
                     Item* item2 = bag->GetItemByPos(j);
-                    if (item2 && item2->GetState() == state)
+                    if (item2 && item2->GetState() == state) 
+                    {
+                        std::string item2OwnerGuidString = item2->GetOwnerGuid().GetString();
                         PSendSysMessage("%s bag: %u slot: %u owner: %s",
-                                        item2->GetGuidStr().c_str(), uint32(item2->GetBagSlot()), uint32(item2->GetSlot()),
-                                        item2->GetOwnerGuid().GetString().c_str());
+                            item2->GetGuidStr().c_str(), uint32(item2->GetBagSlot()), uint32(item2->GetSlot()),
+                            item2OwnerGuidString.c_str());
+                    }
                 }
             }
         }
@@ -520,9 +529,10 @@ bool ChatHandler::HandleDebugGetItemStateCommand(char* args)
 
             if (item->GetOwnerGuid() != player->GetObjectGuid())
             {
+                std::string itemOwnerGuidString = item->GetOwnerGuid().GetString();
                 PSendSysMessage("%s at slot %u owner (%s) and inventory owner (%s) don't match!",
                                 item->GetGuidStr().c_str(), uint32(item->GetSlot()),
-                                item->GetOwnerGuid().GetString().c_str(), player->GetGuidStr().c_str());
+                                itemOwnerGuidString.c_str(), player->GetGuidStr().c_str());
                 error = true; continue;
             }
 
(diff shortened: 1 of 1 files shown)


evil-at-wow:
I'd be interested to see what warnings @Xazax-hun got related to this, because the code seems correct at first sight.

In C++, temporary objects are destroyed as the last step in evaluating the full-expression that contains their creation (with I believe a few exceptions for initialization which don't apply here), which is the function call to ChatHandler::PSendSysMessage here. So while ```GetGuidStr()``` creates a temporary std::string, it should only be destroyed *after* the function call.

Also, if it would be incorrect, please avoid the extra string copy by using a const& to assign the string too. That extends the lifetime of the temporary as well. So for example don't write

```
std::string targetLootRecipientGuidString = target->GetLootRecipientGuid().GetString();
```

but use this instead:

```
const std::string& targetLootRecipientGuidString = target->GetLootRecipientGuid().GetString();
```

But again, I'm not convinced yet this is wrong in the first place.

killerwife:
Wouldnt doing a ref cause the same issue as what he is trying to fix? From my knowledge of the c++ standard the temporary dies immediately. I do not know how it works internally in the class of the fields though.

evil-at-wow:
Just a reference? Yes, that would be a problem. But not a reference to const as I suggested. C++ deliberately specifies that binding a temporary object to a reference to const on the stack lengthens the lifetime of the temporary to the lifetime of the reference itself. Note that you need to keep the string object itself alive, so doing something like this is also not a good idea:

```
const char* s = target->GetLootRecipientGuid().GetString().c_str();
```

At the end of the full expression, the temporary string object is deleted, making the pointer returned by c_str() a dangling pointer.

But as I mentioned before, I don't think we need to fix anything here because the temporary string object in the original code isn't deleted until after the function call. So I'd need to see strong evidence first of the contrary to be convinced.

killerwife:
https://stackoverflow.com/a/10006989/2320539 - lifetime of temporary ends at the end of the full expression. In this case, the line where we call GetString not on next {

evil-at-wow:
And in case someone doesn't want a random stack overflow answer, here's the reference from the  [the C++ 11 specification](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf) (or any other similar version for that matter) that says in 12.2 Temporary Objects:

> There are two contexts in which temporaries are destroyed at a different point than the end of the full-expression. The first context is when a default constructor is called to initialize an element of an array. If the constructor has one or more default arguments, the destruction of every temporary created in a default argument is sequenced before the construction of the next array element, if any.
> The second context is when a reference is bound to a temporary. The temporary to which the reference is bound or the temporary that is the complete object of a subobject to which the reference is bound persists for the lifetime of the reference except:
> * A temporary bound to a reference member in a constructor’s ctor-initializer (12.6.2) persists until the constructor exits.
> * A temporary bound to a reference parameter in a function call (5.2.2) persists until the completion of the full-expression containing the call.
> * The lifetime of a temporary bound to the returned value in a function return statement (6.6.3) is not extended; the temporary is destroyed at the end of the full-expression in the return statement.
> * A temporary bound to a reference in a new-initializer (5.3.4) persists until the completion of the full-expression containing the new-initializer.

So it's the second context above that makes the const& I suggest work. Note that using ```std::string&``` instead of ```const std::string&``` wouldn't even compile because the rvalue can't be bound to a non const reference (but std::string&& would work). And none of these two contexts apply to the initial code, so that's all good IMHO.

cyberium:
I had to modify GetGuidStr() on my old dead code about dynguid.
The code was a bit more complex so i decided to put the result in a variable instead of recomputing it at every call.
Cons : it add some bytes to WorldObject class. (usually something like 20-30 bytes)
Pro: its faster, it return a pointer that will have same lifetime of the object. We can make it return const char* directly so no need for more c_str() call.

For the topic itself yes i dont see the problem with this code :/ And the author did not describe it enough.

RyadaProductions:
As far as I have understood, `GetString()` returns by value, which would mark it to be destroyed at the end of the line if you don't put it in a variable to hold a reference to it. `c_str()` creates a pointer to the buffer of the actual string *if I have understood it properly*
So this means that when you call `GetString().c_str()` you make an assumption about the buffer still being there whenever you access the actual c_str pointer. The pointer doesn't lengthen the lifetime of the actual object as far as I am aware personally.

And yes it would have been better if i described it better with the PR :D

cyberium:
Ok thank for your explanation. But you didn't understood what @evil-at-wow said. Let me explain it in another way.
```cpp
PSendSysMessage("loot recipient: %s with raw data %s from group %u",
                        recipient->GetGuidStr().c_str(),
                        target->GetLootRecipientGuid().GetString().c_str(),
                        target->GetLootGroupRecipientId());
```
Take that line. You have a call in it that call GetString() -> ` target->GetLootRecipientGuid().GetString().c_str(),`
What you are saying is right if i do something like
```cpp
const char* str =  target->GetLootRecipientGuid().GetString().c_str();
printf("%s", str);
```
That code above is wrong and will result in undefined behavior because str is a dangling pointer in that case. String resulting from GetString() call will be deleted before the printf() call.

Now we'll modify it a bit
```cpp
printf("%s", target->GetLootRecipientGuid().GetString().c_str());
```
The code above will work. What the difference? Its considered as one instruction and the string will be deleted only at the end of the line (at `;`). So after the print() call.

So now if we look at our code again we can see it like this
```cpp
PSendSysMessage("loot recipient: %s with raw data %s from group %u", recipient->GetGuidStr().c_str(), target->GetLootRecipientGuid().GetString().c_str(), target->GetLootGroupRecipientId());
```
Yes one line of code too. Thats why it work.

To get the correct explanation (witch is not exactly what i said) reread the comment of @evil-at-wow .

Hope it help. Closing this.

evil-at-wow:
> As far as I have understood, GetString() returns by value, which would mark it to be destroyed at the end of the line if you don't put it in a variable to hold a reference to it.

Not the end of the line. Lines don't really matter in C/C++. You could write the code as

```
PSendSysMessage("loot recipient: %s with raw data %s from group %u",
                                 recipient
                                 ->GetGuidStr()
                                 .c_str(),
                                 target
                                 ->GetLootRecipientGuid()
                                 .GetString()
                                 .c_str(),
                                 target
                                 ->GetLootGroupRecipientId());
```
and still have the exact same behavior. Expressions are what counts, so instead of "destroyed at the end of the line" that should read "destroyed at the end of the full expression". Which is at the semi-colon (;) here, after the function call. This is why one can safely call a function that takes as parameter a reference bound to a temporary object, having the guarantee that the temporary object is still alive during the function call. 

>  c_str() creates a pointer to the buffer of the actual string if I have understood it properly
So this means that when you call GetString().c_str() you make an assumption about the buffer still being there whenever you access the actual c_str pointer. The pointer doesn't lengthen the lifetime of the actual object as far as I am aware personally.

That is correct, for the result of c_str() to be useful - meaning it's not a dangling pointer - the string object still needs to be alive and no other non-const functions may have been called on it since the c_str() call. Clearly no other functions are called on the temporary string object here, so that brings it down to the question wheter the temporary string object is destroyed or not. In this case, it gets destroyed after the function call, which means the pointer is still fine during the call.

RyadaProductions:
> Not the end of the line. Lines don't really matter in C/C++.

Line wasn't the correct term i agree, but with line i meant start of code line -> ; or }

That means that the general replacement on most lines is unneeded as it already behaves properly except on 1 specific place where its actually doing `const char* guidString = guid.GetString().c_str();` Which according to the explanation in multiple comments now is actually a faulty one that will have a dangling pointer.

I will update my PR to reflect this and then ask for it to be reopened.

evil-at-wow:
> except on 1 specific place where its actually doing const char* guidString = guid.GetString().c_str(); Which according to the explanation in multiple comments now is actually a faulty one that will have a dangling pointer.

Yes, very much! That one is valid, and I have to admit I missed that one in your PR.

PR #379, opened 2020-02-29
Fix temp weapon enchant proccing from bleeds

Yatzii93:
## 🍰 Pullrequest

Right now crippling poisen can be procced from bleed effects when crippling poisen is set on the main hand, not the off hand. Properly affects windfury as well or similar temp weapon enchants.

This change ensures the attack is from a melee weapon rather than a dot spell.
### Proof

image

### Issues

- Weapon enchant proccing when it shouldnt be

### How2Test

- add crippling posien to the main hand of your weapon, get it to proc on a mob and then put rupture on the mob. Dont attack and watch as the rupture will refresh crippling poisen.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Spells/UnitAuraProcHandler.cpp
+++ b/src/game/Spells/UnitAuraProcHandler.cpp
@@ -449,7 +449,7 @@ void Unit::ProcDamageAndSpell(ProcSystemArguments&& data)
     if (data.attacker)
     {
 		// trigger weapon enchants for weapon based spells; exclude spells that stop attack, because may break CC
-		if (data.attacker->GetTypeId() == TYPEID_PLAYER && (data.procExtra & (PROC_EX_NORMAL_HIT | PROC_EX_CRITICAL_HIT)) != 0 && (data.procFlagsAttacker == PROC_FLAG_SUCCESSFUL_MELEE_HIT || data.attType == OFF_ATTACK))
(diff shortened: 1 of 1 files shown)


Yatzii93:
Hold up this breaks offhand application, I will amend.

Yatzii93:
Ok seems to be all good now.

src/game/Spells/UnitAuraProcHandler.cpp
```diff
@@ -449,7 +449,7 @@ void Unit::ProcDamageAndSpell(ProcSystemArguments&& data)
     if (data.attacker)
     {
 		// trigger weapon enchants for weapon based spells; exclude spells that stop attack, because may break CC
-		if (data.attacker->GetTypeId() == TYPEID_PLAYER && (data.procExtra & (PROC_EX_NORMAL_HIT | PROC_EX_CRITICAL_HIT)) != 0)
+		if (data.attacker->GetTypeId() == TYPEID_PLAYER && (data.procExtra & (PROC_EX_NORMAL_HIT | PROC_EX_CRITICAL_HIT)) != 0 && (data.procFlagsAttacker == 4 || data.attType == OFF_ATTACK))
```
killerwife: Set enum instead of this random number.

Yatzii93: > 
> 
> Set enum instead of this random number.

Done.

Yatzii93:
Ok this doesnt quite get the intended result either, this stops chance of hit spell equips on weapons from proccing off melee moves. I think a check to see if the attack flags are from periodic damage is the way forward. PROC_FLAG_ON_DO_PERIODIC

Yatzii93:
This is working for me as intended with no apparent consequence now.

killerwife:
You almost got it right :D

PR #383, opened 2020-03-22
Allow playerbot.conf to be specified from command line. Removed obsolete usage functions. Updated realmd.conf to allow # listener threads

monomalvado:
## 🍰 Pullrequest
Updates to the mangosd and realmd executables.
* Removed obsolete Usage functions.
* Added ability to specify "playerbot.conf" from command line.

Updates to the realmd.conf
* Added "ListenerThreads" configuration variable.

Tested on Windows 10 with MSVC 142 and on *nix with GCC.


0x283A:
Thank you, I would really like this to be added.

PR #384, opened 2020-04-01
Fix typo in pdump causing SQL errors

henhouse:
## 🍰 Pullrequest
Seems at some point the string `NULL` was replaced with a C++ `nullptr` when this needed to remain as such for proper MySQL importing.

### Proof

- N/A

### Issues
- None

### How2Test
Use `.pdump write` on a character, and try to reimport it. You'll get SQL errors. Look inside the .sql file to see "nullptr" written as apart of the INSERT statement.

### Todo / Checklist

- [x] None


Warlockbugs:
What happened here likely is global find and replace for c++11.

PR #385, opened 2020-04-16
Chain heal : change the way we're getting units

Dromadoros:
## 🍰 Pullrequest
The chain heal never took the current target and it was healing more than 3 Units.

So I changed the way we get units around and remove an useless condition that check if the next target is maxhealth cause we already order units by their current health otherwise we can have 100 units with max health and chain will not take anyone.

### Proof
image

### Issues
- https://github.com/cmangos/issues/issues/2143

### How2Test
**Testing random units**:
- Go to friendly units
- Use chain heal 
- Be sure you have a pack of 3 units

**Testing player in group**:
- Select a player in group
- Use chain heal 
- Only members in group are in the chain

### DIFF
--- a/src/game/Spells/Spell.cpp
+++ b/src/game/Spells/Spell.cpp
@@ -1049,7 +1049,7 @@ void Spell::DoAllEffectOnTarget(TargetInfo* target)
         Unit::DealDamageMods(caster, damageInfo.target, damageInfo.damage, &damageInfo.absorb, SPELL_DIRECT_DAMAGE, m_spellInfo);
 
         // Send log damage message to client
-        
+
         if (reflectTarget)
             reflectTarget->SendSpellNonMeleeDamageLog(&damageInfo);
         else
@@ -2399,6 +2399,9 @@ void Spell::SetTargetMap(SpellEffectIndex effIndex, uint32 targetMode, bool targ
             if (!m_caster->CanAssistSpell(unitTarget, m_spellInfo))
                 break;
 
+            // if we have a target, always push it into final list
+            tempUnitList.push_back(unitTarget);
+
             if (targetingData.chainTargetCount[effIndex] <= 1)
                 tempUnitList.push_back(unitTarget);
             else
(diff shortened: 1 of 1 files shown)


src/game/Spells/Spell.cpp
```diff
@@ -2437,17 +2440,49 @@ void Spell::SetTargetMap(SpellEffectIndex effIndex, uint32 targetMode, bool targ
...
+                // No targets. No need to process.
+                if (tempAoeList.empty())
+                    break;
+
+                // Allways remove currentTarget cause we already have it into tempUnitList
                 tempAoeList.erase(std::remove(tempAoeList.begin(), tempAoeList.end(), unitTarget), tempAoeList.end());
-                if (!tempAoeList.empty())
-                    tempUnitList.splice(tempUnitList.end(), tempAoeList);
+
+                if (minRadiusCaster)
```
killerwife, line 2471: Chain heal has no min radius. The function fetcher has min radius because I added it there. The jump radius was configured based on patch notes (we are lucky to have jump radius specifically for chain heal from them). So this is entirely unnecessary and not how it works. The filtering is not done here, it is done in FilterTargetMap. So this block here is redundant.

src/game/Spells/Spell.cpp
```diff
@@ -2437,17 +2440,49 @@ void Spell::SetTargetMap(SpellEffectIndex effIndex, uint32 targetMode, bool targ
...
                 UnitList tempAoeList;
-                MaNGOS::AnyFriendlyOrGroupMemberUnitInUnitRangeCheck u_check(m_caster, group, m_spellInfo, max_range);
-                MaNGOS::UnitListSearcher<MaNGOS::AnyFriendlyOrGroupMemberUnitInUnitRangeCheck> searcher(tempAoeList, u_check);
-                Cell::VisitAllObjects(m_caster, searcher, max_range);
+                // Getting spell casting distance
+                float minRadiusCaster = 0, maxRadiusTarget = 0;
+                GetChainJumpRange(m_spellInfo, effIndex, minRadiusCaster, maxRadiusTarget);
+
+                // if we have a group, we don't care about other units around
+                if (group) {
```
killerwife, line 2452: Whats the research on this? On classic chain heal can jump to an assistable NPC, which is for example Escort NPC. This makes it completely not work.

killerwife: Also, AnyFriendlyOrGroupMemberUnitInUnitRangeCheck  does all of this for us.

PR #387, opened 2020-04-20
Naxxramas summoning spell fixes

Valkirie:
## 🍰 Pullrequest
Adressing known issues with Summoning Spells from Naxxramas using unused SpellEffect 41.

### Issues
Fixes the below encounters.
- Gothik the Harvester
- Gluth
- Maexxna
- Anub'Rekhan
- Sapphiron
- Kel'Thuzad


cala:
Thanks for this PR. :+1:
Do you have already tested the aforementioned encounters with these spells or do I have to have to go through them before merging?

Valkirie:
> Thanks for this PR. 👍
> Do you have already tested the aforementioned encounters with these spells or do I have to have to go through them before merging?

I'm on it. Kel'Thuzad and Gothik were tested. Others I just read through the script and breakpointed everywhere and it should be fine but I would be relieved if you could double check everything.

cala:
There is only Sapphiron Blizzard that is not working. The buffet spells are working however.
Other encounters seem fine. 👍

cala:
My mistake: the spell is working. That's the summon NPC that is probably missing its aura.
I noticed the (invisible) NPC is here when activating GM mode.
Merging this PR, then.

Thank you very much. 👍

PR #390, opened 2020-04-27
CombatManager to use ThreatRadius

Valkirie:
## 🍰 Pullrequest
CombatManager to use "unused" ThreatRadius instead of fixed 30.0f value.

### Issues
fixes an issue were creatures summoned further than 30.0f from target/owner would lead them to go back to their spawning location. [fixes #2184](https://github.com/cmangos/issues/issues/2184)

### How2Test
Start Apprentice Mirveda encounter (npc_apprentice_mirvedaAI) and check if creatures keep fighting the player and Mirveda (fixed) once they reach them.

### DIFF
--- a/src/game/Combat/CombatManager.cpp
+++ b/src/game/Combat/CombatManager.cpp
@@ -20,6 +20,7 @@
 
 #include "Entities/Pet.h"
 #include "Maps/Map.h"
+#include "World/World.h"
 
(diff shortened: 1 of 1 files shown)


killerwife:
Noted, not a bad solution as temporary bandaid, but the bug is twofold. Two NPCs when fighting should never evade, I believe apprentice mirveda is attacked by the mobs, not the player. Will fix soon-ish. Thank you for the PR however, it is the correct location for the fix :)

cyberium:
Fixed another way. Thank you for you effort.

PR #392, opened 2020-06-19
implement waypoint sanitizer

esno:
## 🍰 Pullrequest
this drops all invalid waypoints starting with 0
which is defined as the beginning where MovementGenerator
is heading to point 1

### Proof
- https://github.com/cmangos/issues/issues/2222#issuecomment-646337462

### Issues
- relates https://github.com/cmangos/issues/issues/2222

### How2Test
Head over to Arena in BRD and check whether High Justice Grimstone is following his defined waypoints.

https://github.com/cmangos/issues/issues/2222#issuecomment-646302322

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/AI/ScriptDevAI/system/system.cpp
+++ b/src/game/AI/ScriptDevAI/system/system.cpp
@@ -81,13 +81,22 @@ void SystemMgr::LoadScriptWaypoints()
     {
         BarGoLink bar(result->GetRowCount());
         uint32 nodeCount = 0;
+        std::set<uint32> blacklistWaypoints;
 
         do
         {
             bar.step();
             Field* fields = result->Fetch();
 
-            uint32 entry  = fields[0].GetUInt32();
+            uint32 entry = fields[0].GetUInt32();
+            uint32 pathId = fields[1].GetUInt32();
+            uint32 pointId = fields[2].GetUInt32();
+
+            if (pointId == 0)
+            {
+                blacklistWaypoints.insert((entry << 8) + pathId);
+                error_db_log("SD2: DB table `script_waypoint` has invalid point 0 for entry %u in path %u. Skipping.", entry, pathId);
+            }
 
             CreatureInfo const* info = GetCreatureTemplateStore(entry);
             if (!info)
@@ -96,8 +105,6 @@ void SystemMgr::LoadScriptWaypoints()
                 continue;
             }
 
-            uint32 pathId       = fields[1].GetUInt32();
-            uint32 pointId      = fields[2].GetUInt32();
             float position_x    = fields[3].GetFloat();
             float position_y    = fields[4].GetFloat();
             float position_z    = fields[5].GetFloat();
@@ -105,7 +112,9 @@ void SystemMgr::LoadScriptWaypoints()
             uint32 waitTime     = fields[7].GetUInt32();
             uint32 scriptId     = fields[8].GetUInt32();
 
-            sWaypointMgr.AddExternalNode(entry, pathId, pointId, position_x, position_y, position_z, orientation, waitTime, scriptId);
+            // sanitize waypoints
+            if (blacklistWaypoints.find((entry << 8) + pathId) == blacklistWaypoints.end())
+                sWaypointMgr.AddExternalNode(entry, pathId, pointId, position_x, position_y, position_z, orientation, waitTime, scriptId);
 
             ++nodeCount;
         }
--- a/src/game/MotionGenerators/WaypointManager.cpp
+++ b/src/game/MotionGenerators/WaypointManager.cpp
@@ -91,6 +91,7 @@ void WaypointManager::Load()
 
         // error after load, we check if creature guid corresponding to the path id has proper MovementType
         std::set<uint32> creatureNoMoveType;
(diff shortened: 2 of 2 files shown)


killerwife:
We WANT them to start at 0. If code is breaking because of it, then that code is wrong.

cyberium:
Nope we dont if the data start from 0 then data are wrong.
And i except for on new movegen that it seem it need it no we dont want them to start from 0.

Definitively not agree with that.

killerwife:
All old algorithms supported both 0 and 1. The "1" workflow was never mentioned anywhere, and overly complicates not only waypoint movegen but also spline system due to weird point calculations. When you start at 0 you can rely on normal arithmetics. Also, the old system didnt require to start at 0, it enabled starting at even 42. This is one of those old "mangos" things that make no sense that we need to get rid of.

If you have an argument for it, except "all current db data does it", this is probably the place to mention it. (all scripts reworked since 2017 by me have points from 0)

cyberium:
Then they are all wrongs. We clearly mention that somewhere that i can't get my hand on it but its pretty old.

Idea was at generator initialization it should have 0 as node id then it have to go to next node. So first node can have 42 yes no problem but not 0.

mangos zero remember it at least as its clearly [mentionned.](https://mangoszero-docs.readthedocs.io/en/latest/database/script/script-waypoint.html)

I see nothing that have to be changed here except new wrong data.

killerwife:
Okay, but why? There is no architectural reason in 2020 anymore to start from 0.

cyberium:
When we initialize the movement generator we had to choose a value for LastReachedNodeId. That value should be invalid. @Schmoozerd choosed 0 as with that meaning:
> We are somewhere and we need to reach first point in movement list. Thus 0 is not a real point is spawn point that can be anywhere and will not be reach again as last point will loop to first point not spawn point.

That's not exactly his word but its what its what i can remember. There no reason more than that we can rewrite another the waypoint gen to handle differently last reached point but its an overkill for this issue.

killerwife:
Okay. Now finally, after 4 years and 400+ commits I know something that should be on the wiki. @cyberium go write it there. @Warlockbugs please read it since it affects Path Movegen.

killerwife:
Also, this PR has a major problem. I am up for warning. But each one needs adjustments for scripts in core manually and same thing in DB. In many cases for both. Pushing this will break a ton of stuff.

esno:
@killerwife I don't think this PR will break anything which isn't already affected by https://github.com/cmangos/mangos-tbc/commit/c7f5d56c1f1c801674f3889d7f4744b0fa4034eb therefore I see two options: Drop waypoints starting with `0` and log them out to fix them one by one or revert https://github.com/cmangos/mangos-tbc/commit/c7f5d56c1f1c801674f3889d7f4744b0fa4034eb until related waypoints are cleaned up

killerwife:
Yes, that is why I did not merge the commit to our server. Sadly a solution like this is not acceptable to me :) This means that up to 50 scripts were broken. I don't mean that as an attack on your work btw, your PR seems to be in line with the workflow Cyberium is suggesting. I merely do not have the manpower to fix the 50 scripts myself right now.

cyberium:
They are already broken yes. "186" on TBCDB. Though some may be legit. @esno in the blacklist test we should check if the guid/entry use MovementType 2 too. So we permit 0 for PathMovementGenerator.

Majority of them will be easier to fix, some will need to take care of script.

cyberium:
@killerwife iam sure you can do it youself in a couple of hours. Too much easy for you maybe :D

esno:
> I don't mean that as an attack on your work btw

@killerwife no worries. discussions like that are exactly the reason why I raised a PR instead of pushing to master directly. Need to review scripts to verify if I could help fixing them.
@cyberium I'll have an eye on it while following up the work on this

src/game/AI/ScriptDevAI/system/system.cpp
```diff
@@ -81,13 +81,22 @@ void SystemMgr::LoadScriptWaypoints()
...
 
-            uint32 entry  = fields[0].GetUInt32();
+            uint32 entry = fields[0].GetUInt32();
+            uint32 pathId = fields[1].GetUInt32();
+            uint32 pointId = fields[2].GetUInt32();
+
+            if (pointId == 0)
+            {
+                blacklistWaypoints.insert((entry << 8) + pathId);
+                error_db_log("SD2: DB table `script_waypoint` has invalid point 0 for entry %u in path %u.", entry, pathId);
```
cala, line 98: As it seems we are going the route of sticking to "point index should not be zero", I think at least this error log should be added in core to help stupid DB devs like me to spot errors in their data. Thus, waypoints with first point having index=0 should be reported and skipped, like we usually do with invalid data at loading.

This is for silly DB devs to ensure their data are correct, not for the core to assume they are wrong and try to guess-fix them.

The main real issue here is that we have long-time unwritten (and as such forgotten) convention that led to the current situation. I have then written it in the wiki so at least, this one should not happen anymore.

cyberium:
@cala I asked @esno specifically to remove entire path if it start from 0. Better no movement than broken movement.
And ofc show error message :) Thank for adding it to wiki.

PR #394, opened 2020-06-20
Caster ai

miraco:
### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


killerwife:
Meleeing in caster mode is blizzlike. I verified it on retail. There is actually code that needs to change for it to be proper. But the commit in particular is not correct. Casters should still move into LOS and range, so disabling combat movement is not correct either. Correct distance needs to be set in AI (either SD2 or EAI)

PR #398, opened 2020-07-25
Update Lakota Windsong EscortAI

Crypticaz:
Added missing DoScriptText for when the escort ends
Added summoned creature attack on spawn

## 🍰 Pullrequest

Lakota Windsong (https://classic.wowhead.com/npc=10646/lakota-windsong) was missing the end text after finishing escort. 
Furthermore, bandits on spawn don't have a target, this should fix that and make it as close to blizzlike as possible.

### Proof

- How it should word -> https://www.youtube.com/watch?v=8QzqZycqYF8&feature=emb_title

### Issues

- None

### How2Test

- Accept quest Free at Last (4904) from Lakota Windsong in Thousand Needles
- Let escort play through along with the summons

### Todo / Checklist

- [X] None


killerwife:
Just a confirmation before I merge this. The script_text was already in the core sql but it was just missing from the script?

Crypticaz:
Yeah that is correct

PR #402, opened 2020-08-25
Send aura durations for self on login

henhouse:
## 🍰 Pullrequest
The Lua function [UnitBuff()](https://wow.gamepedia.com/API_UnitBuff) returns information regarding an aura in a specific slot. One of the return values is the aura's duration. However, on login and map change these values seem to be cleared from the client making this Lua call not work properly. The result is breaking some AddOns like [ProcIndicator](https://github.com/Oldsalt0/ProcIndicator) until the buff is refreshed and the client has a new duration to track.

### Proof
Uncertain if this was how it really worked in TBC, but seeing as TrinityCore sends self-aura durations on client login in the same area, I don't see why this would harm anything, and since you should assume Lua APIs work, this seems like an appropriate fix.

### Issues
- None tracked

### How2Test
- Use the [addon](https://github.com/Oldsalt0/ProcIndicator) mentioned above, or dump the value of `UnitBuff("player",1)` using an addon like [DevTools](https://www.wowinterface.com/downloads/info3999-DevTools.html) when having only one aura, such as Blessing of Might. Relog, and notice that the dump no longer includes information such as the duration remaining.
- Use the branch in the PR and repeat the step above. You will receive aura duration information on login.

Difference between the two:
image

### Todo / Checklist
- [x] None

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -19360,6 +19360,7 @@ void Player::SendInitialPacketsAfterAddToMap()
             auraList.front()->ApplyModifier(true, true);
     }
 
+    SendAuraDurationsForTarget(this);
     SendEnchantmentDurations();                             // must be after add to map
     SendItemDurations();                                    // must be after add to map
 }


killerwife:
I will delve into sniffs from tbc and validate this. I think I saw something related to this already :)

killerwife:
Okay, I am back, armed with knowledge.

`      case SMSG_INIT_EXTRA_AURA_INFO_OBSOLETE:
        sub_60B890(v5);
        while ( !(*((int (__thiscall **)(CDataStore *))msg->vTable + 5))(msg) )
        {
          CDataStore::GetUint8(msg, &slot);
          CDataStore::GetUint32((int)&spellId);
          CDataStore::GetUint32((int)&maxDuration);
          CDataStore::GetUint32((int)&duration);
          if ( maxDuration )
          {
            v7 = duration + clientUptimeMs;
            if ( !(duration + clientUptimeMs) )
              v7 = 1;
          }
          else
          {
            v7 = 0;
          }
          sub_612E30(v6, st7_0, slot, spellId, maxDuration, v7);
        }
        eventGuid = guid;
        FrameScript_SignalUnitEvent(&eventGuid, EVENT_UNIT_AURA);
        result = 1;
        break;
`

This is the unused packet that signals a specific aura event on load. In sniffs it arrives empty with just a guid in 2.4.3 even though the packet structure enables it. So we need a new function which sends SMSG_INIT_EXTRA_AURA_INFO with just caster guid in it.

Next, protocol order. We do a couple things wrong, blizzard iterates over all items first and sends enchantments and durations mixed together (if you want to fix it yourself, would be appreciated). 

The sending itself in TBC is this:

ServerToClient: SMSG_INIT_EXTRA_AURA_INFO_OBSOLETE (0x03A3) Length: 43 ConnectionIndex: 0 Time: 06/30/2008 08:36:31.000 Number: 356
GUID: Full: 0x003CDE2E UInt64: 3989038 Guid: Muuggener

ServerToClient: SMSG_SET_EXTRA_AURA_INFO_OBSOLETE (0x03A4) Length: 17 ConnectionIndex: 0 Time: 06/30/2008 08:36:31.000 Number: 357
GUID: Full: 0x003CDE2E UInt64: 3989038 Guid: Muuggener
Aura Slot: 56
Aura ID: 16092
Aura Max Duration: 4294967295
Aura Duration: 0

It sends a list of all nondisplayed auras which are passive. Max slots is 55. This is done right after sending 836 spell ID which is the login animation. The rest of the durations for slots 0 to 55 are send beforehand, they can be mixed together with modifier packets.

In summary, the order needs to look smth like this:

Aura modifier packets
Duration packets using SMSG_SET_EXTRA_AURA_INFO  for slots 0 to 55
Item enchantments and durations
Logineffect spell 836
SMSG_INIT_EXTRA_AURA_INFO_OBSOLETE with only player guid
SMSG_SET_EXTRA_AURA_INFO  for slots 56 and above

The mechanism for slots 56 and above isnt done in our core yet, its done in void SpellAuraHolder::_AddSpellAuraHolder() but I can do that one myself. Apply all possible notes from this if you can and the more you do the sooner I will have time to merge it.

PR #403, opened 2020-08-28
do not push next waypoint, when path is not initialized

rnorden:
## 🍰 Pullrequest
Gracefully handle missing waypoints/paths by ignoring them.

### Proof
The Crone spawns her cyclones and the servers continues to run.

### Issues
resolves cmangos/mangos-tbc#2245

### Todo / Checklist
- Add waypoints for the cyclones
- Are there any other underlying problems with uninitialized paths? Why does this code segment run with the cyclones? Why not with others?
- Do **not** add a log message; there are log messages for missing waypoints/paths already.

### DIFF
--- a/src/game/MotionGenerators/WaypointMovementGenerator.cpp
+++ b/src/game/MotionGenerators/WaypointMovementGenerator.cpp
@@ -378,6 +378,10 @@ static const uint32 PreSendTime     = 1500;
 // build and send path to next node
 void WaypointMovementGenerator<Creature>::SendNextWayPointPath(Creature& creature)
 {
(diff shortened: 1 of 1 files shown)


berlingm (commented):
Could be a good idea to make a warning log, as I have seen missing waypoints other places as well. Such as Gargoyles in Hyjal after Alliance fase completion.

berlingm (approved):
I have verified that this fixes the waypoint issue reported, including issues found in Hyjal instance of the same type.

Lillecarl:
This works well when I'm testing.

PR #404, opened 2020-09-15
Prevent the console error "MoveSpline::init_spline: zero length spline".

Olion17:
## 🍰 Pullrequest
For a creature able to flying the following execution pattern was possible:
```C++
Creature::ForcedDespawn(0);
Creature::SetDeathState(JUST_DIED);
    Unit::SetDeathState(JUST_DIED);     // here movement is stopped by:
        StopMoving();
        i_motionMaster.Clear(false, true);
        i_motionMaster.MoveIdle();
     i_motionMaster.MoveFall();         // here movement is reinited to MoveFall() and the error printed
    Unit::SetDeathState(CORPSE);
RemoveCorpse(true);      // at the same tick, prevents any movement animation (just disappearing)
```
There is no way to skip setting `JUST_DIED`, no need to reinit movemet thereafter, and no way to know atm that the `RemoveCorpse()` will follow at the same tick. It must be considered as a design flaw. The proposed fix is far from elegant one, but I'm short of ideas. Either way we need a movement setting control inside of `Creature::SetDeathState(JUST_DIED)`.

### Proof
Not needed.

### Issues
- fixes (partially?) cmangos/issues#2255

### How2Test
Using the 96dda3c3, spend some time at, for example, Light's Hope Chapel, to get the error logged. Retry since this applied. An extensive testing is needed to find another causes of the error, if any.

### Todo / Checklist
When a creature flying at the ground level falls dead, the same error might arise (though I've failed to get the error since this applied). However `Creature::JustDied()` looks to me as a wrong place to compare z-coordinate and the ground level before movement reinit.
The issue must be the same for (all?) cores.

### DIFF
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -1683,7 +1683,7 @@ void Creature::DeleteFromDB(uint32 lowguid, CreatureData const* data)
     WorldDatabase.CommitTransaction();
 }
 
-void Creature::SetDeathState(DeathState s)
+void Creature::SetDeathState(DeathState s, bool falling)
 {
     if ((s == JUST_DIED && !m_isDeadByDefault) || (s == JUST_ALIVED && m_isDeadByDefault))
     {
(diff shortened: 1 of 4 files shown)


craftorix:
Works with WOTLK too! Error is now gone.

killerwife:
My question is why MoveFall produces this error to begin with. Seems like wrong filling.

Olion17:
> My question is why MoveFall produces this error to begin with. Seems like wrong filling.

A good question but irrelevant to this issue, as I (hopefully clearly) explained above. There is no need to init any motion right before despawn at this tick.

In the debugged case, the MoveFall() generator got finishing point above starting one, i.e. was instructed to generate falling up. So I present here a seemingly cleaner solution actually solving nothing, opening the way for further debugging instead. I saw the issue at a flying NPC despawn, for example, NPC 23188 in Shadowmoon. I could speculate that a rough motion stopping before despawn fails to update the NPC serverside coordinates correcty. It doesn't explain why we get way too high ground Z though.

PR #406, opened 2020-10-24
BG commands and fix for no-winner games

henhouse:
## 🍰 Pullrequest
Ports over BG start/stop commands from vmangos into its own sub-command menu instead of the existing under debug.
Also fixes issues where handling of a BG ending in a draw didn't exist, either by the new stop command or premature timer due to not enough players.

The rationale for moving the command to its own submenu is because it doesn't really belong under debug, especially if more commands should be added (such as the `.bg info` command from vmangos too to gather data on queues and in-progress BGs), and other useful GM-related tooling.

### Proof

- None

### Issues

- None

### How2Test

- Enter a BG and force start it via `.bg start` and end it shortly after with `.bg stop`. The game should end with no winner.

### Todo / Checklist

- [X] None


Lillecarl:
Looks good to me. If anything this should (probably) be 2 separate PR's. But it's short and easy to review anyways.

PR #407, opened 2020-10-25
Update the build requirements to CMake 3.1 and C++ 14 compiler support.

evil-at-wow:
Notes:
* CMake 3.1 was released in December 2014, so this seems like a very modest requirement these days.
* We started requiring C++ 11 support in compilers in October 2015, so requiring C++ 14 support about five years later also seems like a fair requirement. Practically speaking, this means you'll need at the very least Visual Studio 2015, GCC 5 or Clang 3.4 (or equivalent in Xcode on macOS).

Let's see what our CI thinks about this. As far as I can see we don't need any adjustments on our CI setup, because we're already using recent systems and compilers (GCC 7, Clang 7, VS 2019)...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@
 # Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 # Minimum required cmake version to run the build
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.1)
 
 # Will cache object files if ccache is available
 # Must be set at top of cmake outside of project header
@@ -26,6 +26,11 @@ if(CCACHE_PROGRAM)
 endif()
 
 project(CMaNGOS_TBC)
+
+# Require C++ 14 support
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 # Define here name of the binaries
 set(CMANGOS_BINARY_SERVER_NAME "mangosd")
--- a/cmake/compiler/clang/settings.cmake
+++ b/cmake/compiler/clang/settings.cmake
@@ -1,21 +1,6 @@
 # Set build-directive (used in core to tell which buildtype we used)
 add_definitions(-D_BUILD_DIRECTIVE='"$(CONFIGURATION)"')
 
-# Check C++11 compiler support
-include(CheckCXXCompilerFlag)
-CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX14)
-CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
-CHECK_CXX_COMPILER_FLAG("-std=c++0x" COMPILER_SUPPORTS_CXX0X)
-if(COMPILER_SUPPORTS_CXX14)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
-elseif(COMPILER_SUPPORTS_CXX11)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-elseif(COMPILER_SUPPORTS_CXX0X)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
-else()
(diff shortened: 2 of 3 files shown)


killerwife:
I would go as far as C++17 immediately. @cyberium  and @cala you two please let us know. I don't know if @cala  can compile c++17 at this point and he can, then its a nobrainer. I already talked to the old man about this on discord.

evil-at-wow:
Let's first look at it from a timing perspective, because we've always been somewhat conservative in this project. Considering we decided to switch to C++ 11 only in October 2015, and C++ standards are on a three-year cadence, we should probably have switched to C++ 14 at the end of 2018 and switch to C++ 17 near the end of 2021 (so next year). So timing alone makes this PR and C++ 14 a very reasonable requirement that probably no one will challenge.

Then there's C++ 17. Changing the 14 into 17 is obviously a trival thing to do, so that's not the problem. The first question is what C++ 17 means in this context: are we looking at full support, or do we want to use a subset of its features for now? So we have to ask ourselves what C++ 17 will bring us, and what we'll have to sacrifice for it. Requiring full support means making a lot of sacrifices. I would say perhaps too much. For example, Apple only implements parts of the required runtime support of std::optional and std::variant in macOS 10.14 and later, so anyone with an older macOS is out of luck, and somewhat older hardware simply can't run macOS 10.14. Sure, that's often hardware from the ~2012 era which is more than eight years old by now, but still you need an OS that's not older than like just 3 years. You'll also need at the very least a GCC 7. I think Ubuntu only started shipping with GCC 7 in Ubuntu 18.04, so anything older than that is likely out of the question too. SuSE versions starting using GCC 7 in SuSE Linux 15 I believe. That means you'll need pretty recent Linux distributions. I still have a GCC 6.2 on my four year old workstation, so that's not going to work either. 

But somewhat older compilers usually do support of subset of C++ 17 though, like my GCC 6.2. So maybe we should starting listing what C++ 17 features we want to use 'now', and perhaps consider using only that a subset of C++ 17 for now?

cala:
Someone summoned me?

I honestly don't really have an opinion about which standard version of C++ we should support. My main points on the matter would be :
- do we really need this or is this for convenience only (*i.e.* are there things we can not do/fix without upgrading) ?
- are the benefits/issues removal bigger than the troubles the transition will raise ?
- does the upgrade force us to drop some support or content or users ?

But these questions are true for any upgrade we can do, like the MySQL8.0 one for example.

As for myself, I've build the core for more than one year now on macOS 10.14 with Apple provided Clang and some extra dependencies installed via HomeBrew, like OpenSSL (and possibly more, but I did the mistake of not writing a guide of what I did, so I can remember for sure). I don't have the exact version and C++ standard support of said Clang in mind, but I can check if that helps finding if compliance to C++14 or C++17 applies.

killerwife:
That was indeed the point for my summons :) C++14 is the convenience patch, C++17 is the one that has nice stuff we can replace dependencies with.

evil-at-wow:
To answer @cala's questions:
* We don't really need C++17. Great software was written without it as well. But it has some nice things that make life a bit easier I guess (but only for the developers, not the users).
* A transition to C++17 should not introduce trouble for anyone, except if you're using a system or compiler that will no longer be supported. For that reason, my personal preference is that we're not going "full C++17" just now, because that basically means that any compiler older than say 2 years is simply too old, and thus many Linux distributions will be out of luck too. I'm sure many people will disagree with me, but I don't think ditching a 3 year old system is reasonable, especially for an educational project like this. Adopting some C++17 features seems more doable, but sadly the most interesting and handy ones seem to be the ones needed the most up to date compilers...

> C++17 is the one that has nice stuff we can replace dependencies with.

I'm not challenging that C++17 has some nice stuff, but can you tell what depenencies you have in mind please? I assume you have some boost things in mind, but given we're also using it for ASIO we're not likely to get rid of that soon.

cala:
Do you need me to check versions on my config or run C++14 or C++17 code samples to test if this work?

evil-at-wow:
No, I pretty sure macOS 10.14 has the runtime support for std::optional and std::variant, and on macOS 10.14 you can run Xcode 11, so pretty much everything in C++17 should work fine for you.

evil-at-wow:
For your information: it is my intention to push this commit this week (Wednesday or Thursday), especially after seeing new build failures due to C++ 14 being used in the most recent exports. By then people got 10 days to look at this PR and respond, which I think is a fair period of time. As far as I'm concerned, the C++ 17 discussion can continue for a bit though.

Lillecarl:
https://en.cppreference.com/w/cpp/compiler_support#cpp17 for those wondering about what's supported where.

PR #410, opened 2020-11-03
Convert item_instance blob with SQL query (s2420_01_characters_item_instance_data_drop.sql)

Lymphome:
Convert the blob to fields with a SQL query.

It is based on https://github.com/eilwin/TrinityCore/blob/master/sql/gc/mangos_converter.sql.

Tested on mariadb 15.1 (debian 10). Must be tested on big tables (mine has only 15k rows).

It is quite slow : about 2sec against 20ms with a perl script, but it's in the update process without to use an extra language.
Quite easy to modify for vanilla or wotlk.

(I don't really know if PR is the good way to submit that kind of changes)


MantisLord:
This looks good, we can amend the update as you suggest and leave the Go script as a backup.

Can you make a wotlk/classic version of this so I can test? You mentioned it's easy to modify for those two.

Lymphome:
I will do a PR for the other cores, but I don't have any to test.

PR #412, opened 2020-11-07
Battleground, Warsong : decrease drop timer, correct place for score update

Lymphome:
Fix two problems in Warsong :

* When flag is dropped, it doesn't disappear and respawn at base after 10sec
  m_flagsDropTimer is not decreased in Update()

* When player captures a flag, score is not updated accordingly in the client
  the score is sent to the client before the update

Also, fix a test which should obviously use m_flagsTimer.

Tested on last core revision (s2420).


MantisLord:
I fixed the flag return after drop timer earlier today but failed to notice the score problem, good catch.

PR #414, opened 2020-11-17
A try for large object visibility

Lymphome:
## 🍰 Pullrequest
A try to have LARGE objects displayed

### Issues
1. Large objects are not displayed at the good distance.
2. Players going far away from an another player stay displayed and not are updated for other clients if they are in a distance > ~350 on map with a large distance of view.

### How2Test
Put a Large object on a map, look the distance when it is displayed while you are approaching.

### Todo / Checklist
I needed to change the value in CellImpl.cpp (not only camera) because players weren't correctly updated when getting out farther than ~350 (on maps like BG with a big dist of view). Trinity sets this value to MAX_GRID_SIZE.

Can impact server load because more cells are explored and more objects are processed.

(scuse my bad english... and still learning how to use github)


killerwife:
This would completely murder performance on a big server. The more desired strategy would be to create a tiered list of npcs, based on distance. Normal ones would be like players, and the higher the npc, the less frequently would it be done. But I am always up for discussion on this.

Lymphome:
Maybe use a flag to allow the core to choose a different distances. I'm still learning (trying to understand) how maps and cells are managed in the core. About server load, I can't do tests with thousand of players but it's obvious it's a big work for the server. Anyhow, trinity has a bigger distance in cellimpl than cmangos. I will try to find a better way to accomplish that.

killerwife:
The information is already in the core. GOs have IsLargeGameObject function and creatures have it inside creature_template_visibility, where it was added from datamined sources. Its all available inside ObjectVisibility class per creature. Problem is that we should not use the gigantic range for all mobs, just for the gigantic mobs. Thats the same problem our and TCs system have, it always only anticipated one size.

cyberium:
Yes this is much complex than what it is in first look and we can't accept this at this stage.

PR #416, opened 2020-11-19
Adding JustDisorientedTarget for creatures that use spells like scatt…

miraco:
…er shot to distance their self after.

## 🍰 Pullrequest
Creatures that use "confusing" spells like [Scatter Shot](https://www.wowhead.com/spell=23601/scatter-shot) can now distance their self if CastFlag - CAST_DISTANCE_YOURSELF is given.

For example [Shattered hand sharpshooter](https://www.wowhead.com/npc=16704/shattered-hand-sharpshooter#abilities;mode:normal)


killerwife:
Will verify with retail if this is legit.

killerwife:
We looked on retail. Scatter shot like abilities do not cause distancing but cause suppression, different effect. I have a video of it, but I do not post videos here because our testing youtube acc got already banned once because of it. I can send it on discord if you wish. closing this.

PR #417, opened 2020-11-20
add send ai to master function

miraco:
## 🍰 Pullrequest

Currently we are able to send special AI events to nearby creatures. With this function we can send these AIs to master NPCs from creature linking.
Why do we need this?
In Shattered Halls every group with a legionnair has a different way to handle the death of a friendly creature. 
### Proof

atm every creature sends AI Event B on Death to nearby legionnair. So, if you pull Mobs from the first group to the 2nd group and kill them in the range of the 2nd legionnair, he will get this AI Event and will try to spawn a new mob/get enrage. 

Thats wrong like you can see in this video:
https://www.youtube.com/watch?v=iA75EO3UGpM
### Issues

Didnt had any issue at any group i currently scripted local.

### How2Test

Cant realy test this atm cause my legionnair script didnt got uploaded yet.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -1345,6 +1345,13 @@ bool CreatureEventAI::ProcessAction(CreatureEventAI_Action const& action, uint32
             SetRootSelf(action.immobilizedState.apply, action.immobilizedState.combatOnly);
             break;
         }
+        case ACTION_T_SEND_AI_TO_MASTER:
+        {
+            if (Unit* target = m_creature->GetMap()->GetCreatureLinkingHolder()->GetMaster(m_creature))
+                if (target->IsAlive())
+                    m_creature->AI()->SendAIEvent(AIEventType(action.sendEvent.eventType), m_creature, target);
+            break;
+        }
         default:
             sLog.outError("%s::ProcessAction(): action(%u) not implemented", GetAIName().data(), static_cast<uint32>(action.type));
             return false;
--- a/src/game/AI/EventAI/CreatureEventAI.h
+++ b/src/game/AI/EventAI/CreatureEventAI.h
@@ -142,6 +142,7 @@ enum EventAI_ActionType
     ACTION_T_SET_FACING                 = 59,               // Target, 0 - set, 1 - reset
     ACTION_T_SET_SPELL_SET              = 60,               // SetId
     ACTION_T_SET_IMMOBILIZED_STATE      = 61,               // state (true - rooted), combatonly (true - autoremoved on combat stop)
+    ACTION_T_SEND_AI_TO_MASTER          = 62,               // Sending custom AI event to linking master
 
     ACTION_T_END,
 };
@@ -549,6 +550,11 @@ struct CreatureEventAI_Action
             uint32 apply;
             uint32 combatOnly;
         } immobilizedState;
+        // ACTION_T_SEND_AI_TO_MASTER
+        struct
+        {
+            uint32 eventType;
+        } sendEvent;
         // RAW
         struct
         {
--- a/src/game/AI/EventAI/CreatureEventAIMgr.cpp
+++ b/src/game/AI/EventAI/CreatureEventAIMgr.cpp
@@ -1083,6 +1083,13 @@ void CreatureEventAIMgr::LoadCreatureEventAI_Scripts()
                         break;
                     case ACTION_T_SET_IMMOBILIZED_STATE:
                         break;
+                    case ACTION_T_SEND_AI_TO_MASTER:
+                        if (action.sendEvent.eventType >= MAXIMAL_AI_EVENT_EVENTAI)
+                        {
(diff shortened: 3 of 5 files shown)


miraco:
Just to show you an example of how my legionnair script currently works:

ss+(2020-11-20+at+01 54 40)
Checking what guid we have on Spawn

ss+(2020-11-20+at+01 56 01)
Received AIEvent from a group member.

ss+(2020-11-20+at+01 56 58)
Spawning the new Mob based on what guid we have.

`('1652306','16523','6','0','100','0','0','0','0','0','0','0','62','6','00','0','0','0','0','0','0','0','0','0','Shattered Hand Savage - Send AI Event B on Death'),`

And this is a example of the EventAI from a friendly mob.

miraco:
I have some free time next week and wanna continue on SH.
So a feedback on this would be great because my entire script is mostly based on this. 
@killerwife

killerwife:
Thx for pinging me again, I did not want to particularly ignore this. So the bone to pick with this, is that currently EAI and others are intentionally linking agnostic. Currently we are trying to develop something called Creature Groups, which do not have the notion of Master (yes, everything that is currently possible, will be possible and no work on Linking will be lost I 100% guarantee that) however that poses a problem implementation wise with this.

Basically from what you are saying, is that a particular group MUST send AI event to particular legionnaire. Why dont you launch a dbscript on death which by guid search targets a particular legionnaire? That way we can stay linking agnostic in EAI and AI in general.

miraco:
Yeah i agree that our current implementation with creature groups are not good. Mostly the way groups move (atleast imo) 

> Basically from what you are saying, is that a particular group MUST send AI event to particular legionnaire. Why dont you launch a dbscript on death which by guid search targets a particular legionnaire? That way we can stay linking agnostic in EAI and AI in general.

Yes, you are right. Didnt thought about making it with using guid.Gonna try it.

killerwife:
Do let me know, I am already a hindrance as it is for many PRs due to how stupidly I am behind on exports. Def want to help you find a sane solution to this, just I would rather not add linking related solutions right now.

miraco:
Hm yeah, i just started to play around with this. But i am not sure how to handle it.
` // ACTION_T_SEND_AI_TO_GUID
        struct
        {            
            uint32 eventType
            uint32 guid;
        } send_ai_guid;`

So the table structure would be something like: ActionType=  63, Actionaram = EVENT_TYPE, ActionParam2 = GUID, (maybe can add an alive check too so) Actionparam3 = 1/2 (true/false alive)

`case ACTION_T_SEND_AI_TO_GUID:
        {
            Creature* pReciver = nullptr;
            pReciver = m_creature->GetMap()->GetCreature(action.send_ai_guid.guid);
            
            m_creature->AI()->SendAIEvent(AIEventType(action.sendEvent.eventType), m_creature, pReciver);
            break;
        }`
        
i just dont know how to get the uin32 into an ObjectGuid, or is there another function to get CreatureWithGUID in range?

killerwife:
Well, I will suggest a heretical approach, but it will make your life 100% easier. On Receive AI Event check GetDbGuid() against the ones set in DB for this script. Since we adopted instance files this should no longer be an issue.

miraco:
I thought about it again, and imo the best way still is the way i did it using groups. 
I think you might not understand me right with how this instance works. I will try to explain it with screens etc.

This instance have multiple groups with 1 Legionnair as their leader. Everytime a creature of this groups die, the legionnair will notice and do special events.

First Group:
Group 1
On the first group its kinda simple, and the way we implemented is in my opionion the easiest way. When killing NPC **A, B, C, D** or **E** and having the legionnaire **L1** alive, the legionnaire gets the Buff enrage and the dead npc just instant respawn behind and run to their position.
The easiest way to implement the respawn was is setting spawntimesecsmin and max to 5 second and prevent respawning when legionnair is dead with FLAG_CANT_SPAWN_IF_BOSS_DEAD.

Second Group
Group 2
The 2nd group is already a bit more complexe.
When NPC **F**, **G**, **H**, **I**, **J**, **K**, **M** or **N** dies an legionnaire **L2** is alive, legionnaire gets buff enrage, and spawns a FelOrc (id 17083) on the red point. The Fel orc moves to the other red point and gets infight with the group.

A member of Group 1 cant trigger the event at Legionnair from Group 2. 

So even with using ACTION_T_SEND_AI_TO_GUID wouldnt work.
i could check what npc throws the Event but this would already look something like this for the 2nd Group (not using hc guids)
code

Currently the npcs use ACTION_T_THROW_AI_EVENT on death. With this every legionnair in range would get trigger his special event.

The best way to prevent this is in my opionion using the group system like i did with this commit.  Or creature_ai based on guid not on entry but this would result in 10k database lines.

Hope its understandable now.

cyberium:
Thank you very much for this. However a new grouping method is making its way to the core and it will be way much easier than this to handle this case. This will not be right to continue to extend creaturegroup linking now while we are going to remove it.

miraco:
Okay nice to hear.
Looking forward to the changes. Do you have an eta for this?
Does this include creature group moving?

cyberium:
PR should come in next few week with the new base on it.
Yes Formation will be possible for groups.

PR #419, opened 2020-12-18
Transports: pet looks real when moved to Transports

cocely:
### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


killerwife:
This is not blizzlike. Pets teleport onto transports on retail. Be it classic, old tbc or current BFA.

PR #422, opened 2021-01-12
Allow for the .mod speed commands to go up to 100 instead of 10

Numidiaa:
Modern computers can definitely handle more than what was originally allowed by this command and the original limit was way too low for most uses.

## 🍰 Pullrequest

Changes the maximum allowed values of .modify aspeed , .modify speed, .modify bwalk, .modify swim from 10 to 100.

### Proof

- None

### Issues

- Fixes how slow the mod speed commands allowed the player to go to originally.

### How2Test

- Go in game and do .modify aspeed 100

### Todo / Checklist

- [X] None


cyberium:
Do team really think that is needed? 10 is already a lot to handle when you just want to move from one place to another. Teleportation is easy to use enough for other cases on my opinion.

I have no issue to implement it in sligtly different way tho if you feel its needed

cyberium:
Look like not much of you are interested to this one. Closing this for now if you are not agree please say it loudly enough ;)

Thank you very much for your contribution anyway.

PR #431, opened 2021-02-25
Shadowfang Keep: Fix Arugal Voidwalkers

insunaa:
## 🍰 Pullrequest

If, after killing Fenrus in Shadowfang Keep you wiped or in some other way got out of combat with the voidwalkers, they would start pathing down the staircase and at some point they would start moving straight out of the map to undefined coordinates.

### Proof

- None

### Issues

- fixes https://github.com/cmangos/issues/issues/1204
### How2Test

(Copied from the original issue)
- .go creature fenrus
- Kill the dog
- Don't engage the 4 voidwalkers, just follow them as they start to patrol
- See some of them getting lost, especially at the bottom of the stairs

### Todo / Checklist

- [ ] Maybe fix them going invisible sometimes as they walk down the stairs? I believe this is caused by them clipping beneath the stairs for a few pixels, so the game culls their visibility and nameplate. No idea how to, tho.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/shadowfang_keep/shadowfang_keep.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/shadowfang_keep/shadowfang_keep.cpp
@@ -216,11 +216,12 @@ struct mob_arugal_voidwalkerAI : public ScriptedAI
         m_bWPDone = true;
 
         Creature* pLeader = m_creature->GetMap()->GetCreature(m_leaderGuid);
-        if (pLeader && pLeader->IsAlive())
+        if (pLeader && pLeader->IsAlive() && m_creature != pLeader)
         {
             m_creature->GetMotionMaster()->MoveFollow(pLeader, 1.0f, M_PI / 2 * m_uiPosition);
         }
-        else
+
+        if (pLeader && !pLeader->IsAlive())
         {
             CreatureList lVoidwalkerList;
             Creature* pNewLeader = nullptr;


insunaa:
Short explanation for why the issue occurred and how the PR fixes this issue:
The issue occurred, because 4 voidwalkers were spawned in a `for` loop, but they didn't necessarily all enter the world at the same time, because asynchronicity or something. The `Reset()` method is built to have the waypoint pathing+formation walk robust to the leading voidwalker dying, so upon reset (and thus also upon spawn) it would check if 1) the leader exists, 2) the leader is alive. Normally this is good practice to check if a creature exists before calling its methods, but in this case it's realistic that the leader didn't exist in the world yet, as the leader was being retreived by `Creature* pLeader = m_creature->GetMap()->GetCreature(m_leaderGuid);`, which meant a creature which was not the leader would start re-assigning the leader role (to itself), but after a while the actual leader would spawn and also have the leader role.

The creature with the leader role is the only creature allowed to update the waypoint state of the other voidwalkers, but because there were not two voidwalkers with the leader role, either of them could reach a waypoint and broadcast the waypoint update, which meant at the last waypoint, both of them could (and did) increment the waypoint counter past the highest entry in the waypoint array, basically causing one of the leaders to read a waypoint value from outside of the waypoint array, so basically invalid memory, and off they would go to infinity.

tl;dr: Issue was caused by a race condition that caused a read outside of allocated memory.

My fix addresses this by only actually searching for a new leader if the old leader is confirmed as `!isAlive()` and not in an else clause which might also just mean leader is a nullptr

cyberium:
Warning, I dont like that script. It clearly need some rework.

That being said are you sure the issue will not come back if we kill the leader?
The TODO part can be ignored as its not the current concern of this issue and its a general issue.
Please confirm me you consider it fixed. So only port to other cores is still needed for this to be merged.

insunaa:
I have tested it by successively killing the current leader and letting them re-organize, then wait to complete 2 cycles until I killed the next leader, until all of them were dead. I agree that the script probably needs rework, but I honestly don't know where to begin.

But this does definitely fix the issue at hand.

cala:
I'll look at this PR. And also see if the (initial) script needs to be reworked.

cala:
Ok, I just had a look at `shadowfang_keep.cpp` and almost every script in here is outdated in regard of all changes and improvements that were made in core lately. I shall rework the whole file.

insunaa:
If I can help feel free to assign some parts of the re-work to me, but be aware that you'll probably have to re-work my re-work as well 😬

cala:
@insunaa I'm fine with reworking your rework! 😄 

CMaNGOS is an opensource and educational project, so basically this is what we do to some point when reviewing PRs and suggesting/amending changes.

I will now rewrite `npc_shadowfang_prisonerAI` for the other open PR.

Feel free to update another part of the file if you want. I would suggest that you start with `npc_arugalAI` which basically is a timed speech with no combat. You can look at https://github.com/cmangos/mangos-classic/blob/0d7df3c404fc729f2ba1a361a3d71ec0d0d6ae35/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/blackrock_spire/instance_blackrock_spire.cpp#L561 or https://github.com/cmangos/mangos-classic/blob/0d7df3c404fc729f2ba1a361a3d71ec0d0d6ae35/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/naxxramas/instance_naxxramas.cpp#L76 to get examples of objects to handle scripted dialogs in instances.

PR #432, opened 2021-03-02
Correct Faerlina's outer door

insunaa:
## 🍰 Pullrequest

Previously there was no outer door from Faerlina to the outer ring of Naxxramas and it also had the wrong Gameobject ID in the source

Fixed this issue in conjunction with tbc-db PR https://github.com/cmangos/tbc-db/pull/818

### Proof

- None

### Issues

- None

### How2Test

- None


cala:
Assigning myself to this. Should be reviewed quickly for once (this evening if every is as expected).

cala:
Merged. Thank you very much! 👍

PR #434, opened 2021-03-05
Make Theramore Guards engage Dummies in Combat

insunaa:
## 🍰 Pullrequest

Theramore Practicing Guards didn't attack the target dummies before, now they do.

### Proof

- None

### Issues

- None

### How2Test

- `.go creature theramore combat`

### DIFF
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -628,6 +628,7 @@ UPDATE creature_template SET ScriptName='boss_tethyr' WHERE entry=23899;
 UPDATE creature_template SET ScriptName='npc_major_mills' WHERE entry=23905;
 UPDATE creature_template SET ScriptName='mob_invis_firework_helper' WHERE entry=24025;
 UPDATE creature_template SET ScriptName='npc_smolderwing' WHERE entry=23789;
+UPDATE creature_template SET ScriptName='npc_theramore_practicing_guard' WHERE entry=4951;
 
 INSERT INTO scripted_areatrigger VALUES
 (302,'at_sentry_point');
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -629,6 +629,7 @@ UPDATE creature_template SET ScriptName='npc_major_mills' WHERE entry=23905;
 UPDATE creature_template SET ScriptName='mob_invis_firework_helper' WHERE entry=24025;
 UPDATE creature_template SET ScriptName='npc_smolderwing' WHERE entry=23789;
 UPDATE creature_template SET ScriptName='npc_theramore_practicing_guard' WHERE entry=4951;
+UPDATE creature_template SET HealthMultiplier='1' WHERE entry=4952;
 
 INSERT INTO scripted_areatrigger VALUES
 (302,'at_sentry_point');
--- a/sql/scriptdev2/scriptdev2.sql
+++ b/sql/scriptdev2/scriptdev2.sql
@@ -629,7 +629,7 @@ UPDATE creature_template SET ScriptName='npc_major_mills' WHERE entry=23905;
 UPDATE creature_template SET ScriptName='mob_invis_firework_helper' WHERE entry=24025;
 UPDATE creature_template SET ScriptName='npc_smolderwing' WHERE entry=23789;
 UPDATE creature_template SET ScriptName='npc_theramore_practicing_guard' WHERE entry=4951;
-UPDATE creature_template SET HealthMultiplier='1' WHERE entry=4952;
+UPDATE creature_template SET HealthMultiplier='1', ScriptName='npc_theramore_combat_dummy' WHERE entry=4952;
 
 INSERT INTO scripted_areatrigger VALUES
 (302,'at_sentry_point');
--- a/src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
@@ -32,8 +32,11 @@ at_nats_landing
 boss_tethyr
 EndContentData */
 
+#include "AI/BaseAI/UnitAI.h"
+#include "AI/ScriptDevAI/ScriptDevAIMgr.h"
 #include "AI/ScriptDevAI/include/sc_common.h"
 #include "AI/ScriptDevAI/base/escort_ai.h"
+#include "AI/ScriptDevAI/include/sc_creature.h"
 #include "Common.h"
 #include "Entities/TemporarySpawn.h"
 #include "Globals/SharedDefines.h"
@@ -1039,17 +1042,6 @@ struct npc_theramore_practicing_guardAI : public ScriptedAI
         m_bsitDown = false;
         m_bstandUp = false;
         attackableDummy = GetClosestCreatureWithEntry(m_creature, NPC_THERAMORE_COMBAT_DUMMY, 2.f);
-        if (attackableDummy)
-        {
-            CreatureAI* targetDummyAI = dynamic_cast<CreatureAI*>(attackableDummy->AI());
-            if (targetDummyAI)
-            {
-                targetDummyAI->SetDeathPrevention(true);
-                targetDummyAI->SetReactState(REACT_PASSIVE);
-                targetDummyAI->SetRootSelf(true);
-                attackableDummy->RemoveFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_NPC);
-            }
-        }
     }
 
     void UpdateAI(const uint32 diff) override
@@ -1066,6 +1058,7 @@ struct npc_theramore_practicing_guardAI : public ScriptedAI
                     m_bisAttacking = false;
                     m_attackTimer = urand(120, 135) * IN_MILLISECONDS;
                     m_creature->CombatStop();
+                    attackableDummy->CombatStop();
                     m_bsitDown = true;
                     m_sitTimer = 2 * IN_MILLISECONDS;
                 }
@@ -1107,6 +1100,35 @@ struct npc_theramore_practicing_guardAI : public ScriptedAI
                 }
             }
         }
+    }
+};
+
+/*######
+## npc_theramore_combat_dummy
+######*/
+
+struct npc_theramore_combat_dummyAI : public ScriptedAI
+{
+    npc_theramore_combat_dummyAI(Creature* pCreature) : ScriptedAI(pCreature)
+    {
+        m_creature->RemoveFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_NPC);
+        SetReactState(REACT_PASSIVE);
+        SetRootSelf(true);
+    }
+
+    void Reset() override {}
+
+    void DamageTaken(Unit* /*dealer*/, uint32& damage, DamageEffectType /*damagetype*/, SpellEntry const* /*spellInfo*/) override
+    {
+        if (damage >= m_creature->GetHealth()){
+            m_creature->SetHealth(std::min(damage + 1, m_creature->GetMaxHealth()));
+        }
+        damage = std::min(damage, m_creature->GetHealth() - 1);
+    }
+
+    void UpdateAI(const uint32 diff) override
+    {
+        return;
     }
 };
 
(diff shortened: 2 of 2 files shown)


insunaa:
Basically the same as on the reference client with one notable difference: On CMaNGOS PreventDeath(true) makes all attacks that would cause damage beyond the HP of the creature to miss, dodge, etc.
On the reference client all of those attacks still hit but just don't cause damage that goes beyond 1HP.

The way to fix this would be to add a separate script for the combat dummy that does
```cpp
void DamageTaken(Unit* /*dealer*/, uint32& damage, DamageEffectType /*damagetype*/, SpellEntry const* /*spellInfo*/) override
{
    damage = std::min(damage, m_creature->GetHealth() - 1);
}
```
But I was too lazy.

insunaa:
OK, this script is now 99% in line with the reference client. The only thing I'm still unsure about is the timers. My research suggests they are ~15-40 for the break period and 130-140 for the attacking period, but I think it's fine to leave them as they are here.

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
```diff
@@ -1000,6 +1000,126 @@ UnitAI* GetAI_mob_invis_firework_helper(Creature* pCreature)
...
+        m_breakTimer = 20 * IN_MILLISECONDS;
+        m_bisAttacking = false;
+        m_bsitDown = false;
+        m_bstandUp = false;
+        attackableDummy = GetClosestCreatureWithEntry(m_creature, NPC_THERAMORE_COMBAT_DUMMY, 2.f);
+    }
+
+    void UpdateAI(const uint32 diff) override
+    {
+        if (attackableDummy && m_creature->GetVictim() && m_creature->GetVictim() != attackableDummy)
```
insunaa, line 1040: This exists because
```cpp
m_creature->SetNoCallAssistance(true);
```
apparently has no effect, meaning that when one Theramore Practicing Guard starts attacking its dummy, the other one also reacts and wants to attack that dummy. This line is preventing this from happening, but you can still hear the shout NPCs do when they enter combat.

If there is a more elegant solution to this, I'm very open to it

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
```diff
@@ -1000,6 +1000,126 @@ UnitAI* GetAI_mob_invis_firework_helper(Creature* pCreature)
...
+
+/*######
+## npc_theramore_combat_dummy
+######*/
+
+struct npc_theramore_combat_dummyAI : public ScriptedAI
+{
+    npc_theramore_combat_dummyAI(Creature* pCreature) : ScriptedAI(pCreature)
+    {
+        m_creature->RemoveFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_NPC);
```
killerwife, line 1105: Why are you putting it here, why is it in db in the first place.

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
```diff
@@ -1000,6 +1000,126 @@ UnitAI* GetAI_mob_invis_firework_helper(Creature* pCreature)
...
+{
+    NPC_THERAMORE_COMBAT_DUMMY = 4952,
+};
+
+struct npc_theramore_practicing_guardAI : public ScriptedAI
+{
+    npc_theramore_practicing_guardAI(Creature* pCreature) : ScriptedAI(pCreature)
+    {
+        Reset();
+        m_creature->SetFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_IMMUNE_TO_PLAYER);
```
killerwife, line 1016: Same point as dummy

sql/scriptdev2/scriptdev2.sql
```diff
@@ -621,6 +621,8 @@ UPDATE creature_template SET ScriptName='boss_tethyr' WHERE entry=23899;
 UPDATE creature_template SET ScriptName='npc_major_mills' WHERE entry=23905;
 UPDATE creature_template SET ScriptName='mob_invis_firework_helper' WHERE entry=24025;
 UPDATE creature_template SET ScriptName='npc_smolderwing' WHERE entry=23789;
+UPDATE creature_template SET ScriptName='npc_theramore_practicing_guard' WHERE entry=4951;
+UPDATE creature_template SET HealthMultiplier='1', ScriptName='npc_theramore_combat_dummy' WHERE entry=4952;
```
killerwife, line 625: This HP multi should be in DB

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
insunaa: No idea why it's in the DB, but it's causing the Guard to be unable to attack the dummy, it's just standing in the idle "ready to attack animation" and staring at the dummy.
I didn't want to mess with the DB regarding Unit Flags, because I don't know how or why they got there, so I remove them as I need them for the purposes of my script.

sql/scriptdev2/scriptdev2.sql
insunaa: The HP Multi in DB is the one that was sniffed to make the NPCs HP be `1` (so in DB it was `0.021` or something similar to that). On the reference client, the dummy has 42 HP, which corresponds to a multiplier of `1`. The problem with giving the dummy the `0.021` multiplier as it is currently is that any method of death-prevention would cause the server to send a damage value of `0` to clients, which will be interpreted as a miss, dodge, resist, etc. so it doesn't do the correct "getting-attacked animation" (well the visual animation is correct, but because it's all misses and dodges, it will always just "whoosh" instead of the response attacks get when they actually connect.

insunaa: If you want I can also edit this multiplier value in the db repo and create a separate PR over there, but it currently conflicts with sniff-data

insunaa:
OK, latest commit now handles interference from the outside (like a player attacking the guard), without having to resort to the guard just being immune to players. I'm not sure if this is the right way to do it, but it's the best I could think of

cala:
To be honest, I don't remember clearly how this was back on original retail, so I will need some clarifications in order to review properly this PR, if you don't mind.

Are we sure that the guards were really attacking the dummy targets (I mean, the action was visible in the combat log) and not that they were pretending attacking them (just playing the attack animation) ?

insunaa:
Unfortunately I don't know how it was in 1.12. I know that the way I'm doing it is how it's working in 1.13. I have tried to find videos from 2004-2008 but I couldn't find any.

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
```diff
@@ -1000,6 +1001,175 @@ UnitAI* GetAI_mob_invis_firework_helper(Creature* pCreature)
...
+        Reset();
+    }
+
+    uint32 m_attackTimer;
+    uint32 m_breakTimer;
+    uint32 m_sitTimer;
+    bool m_bisAttacking;
+    bool m_bsitDown, m_bstandUp;
+    bool m_binCombatWithPlayer;
+    Creature* attackableDummy;
```
Chero, line 1026: Consider storing the creature guid instead of creature pointer. If the attackableDummy gets removed from the world for any reason, you will lose traceability.

If it gets deleted -
if (attackableDummy != nullptr) will return true, 
attackableDummy->AnyMethod will crash your application

Instead, store the creatures guid in ObjectGuid and get it in realtime when you need it from Map.

insunaa: Oh right yeah I forgot to change it, thanks

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
```diff
@@ -1000,6 +1001,175 @@ UnitAI* GetAI_mob_invis_firework_helper(Creature* pCreature)
...
+enum
+{
+    NPC_THERAMORE_COMBAT_DUMMY = 4952,
+};
+
+struct npc_theramore_practicing_guardAI : public ScriptedAI
+{
+    npc_theramore_practicing_guardAI(Creature* pCreature) : ScriptedAI(pCreature)
+    {
+        attackableDummy = GetClosestCreatureWithEntry(m_creature, NPC_THERAMORE_COMBAT_DUMMY, 2.f);
```
Chero, line 1016: This can result in nullptr when the Guard spawns sooner than the Dummy.

Creatures do not spawn at once, but one after another.

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
```diff
@@ -1004,6 +1007,192 @@ UnitAI* GetAI_mob_invis_firework_helper(Creature* pCreature)
...
+        m_bisAttacking = false;
+        m_bsitDown = false;
+        m_bstandUp = false;
+        m_binCombatWithPlayer = false;
+        m_creature->SetStandState(UNIT_STAND_STATE_STAND);
+    }
+
+    void UpdateAI(const uint32 diff) override
+    {
+        if (!attackableDummy && !m_creature->IsInCombat() && !m_creature->IsMoving())
```
Chero, line 1098: Consider adding a timer here so this is not called every app heartbeat.

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
```diff
@@ -1004,6 +1007,192 @@ UnitAI* GetAI_mob_invis_firework_helper(Creature* pCreature)
...
+            if(myDummy && myDummy->IsInCombat())
+            {
+                myDummy->CombatStop();
+            }
+            ScriptedAI::UpdateAI(diff);
+            return;
+        }
+        if (myDummy && myDummy->IsAlive())
+        {
+            Unit* pTarget = m_creature->GetTarget();
```
Chero, line 1113: Since you are using pTarget only to get its guid, its faster to use m_creature->GetTargetGuid() if you only want to compare the guids.

By this, you will avoid one unnecessary lookup in Map storage.

insunaa:
Had to rebase the branch because the server wouldn't run with the old base, hence the weird commit history.

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
insunaa: ooh, yeah makes sense. Didn't know that existed!

insunaa: It's a protected function, that's probably why I didn't know about it

Chero: Ohh, shhuu. @killerwife maybe time to open those calls to public :P

Chero: On the other hand, there is bool HasTarget(ObjectGuid)

src/game/AI/ScriptDevAI/scripts/kalimdor/dustwallow_marsh.cpp
insunaa: it's called every heartbeat, but most heartbeats shouldn't have an uninitialized `attackableDummy` ObjectGuid. I think that has less impact than actually checking for a timer.

insunaa:
I guess I'll re-implement it in CombatAI at some point, that might be cleaner

insunaa:
OK, reimplemented the whole thing in CombatAI, should be much cleaner now

insunaa:
OK, this should be done, now. If requested I'll squash the commits, but I think this should now be a nicely working script.

Grz3s:
@cala 
`(...)Are we sure that the guards were really attacking the dummy targets(...)` - they are - at least in wotlk sniffs.. where `OverDamage` is used... not sure what exact mechanic was used in vanilla or TBC.. 
We may need to  uodate  `42  ACTION_T_SET_DEATH_PREVENTION` -- to allow that overdmg to happen... then we can see dmg done and `spark = hit` animation

cala:
Should be done in https://github.com/cmangos/wotlk-db/commit/33b64c52b626098bd9bbf4375c20537b48c0bf7f
Is it what you wanted to do here ?

insunaa:
Yuppers, that looks to be what mine was intended to do :)

cala:
Great! 👍🏻

PR #435, opened 2021-03-06
Make Travis work at least

insunaa:
## 🍰 Pullrequest

Travis used to fail on OSX because an OSX Version unsupported by Homebrew was used. This PR changes the OSX Version to 10.14, the oldest OSX supported by Homebrew.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ matrix:
 
     # osx builds on clang x64
     - os: osx
-      osx_image: xcode10
+      osx_image: xcode10.2
       arch: amd64
       compiler: clang
       env:


cala:
> This PR changes the OSX Version to 10.14, the oldest OSX supported by Homebrew.

Ouch! 😒 

If Mojave (macOS 10.14) is the oldest version of macOS that is supported by Homebrew, that means that next year it will be dropped when the successor of BigSur is released. That's a real trouble for users like me that stick to Mojave because this is the latest version of macOS able to launch Classic and TBC client (and maybe also WotLK, but I'm not sure)...

cala:
Merged, thank you very much!

PR #436, opened 2021-03-07
Silence Clang compiler warnings

insunaa:
## 🍰 Pullrequest

This PR fixes current clang compiler warnings.
Most are caused by switch statements that don't handle every case. Some are evaluations of assignments, some are dangling else statements and one is what Clang sees as an evaluation of an address

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/netherstorm.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/netherstorm.cpp
@@ -181,7 +181,9 @@ struct npc_manaforge_spawnAI : public ScriptedAI
         Creature *manaforge;
 
         if (m_manaforgeGuid)
-            if ((manaforge = m_creature->GetMap()->GetCreature(m_manaforgeGuid)))
+        {
+            manaforge = m_creature->GetMap()->GetCreature(m_manaforgeGuid);
+            if (manaforge)
             {
                 uint32 uiManaforgeEntry = manaforge->GetEntry();
 
@@ -230,6 +232,7 @@ struct npc_manaforge_spawnAI : public ScriptedAI
                     m_creature->GetMotionMaster()->MovePoint(uiMPID, fX, fY, m_creature->GetPositionZ());
                 }
             }
+        }
     }
 
     void UpdateAI(const uint32 /*uiDiff*/) override
--- a/src/game/AI/ScriptDevAI/scripts/outland/netherstorm.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/netherstorm.cpp
@@ -181,7 +181,7 @@ struct npc_manaforge_spawnAI : public ScriptedAI
         Creature *manaforge;
 
         if (m_manaforgeGuid)
-            if (manaforge = m_creature->GetMap()->GetCreature(m_manaforgeGuid))
+            if ((manaforge = m_creature->GetMap()->GetCreature(m_manaforgeGuid)))
             {
                 uint32 uiManaforgeEntry = manaforge->GetEntry();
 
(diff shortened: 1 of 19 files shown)


src/game/Spells/Spell.cpp
```diff
@@ -792,7 +792,7 @@ void Spell::AddUnitTarget(Unit* target, uint8 effectMask, CheckException excepti
     targetInfo.heartbeatResistChance = 0;
 
     // Calculate hit result
-    targetInfo.missCondition = (m_ignoreHitResult ? SPELL_MISS_NONE : m_caster->SpellHitResult(target, m_spellInfo, targetInfo.effectMask, m_reflectable, &targetInfo.heartbeatResistChance));
+    targetInfo.missCondition = (m_ignoreHitResult ? SPELL_MISS_NONE : m_caster->SpellHitResult(target, m_spellInfo, targetInfo.effectMask, m_reflectable, &(targetInfo.heartbeatResistChance)));
```
insunaa, line 795: I'm not sure about this one, can somebody else confirm that this is correct?

src/mangosd/CliRunnable.cpp
```diff
@@ -135,7 +135,7 @@ bool ChatHandler::GetDeletedCharacterInfoList(DeletedInfoList& foundList, std::s
     {
         if (resultChar->GetRowCount() > 100)
         {
-            PSendSysMessage("Too many results %llu. Narrow it down.", resultChar->GetRowCount());
+            PSendSysMessage("Too many results %lu. Narrow it down.", resultChar->GetRowCount());
```
insunaa, line 138: This one is also a bit suspect, as it is dependent on the environment (for example on Linux, `lu` is equivalent to uint64, on MacOS `llu` is equivalent to uint64, so I'm guessing this one should probably remain llu and we'll just ignore that particular warning)

cyberium: I suggest to cast the result to uint32 and use %u format

src/game/Spells/Spell.cpp
cyberium: What's the warning message?

insunaa:
```
field heartbeatResistChance
Type: uint32
Value = true

// In Spell::TargetInfo
public: uint32 heartbeatResistChance
Address of 'targetInfo.heartbeatResistChance' will always evaluate to 'true'clang(-Wpointer-bool-conversion)
```

src/game/AI/ScriptDevAI/scripts/outland/netherstorm.cpp
```diff
@@ -181,7 +181,7 @@ struct npc_manaforge_spawnAI : public ScriptedAI
         Creature *manaforge;
 
         if (m_manaforgeGuid)
-            if (manaforge = m_creature->GetMap()->GetCreature(m_manaforgeGuid))
+            if ((manaforge = m_creature->GetMap()->GetCreature(m_manaforgeGuid)))
```
cyberium, line 184: I understand that you tried to make compiler happy there.
But this way to test the result of the GetCreature() result is not good to me and its the main issue here.
I would suggest to broke that in 2 instructions

```cpp
if (m_manaforgeGuid)
{
     manaforge = m_creature->GetMap()->GetCreature(m_manaforgeGuid);
     if (manaforge)
     {
          .....
     }
}

```

insunaa: I had it that way before that, but I saw that several other similar comparisons were doing it in this style, so I adapted it there, too

src/game/Spells/Spell.cpp
cyberium: This one is an error. SpellHitResult prototype have probably changed over time and this happened. Will summon @killerwife for this. Ignore this for this pr.

killerwife: Fixed on my own accord.

cyberium:
Merged thank you

PR #438, opened 2021-03-10
Cathedral Ashbringer Event Hackfix

insunaa:
## 🍰 Pullrequest

When MoveInLineOfSight is called, the call isn't forwarded if the Ashbringer Event is in Progress

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/scarlet_monastery/instance_scarlet_monastery.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/scarlet_monastery/instance_scarlet_monastery.cpp
@@ -40,14 +40,18 @@ void instance_scarlet_monastery::OnCreatureCreate(Creature* pCreature)
     switch (pCreature->GetEntry())
     {
         case NPC_MOGRAINE:
-            m_npcEntryGuidStore[pCreature->GetEntry()] = pCreature->GetObjectGuid();
-            if (GetData(TYPE_ASHBRINGER_EVENT) == IN_PROGRESS)
-                DoOrSimulateScriptTextForThisInstance(SAY_ASHBRINGER_ENTRANCE, NPC_MOGRAINE);
-            break;
+            DoOrSimulateScriptTextForThisInstance(SAY_ASHBRINGER_ENTRANCE, NPC_MOGRAINE);
         case NPC_WHITEMANE:
         case NPC_VORREL:
             m_npcEntryGuidStore[pCreature->GetEntry()] = pCreature->GetObjectGuid();
             break;
+    }
+    if (!pCreature->IsCritter()){
+        monasteryNPCs.push_back(pCreature);
+        if (GetData(TYPE_ASHBRINGER_EVENT) == IN_PROGRESS)
+        {
+            pCreature->SetFactionTemporary(35, TEMPFACTION_RESTORE_RESPAWN);
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


al3xc1985:
does this fix also the npc's being enemies and not friendly?

insunaa:
Nope, that's independent.
But there's a different issue for that already, I think. ( https://github.com/cmangos/issues/issues/1983 )
Tho that is slightly different

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/scarlet_monastery/instance_scarlet_monastery.cpp
```diff
@@ -40,15 +40,20 @@ void instance_scarlet_monastery::OnCreatureCreate(Creature* pCreature)
     switch (pCreature->GetEntry())
     {
         case NPC_MOGRAINE:
-            m_npcEntryGuidStore[pCreature->GetEntry()] = pCreature->GetObjectGuid();
             if (GetData(TYPE_ASHBRINGER_EVENT) == IN_PROGRESS)
                 DoOrSimulateScriptTextForThisInstance(SAY_ASHBRINGER_ENTRANCE, NPC_MOGRAINE);
```
insunaa, line 44: This is never executed for some reason. Guessing 2 reasons: 1) The Dialogue isn't implemented and/or 2) Mograine doesn't spawn on instance enter, so this code only gets called later when he is eventually spawned

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/scarlet_monastery/instance_scarlet_monastery.cpp
```diff
@@ -40,15 +40,20 @@ void instance_scarlet_monastery::OnCreatureCreate(Creature* pCreature)
...
-            break;
         case NPC_WHITEMANE:
         case NPC_VORREL:
             m_npcEntryGuidStore[pCreature->GetEntry()] = pCreature->GetObjectGuid();
             break;
     }
+    if (!pCreature->IsCritter()){
+        monasteryNPCs.push_back(pCreature);
+        if (GetData(TYPE_ASHBRINGER_EVENT) == IN_PROGRESS)
```
insunaa, line 52: This might seem redundant, but not all creatures spawn when the user enters, some of them spawn way later.
If I *only* do this OnPlayerEnter, all NPCs in the chapel will remain hostile. If I *only* do this OnCreatureCreate: the first creatures will be created before the ASHBRINGER_EVENT is set to `IN_PROGRESS` so they also wouldn't get the new faction.

insunaa:
OK, this works as a temporary fix-all solution now.

The actual issue seems to be that https://classic.wowhead.com/spell=28282/ashbringer is not applied in the 5s interval it should be, instead it's applied once on item-equip and then not again. My hackfix now manually casts this aura on the user when they enter the instance, turning the NPCs friendly.

The real solution would be to change the spellscript so that this aura actually is applied periodically.

MantisLord:
While your changes here are essentially valid, please hold off on doing more on this. There's unexported work related to this which I will take care of applying soon, the event should soon be in a fully working state. We will no longer use MoveInLOS checks for anything.

insunaa:
Good to hear!
The idea of this PR was to be a stopgap, but if a permanent fix is incoming, then it's even better :)

al3xc1985:
ty @MantisLord for the good news

PR #440, opened 2021-03-11
Hackfix for Ashcrombe not really wanting to look at the door

insunaa:
## 🍰 Pullrequest

Makes Ashcrombe look at the door while unlocking and despawn upon finishing his teleport.

### Proof

- None

### Issues

- Hack fixes https://github.com/cmangos/wotlk-db/issues/363

### How2Test

- Create alliance character
- `.gm on`
- `.tele sfk`
- `.die` on the worgen boss
- talk to and follow ashcrombe

### Todo / Checklist

- [ ] Fix door-adjacent waypoints in DB. I have no idea how to do that, someone else will have to.
- [ ] Make this thing not-hack-fix-y. The entire door-sequence is sequenced by timed waypoints, instead of a proper script. I don't know how to do it properly, but this seems very improper.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/shadowfang_keep/shadowfang_keep.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/shadowfang_keep/shadowfang_keep.cpp
@@ -86,7 +86,11 @@ struct npc_shadowfang_prisonerAI : public npc_escortAI
                 break;
             case 12:
                 if (m_uiNpcEntry == NPC_ASH)
-                    DoCastSpellIfCan(m_creature, SPELL_UNLOCK);
+                    {
+                        m_creature->GetMotionMaster()->PauseWaypoints(2000);
+                        DoCastSpellIfCan(m_creature, SPELL_UNLOCK);
+                        m_creature->SetFacingToObject(m_pInstance->GetSingleGameObjectFromStorage(GO_COURTYARD_DOOR));
+                    }
                 else
                     DoScriptText(EMOTE_UNLOCK_DOOR_AD, m_creature);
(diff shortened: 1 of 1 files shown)


src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/shadowfang_keep/shadowfang_keep.cpp
```diff
@@ -86,7 +86,11 @@ struct npc_shadowfang_prisonerAI : public npc_escortAI
                 break;
             case 12:
                 if (m_uiNpcEntry == NPC_ASH)
-                    DoCastSpellIfCan(m_creature, SPELL_UNLOCK);
+                    {
+                        m_creature->GetMotionMaster()->PauseWaypoints(2000);
```
cala, line 90: Waypoint pause for a set amount of time should be handled in the waypoint path itself and thus, it should allow to also set the proper orientation there, making this whole change unneeded.

cala (commented):
Thank you for this PR. I agree the whole script is a bit old and could benefit from a small rework, though using timed waypoint is still a common thing.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/shadowfang_keep/shadowfang_keep.cpp
insunaa: technically setting the facing is also unneeded, if the waypoint terminated while the character was standing directly in front of the door, having come from a waypoint that is centered and perpendicular to the door, so I guess all that's really needed is to rework the waypoints.

I don't have any proper tools for that, tho :/

PR #441, opened 2021-03-11
Fix characters start naked bug

insunaa:
## 🍰 Pullrequest

This PR should fix the bug that sometimes characters start naked and new items aren't saved at some point.

### Proof

- None

### Issues

- Fixes https://github.com/cmangos/issues/issues/2317

#### Before applying patch:
- Create a warrior character
- .level 69
- .additem 23529
- Restart the server.
- Log in to the new character and apply the sharpening stone to your weapon
- log out
- Any subsequently creater new items are no longer saved, including the clothes/gear of new characters
#### After applying patch:
- Same steps as above, but note how things work, now.

### Todo / Checklist

- [ ] Long term testing
- [x] Find a way to make a different SqlStatementID for Insert and Update, because SqlStatement doesn't allow for uninitialized variables. Maybe by indirecting it with a pointer?
- [x] Find out why it sometimes started to happen after the server has already been running fine for hours or days.

Edit: I would just like to exclaim in this footnote how remarkably upset I am at my IDE for constantly inserting unnecessary includes. It is extremely infuriating.

### **The cause was found by discord user mostlikely#6332**

### DIFF
--- a/src/game/Entities/Item.cpp
+++ b/src/game/Entities/Item.cpp
@@ -301,16 +301,9 @@ void Item::SaveToDB()
         case ITEM_NEW:
         case ITEM_CHANGED:
         {
-            static SqlStatementID dummyID, insItem, updItem;
-            SqlStatement stmt = CharacterDatabase.CreateStatement(dummyID, "");
-            if (uState == ITEM_NEW)
-            {
-                stmt = CharacterDatabase.CreateStatement(insItem, "REPLACE INTO item_instance (owner_guid, itemEntry, creatorGuid, giftCreatorGuid, count, duration, charges, flags, enchantments, randomPropertyId, durability, itemTextId, guid) VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?)");
-            }
-            else
-            {
-                stmt = CharacterDatabase.CreateStatement(updItem, "UPDATE item_instance SET owner_guid = ?, itemEntry = ?, creatorGuid = ?, giftCreatorGuid = ?, count = ?, duration = ?, charges = ?, flags = ?, enchantments = ?, randomPropertyId = ?, durability = ?, itemTextId = ? WHERE guid = ?");
-            }
+            static SqlStatementID insItem, updItem;
+
+            SqlStatement stmt = CharacterDatabase.CreateStatement(uState == ITEM_NEW ? insItem : updItem, uState == ITEM_NEW ? "REPLACE INTO item_instance (owner_guid, itemEntry, creatorGuid, giftCreatorGuid, count, duration, charges, flags, enchantments, randomPropertyId, durability, itemTextId, guid) VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?)" : "UPDATE item_instance SET owner_guid = ?, itemEntry = ?, creatorGuid = ?, giftCreatorGuid = ?, count = ?, duration = ?, charges = ?, flags = ?, enchantments = ?, randomPropertyId = ?, durability = ?, itemTextId = ? WHERE guid = ?");
 
             stmt.addUInt32(GetOwnerGuid().GetCounter());
             stmt.addUInt32(GetEntry());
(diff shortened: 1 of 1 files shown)


killerwife:
I will wait a bit until some of the people who experience the bug test it since I cannot test it. Code wise, I partly understand why this issue occurs.

insunaa:
I don't know if it makes more sense to keep the StatementID as static or to make it non-static. non-static probably has performance implications

src/game/Entities/Item.cpp
```diff
@@ -301,11 +302,16 @@ void Item::SaveToDB()
         case ITEM_NEW:
         case ITEM_CHANGED:
         {
-            SqlStatementID insItem;
-            SqlStatement stmt = CharacterDatabase.CreateStatement(insItem, uState == ITEM_NEW ?
-                "REPLACE INTO item_instance (owner_guid, itemEntry, creatorGuid, giftCreatorGuid, count, duration, charges, flags, enchantments, randomPropertyId, durability, itemTextId, guid) VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?)" :
-                "UPDATE item_instance SET owner_guid = ?, itemEntry = ?, creatorGuid = ?, giftCreatorGuid = ?, count = ?, duration = ?, charges = ?, flags = ?, enchantments = ?, randomPropertyId = ?, durability = ?, itemTextId = ? WHERE guid = ?"
-            );
+            static SqlStatementID dummyID, insItem, updItem;
+            SqlStatement stmt = CharacterDatabase.CreateStatement(dummyID, "");
```
evil-at-wow, line 305: That's a bit of a weird construction I think. If the difference between the statements for new items and updates is what causes the bug, just create the statements with an index as needed? Something like this:
```
static SqlStatementID insItem, updItem;
SqlStatement stmt = CharacterDatabase.CreateStatement(uState == ITEM_NEW ? insItem : updItem, uState == ITEM_NEW ? "..." : "..." );
```

insunaa: I must be misremembering things then, because I believe that some time ago I was told that I can't have 2 ternary operators in the same line.

If that works tho, yours is definitely more elegant

killerwife: @evil-at-wow could this be an issue with multithreading due to static being used in multiple threads at the same time? Since it is being altered, it should instead be a thread-local variable I believe.

evil-at-wow: I've been asking myself the same question actually, and race conditions might *also* be involved in the end. Now, merely making the static a thread-local variable won't help if the different queries are the problem here (which seems very plausible to me at first sight as the SqlStatementID seems to be mapped to the actual query string). With just a thread-local instead of the static a given thread will initialize it the first time (say using the query for a new item) and will later simply reuse the already initialized index (similar to the static case, but per thread), which will likely execute the wrong query in case it's meant to be an update.

At first sight two threads racing to initialize a (static) SqlStatementID doesn't seem too terrible. There's a lock in Database::CreateStatement(), so all threads will eventually map the same query to the same ID. All threads will call init() on the SqlStatementID, which is not ideal and strictly speaking a data race, but since they will all write the same data you probably get away with it in practice, although a thread-local would probably solve that race too.

killerwife: Well big IMHO, all these SQLStatement static variables should be changed to thread local and use two variables for each of the IF branches or go the simple route of always constructing variable. Maybe a wee bit of optimization doesnt hurt however.

canidae: Looking briefly at this code, it seems to me that it more or less "always" used either `REPLACE INTO` or `UPDATE` as they both share the same `SQLStatementID`. `Database::CreateStatement()` seems to cache prepared statements (using SQLStatementID as key) and since `REPLACE INTO` likely will work in most (all?) cases where items are persisted, I'm guessing initialization of the `REPLACE INTO` statement _usually_ won the race condition. If the cache is cleaned out from time to time then the `UPDATE` statement might take precedence, which will cause items not to be persisted.
If that's the case, then adding one variable for each statement like in this PR, then the issue should indeed be fixed. Do add the change from evil-at-wow though, you can use two ternary expressions on the same line and the "dummyID" variable is ugly.

PR #443, opened 2021-03-15
Fix Distract Spell breaking Stealth of Target

insunaa:
## 🍰 Pullrequest

With this PR Distract does not break stealth of its target anymore.

### Proof

- None

### Issues

- Fixes https://github.com/cmangos/issues/issues/2415

### How2Test

- Create 2 Rogues
- Learn spell 1725
- Both rogues enter stealth and duel
- Cast Distract
- Profit

### Todo / Checklist

- [ ] Maybe some flags in the database for that spell are incorrect, this spell is interpreted as attack/having a negative effect, so it breaks target stealth. I don't know how spell flags work and I couldn't see anything wrong according to https://github.com/cmangos/issues/wiki/spell_template#attributesex so I hardcoded the spell as a positive effect.

### DIFF
--- a/src/game/Spells/SpellMgr.h
+++ b/src/game/Spells/SpellMgr.h
@@ -1079,6 +1079,7 @@ inline bool IsPositiveEffect(const SpellEntry* spellproto, SpellEffectIndex effI
 
     switch (spellproto->Id) // Spells whose effects are always positive
     {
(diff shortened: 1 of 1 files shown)


diluh:
@killerwife @insunaa 
Fixed thanks I just took the test everything works perfectly.

Tested with Rogue and Druid..

PR #445, opened 2021-03-20
[Instance] Rework Dialogue in SFK

insunaa:
## 🍰 Pullrequest

Reworking how dialogue is done within Shadowfang Keep.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [ ] Figure out why the lightning effect that should happen 5 seconds into Arugal's dialogue doesn't happen.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/shadowfang_keep/instance_shadowfang_keep.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/shadowfang_keep/instance_shadowfang_keep.cpp
@@ -269,43 +269,45 @@ void instance_shadowfang_keep::JustDidDialogueStep(int32 iEntry)
     {
         case VINCENT_DEATH:
             if (Creature* pVincent = GetSingleCreatureFromStorage(NPC_VINCENT))
+            {
                 pVincent->SetStandState(UNIT_STAND_STATE_DEAD);
+            }
             break;
         case ARUGAL_VISIBLE:
-        if (Creature* pCreature = GetSingleCreatureFromStorage(NPC_ARUGAL))
-        {
-            pCreature->SetVisibility(VISIBILITY_ON);
-        }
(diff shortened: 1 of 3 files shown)


insunaa:
OK Windows build is failing, but it's working fine on clang. I'll try to figure it out myself but if anybody has an idea feel free to reply

insunaa:
@cala fyi

cala:
Thank you very much for your work.
Assigning myself on this PR.

Tobschinski:
how selfish

insunaa:
One thing I'd like to note: I'm faithfully recreating the dialogue scene based on 1.13. When you enter the dungeon vincent is "alive"/standing. 2 seconds after entering Vincent dies/falls over then another 8 seconds after that Arugal teleports in, does his monologue and teleports out.

I've watched several videos on youtube to confirm this, this is how it happens on 1.13.

I have some doubts that this is how it's intended to play out. My instinct says the monologue should start and only after the first text bubble should Vincent die. But that was neither how it worked before my rework, nor was it how it works in 1.13 so I stuck with the latest version of things for now, until I have better evidence for how it actually is supposed to be.

cala:
Now that I'm done with the two prisoners in Shadowfang Keep, I will selfishly review this PR.

I will stick with the timing you've updated as I highly suspect that the previous implementation was mostly guess work.

cala:
Merged, thank you very much. It took me a little longer than expected because I wanted to find a way to prevent the script for Arugal's Voidwalkers to be split in two files.

The lightning visual is still not working but the code is ran. So, either we have wrong DB data for the related GameObject, either we are somehow handling this case incorrectly.

insunaa:
And you found an elegant solution. I'm sorry I didn't come up with it :/
The lightning visual wasn't working before either, so I don't think we're handling this incorrectly script-wise. I think the GO may be working differently than we think, or that the DB is missing some entries for it, or something like that.

PR #446, opened 2021-03-20
Make Coilfang Frenzies not spawn in shallow water

insunaa:
## 🍰 Pullrequest

Stop Coilfang Frenzys spawning in the wrong places

### Proof

- None

### Issues

- Fixes https://github.com/cmangos/issues/issues/2425
### How2Test

- `.go cr morogrim`

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -20735,7 +20735,8 @@ void Player::UpdateTerainEnvironmentFlags(Map* m, float x, float y, float z)
                                     CastSpell(this, liquidSpellId, TRIGGERED_OLD_TRIGGERED);
                                 else
                                 {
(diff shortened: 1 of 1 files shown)


AnonXS:
i added comment ` // Exclude Water at Lady Vashj` so its visible why this "extra" is needed.

killerwife:
1. This keyword is redundant. 2. Why the -20 check if I might ask? Isnt morogrim water puddle not swimmable?

insunaa:
I like using the `this` keyword so that it's unambiguous to the reader what function is being called.
The -20 check is so that the frenzies don't spawn in the basin around Lady Vashj

killerwife:
Lady vashj basin is confirmed to spawn fish, confirmed yday, and this keyword is not used per codestyle of cmangos. It is only to be used when context is unambiguous to the compiler.

insunaa:
Alright, I'll remove the keyword and Z check

PR #448, opened 2021-03-29
Fix memory leak when deleting LoginQueryHolder.

Chero:
## 🍰 Pullrequest

m_accountId and m_guid are leaking when LoginQueryHolder is deleted.

new-delete-type-mismatch


cyberium:
Merged thank you very much https://github.com/cmangos/mangos-tbc/commit/ab1f9ad6db925049b8b760612075963f84209376

PR #449, opened 2021-03-29
Fix code execution order in guild removal process to avoid using dele…

Chero:
…ted data.

## 🍰 Pullrequest
DelMember method deletes slot from members, making slot pointer to point to a deleted memory.


cyberium:
https://github.com/cmangos/mangos-tbc/commit/f368fd77ef3e49e8064e34f6c5b29c97538e2688

Thank you very much

PR #451, opened 2021-03-30
[Instance] SFK: Improve timing for Arugal dialogue

insunaa:
## 🍰 Pullrequest

Arugal timing was off, he became visible and then started casting the teleport-in animation, all around it looked weird, the timing was off and a lot too fast in many steps.
It looked fine in my original re-work, but after merging the timing seemed off, so I double-checked and re-did the timings.


cala:
I may be partly in fault for the visual spell timing as I changed a bit of it. Not while merging but in a later commit. I changed the spell ID for the teleport visual/spawn-in for the one used by Arugal. There is an important difference between the two spells : the previous one has no casting time while the Arugal one has a 2 seconds casting time. Thus, I slighty delay the teleport out to let the visual play. Nothing more.

insunaa:
That might've been it. But the dialogue was still a bit too quick, beat by beat.

Either way on 1.13 at least, the teleport isn't a spell that's being cast
https://www.youtube.com/watch?v=1aWJD-LmL80

Tobschinski:
Too lazy to check what spells you guys currently use, but the correct ids are 7741 and 7136.

cala:
Thanks. 7741 for teleport in and 7136 for teleport out ?

insunaa:
Yup exactly. Just tested it.
I'll change the PR around to fit the correct IDs and adjusted timing.

insunaa:
Just noticed that spell is already defined as `SPELL_SHADOW_PORT_STAIRS` for the Archmage Arugal Bossfight. Should it rather be defined twice or should we move spell definitions around?

cala:
I won't have access to a computer until tomorrow, so I can't check for sure but if these are spells used by Arugal during the boss fight, they are real teleport spells, not just dummy visual.

insunaa:
@cala the first one is just the visual, the second one is an actual teleport. In the video I linked you can actually see tho, that in 1.13 Arugal actually changes position (because the textbox that's attached to him moves across the screen) and the direction he moves in looks suspiciously much like his boss-room, so I'm mostly sure that this is legitimately what's done.

insunaa:
Oof, side by side I'm seeing that the "pointing" emote comes too soon

insunaa:
OK, watching the event as it is now side-by-side with that video from 1.13 they're identical now, as far as I can tell.

cala:
Merged, thank you very much for:
- this PR along with the previous ones
- taking the time to polish and (re)-check your changes and changes others made to you code, that's how quality work is done

PR #455, opened 2021-04-05
Use 3D distance when checking for nearby polygons.

Chero:
## 🍰 Pullrequest

Remove pathfinding issues caused by incorrectly picking the nearest polygon to the destination point.
These cases happen where there are multiple platforms. If the player is on platform level 0 and the chasing creature is coming from a level 1 platform, it will pick a polygon that is closer to the creature (on level 1), ignoring height at all.

This can be reproduced for example in Blackwing Lair - Vaelastrasz' room (the balconies).

### Proof

- None

### Issues

- None

### How2Test

- Go to Blackwing Lair, Vaelastrasz' room, stay under the balcony and pull a creature from the pack that is located behind the gate on the platform above. Without this change, the creature will come to a point above your head.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/MotionGenerators/PathFinder.cpp
+++ b/src/game/MotionGenerators/PathFinder.cpp
@@ -137,8 +137,7 @@ dtPolyRef PathFinder::getPathPolyByPosition(const dtPolyRef* polyPath, uint32 po
         return INVALID_POLYREF;
 
     dtPolyRef nearestPoly = INVALID_POLYREF;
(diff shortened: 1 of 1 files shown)


Chero:
Just noticed it's already corrected in TC, at https://github.com/TrinityCore/TrinityCore/commit/2d6237f06ae6b74bb8ebfe745d8c0f1ce440e2fe

PR #458, opened 2021-05-04
Rework of main timer in WorldRunnable.cpp, smoother updates

Lymphome:
## 🍰 Pullrequest
Smoother timer method. Current timing relies on the previous duration of update, but update durations are very inequal.
* Clean unused vars
* Rework of main timer

### Proof
"we want that next d1 + t1 == WORLD_SLEEP_CONST", what else

### Issues
The duration of an update doesn't express the duration of the following update.

### How2Test
Have a slow server and monitor all vars for each loop then build a nice graph with gnuplot :)

### Todo / Checklist
May be more appropriate to have a mean or a median to compute WORLD_SLEEP_CONST rather to have a constant.

### DIFF
--- a/src/mangosd/WorldRunnable.cpp
+++ b/src/mangosd/WorldRunnable.cpp
@@ -42,33 +42,32 @@ void WorldRunnable::run()
     WorldDatabase.ThreadStart();                            // let thread do safe mySQL requests (one connection call enough)
     sWorld.InitResultQueue();
 
-    uint32 realCurrTime = 0;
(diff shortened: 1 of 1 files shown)


Lymphome:
As usual, just a proposal. This KISS code works great on my server with a quite big visibility distance and a smaller constant.

killerwife:
Sorry I havent looked at this sooner. Looks really good. I do not see any potential downsides, do you have any input on i?

PR #459, opened 2021-06-06
Avoid mobs chainpulling each other

shagu:
## 🍰 Pullrequest
Right now, pulling one mob leads to assistance calls to nearby mobs (which is correct).
However, it seems that all mobs who receive an assistance call, are calling others for assistance again.

This leads to pulling a massive amount of enemies, where rather only one group should be pulled.

Tested in Zul'Farrak stair event.

As seen in [this video](https://www.youtube.com/watch?v=DQJ_el1VbY0) (5/2007), only the enemies that are directly affected by the assistance call are added. 
The other ones are stay out of fight.

The issue was introduced via https://github.com/cmangos/mangos-tbc/commit/ee7174d82f217c630f5770e480d08ec54baafe82. I'm not entirely sure why this 
```cpp
for (Creature* receiver : receiverList)
```
assistance-call loop was added here as it does not seem to have existed before.

**Video:**
Current state of cmangos:
[Head](https://youtu.be/fC2vftdSPgg)

State after this patch is applied:
[Patched](https://youtu.be/OwtNlqIumvM)

### DIFF
--- a/src/game/AI/BaseAI/UnitAI.cpp
+++ b/src/game/AI/BaseAI/UnitAI.cpp
@@ -561,21 +561,8 @@ void UnitAI::SendAIEventAround(AIEventType eventType, Unit* invoker, uint32 dela
 
         if (!receiverList.empty())
         {
-            if (delay)
-            {
-                AiDelayEventAround* e = new AiDelayEventAround(eventType, invoker ? invoker->GetObjectGuid() : ObjectGuid(), *m_unit, receiverList, miscValue);
-                m_unit->m_events.AddEvent(e, m_unit->m_events.CalculateTime(delay));
-            }
-            else
-            {
-                for (Creature* receiver : receiverList)
-                {
-                    receiver->AI()->ReceiveAIEvent(eventType, m_unit, invoker, miscValue);
-                    // Special case for type 0 (call-assistance)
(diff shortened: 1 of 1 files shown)


shagu:
Closing this, due to lack of interest.

killerwife:
I havent even noticed this up until now. The issue with this, is that with the change you reverted I basically executed immediate 0 delay event sending immediately instead of postponing it until next map update. (because the queued event would execute in the next update loop). If we are having a recursive issue with this, we can perhaps enclose this in a guard of lets say, the variable changed in m_creature->SetNoCallAssistance(true); As such I do not know the immediate fix to this, just explaining the intent of the original commit.

And generally the turnaround on PRs is 1-2 weeks, there is much left to be desired during summer months sadly.

shagu:
I didn't see what the original commit was about to fix at all, I couldn't find any mob that is positivly affected by your change. All I noticed is that all humanoid groups are broken with that. So I thought reverting the broken part of your commit might be the best solution until a proper way is found to fix the original problem (whatever it was).

shagu:
Okay, understood. Since questing became impossible with your commit, I'll keep my revert. I'll look into this again as soon as I spot any boss with broken assist mechanic.

PR #464, opened 2021-06-27
[AQ40] Make bug trio boss unlootable when being consumed

insunaa:
## 🍰 Pullrequest

as the title says, while being consumed, the bosses are lootable for 3 seconds before they despawn, this shouldn't happen.

### Proof

- https://www.youtube.com/watch?v=uor-DHA0elE

### Issues

- fixes https://github.com/cmangos/issues/issues/2475

### How2Test

- As described in the bug

### Todo / Checklist

- [ ] Actually test this. Can't compile and run a server right now, so that'd be appreciated

PS: please excuse the misleading branch name, I wanted to fix something else first and noticed that it can't be done in SD2


cala:
Merged in https://github.com/cmangos/mangos-tbc/commit/327e86c40a2fa0f1ca803d273621a1c4dd154551
Thank you very much.

PR #468, opened 2021-07-01
[Naxx] Move Grobbulus to CombatAI

insunaa:
### Proof

- None

### Issues

- None

### How2Test

- Do the regular Grobbulus Fight

### Todo / Checklist

- [ ] Test everything, but especially slime stream. I wasn't sure how exactly it worked and as I said in the Gluth PR I can't currently test things, so all code changes are theoretical.


cala:
Merged in https://github.com/cmangos/mangos-tbc/commit/1de3ec53d030119bf1be5346b45a05cdc7765c54
Thank you very much.

I took the liberty to do minor changes like removing a sanity check about `m_instance` because the variable was never used or reworking the Slime Stream code that was not working as intended (the spell was continuously reapplied, which breaks its purpose because DoT is reset on each reappliance, leading to no damage being ever dealt).

PR #472, opened 2021-07-01
[Naxx] Move Noth to CombatAI

insunaa:
## 🍰 Pullrequest

Moves Noth to CombatAI

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [x] Testing. I currently can't test changes myself, so all code changes are theoretical.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/naxxramas/boss_noth.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/naxxramas/boss_noth.cpp
@@ -288,6 +288,7 @@ struct boss_nothAI : public CombatAI
             {
                 DoCastSpellIfCan(m_creature, SPELL_CURSE_PLAGUEBRINGER);
                 ResetCombatAction(action, 28u * IN_MILLISECONDS);
+                return;
             }
             case NOTH_BLINK:
             {
@@ -297,6 +298,7 @@ struct boss_nothAI : public CombatAI
                     DoResetThreat();
                     ResetCombatAction(action, 25u * IN_MILLISECONDS);
                 }
+                return;
             }
             case NOTH_SUMMON:
             {
@@ -334,6 +336,7 @@ struct boss_nothAI : public CombatAI
                     }
                     ++m_subWavesCount; // Keep track of which wave we currently are
                 }
+                return;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


cala:
@insunaa Thanks for this PR. I tested it and it appears that the waves are coming too quickly. I did not look at the code to understand why.

What do you prefer: I let you investigate and amend this or do I look for the issue by myself as you can't test by yourself ?

insunaa:
I'll take a look. Tho I actually don't really know why it would do that, since I used the timers of the original, but the timer/phase things are pretty complicated and it might make sense to rework these in general anyway.

insunaa:
OK, what I'm guessing might be happening is because the timer to summon skeletons is 30 seconds and the timer to teleport is 90 seconds that he summons the 3rd wave as he teleports because of a timer-race (when he shouldn't actually summon the adds), same when he teleports back down.
Is there any way to check the remaining time of another combataction while you are currently inside one?

cala:
> but the timer/phase things are pretty complicated and it might make sense to rework these in general anyway.

and yet, I spent a good amount of time trying to make them as simple as I could given the fact that there is close to no real periodicity in this encounter. :)

> Also blink apparently is not supposed to be a fixed 25 second timer, but rather somewhat random? Because at least deadly boss mods only says "possible blink in"

Some bosses have abilities on a periodic timer that still cannot trigger if the boss is somehow unable to cast (competitive casting or player interaction). That might be the case here, meaning that Blink should not be a triggered cast.

> Is there any way to check the remaining time of another combataction while you are currently inside one?

I don't think this is possible but I don't know enough all the `CombatAI` methods to be sure. I might also be possible that Noth is not a good candidate for `CombatAI` and should remain with `ScriptAI` instead.

Regarding the overall timers of the encounters, I set many of them using old Classic videos (and by old, I really mean old with blured quality at the best). So if you have sources (videos or sniff) based on recent Classic re-release that set a new light on these timers, I would gladly review this.

insunaa:
No, your timers seem on point. You're right might be a good idea to keep it with ScriptedAI, because as you said some timers are interdependent so it makes more sense to keep it as it is :)

insunaa:
Oof the issue was that I forgot to set `break;` after the combat actions >_>

cala:
It is possible that I still did not finish handling this PR? o_O

cala:
Closed (finally) in https://github.com/cmangos/mangos-tbc/commit/21318d3da44df5d737203b54e2283bdc3c77d8dd
Thank you for this PR. 👍🏻 

I took the liberty to do some small changes to make Clang happy, to use `AddOnKillText()` that was not yet implemented at the time you made the PR and to fix blink and curse actions that were still running during balcony phase (they shouldn't).

PR #473, opened 2021-07-08
Update missleading text in ExtractResources.sh

Nolin7777:
## 🍰 Pullrequest

See title.

### Proof

- None

### Issues

- None

### How2Test

- Run ExtractResources.sh to check the output.

### Todo / Checklist

- [X] None


cala:
Thank you very much. Though I fully agree with this PR, a bit of explanation would have been appreciated.

PR #478, opened 2021-08-28
Fix "Could not find configuration file"

Nolin7777:
### Issues

- Config files cannot found on world- and authserver startup

### How2Test

- Start Auth- or Worldserver


Nolin7777:
Fixed with https://github.com/cmangos/mangos-tbc/commit/1cf89e300a52fbedad0a357a509a2361199a18ef

PR #479, opened 2021-08-30
Compile fix for older builds

Nolin7777:
## 🍰 Pullrequest

Rather use the boost libary for the filesystem instead of std to make the Anti Cheat backwards compatible from C++17 to C++14.

Killerwife mentioned in cmangos' discord that he aims to make two seperate releases with VS 2021 coming, and boost is included anyways, so why don't we use it for now.?

Thanks for the linker hint @woblight


killerwife:
I am merging this until VS 2021 comes out.

PR #480, opened 2021-08-31
Install Warden modules

insunaa:
## 🍰 Pullrequest

Automatically installes the warden modules to the bin directory

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


esno:
could you please rebase your PR to a single commit?

insunaa:
done

insunaa:
ok, so the weird part is that technically this should already be handled in this
image

but it's not. or at least it's not working correctly

killerwife:
The problem is, it worked for me. We need to find where the cmake commands failed. I actually took time to test this.

insunaa:
image
this is all a regular `make install` does for me.

does it maybe require a new cmake flag? or something else?

killerwife:
I will be dealing with this in a few days when I will be working on anticheat on vanilla core. The cmake in place worked on windows but clearly did not on linux, I will need to check it out and make a test environment on my end.

PR #485, opened 2021-09-10
[Karazhan] Spotlight spawns in front of Barnes

insunaa:
## 🍰 Pullrequest

With this PR the spotlight spawns in/on top of barnes, instead of in front of him (see screenshot for the previous, wrong behavior)

image

### How2Test

- `.gm on`
- `.go cr Barnes`

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/karazhanScripts.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/karazhanScripts.cpp
@@ -149,7 +149,7 @@ struct npc_barnesAI : public npc_escortAI, private DialogueHelper
                         break;
                 }
                 SetEscortPaused(true);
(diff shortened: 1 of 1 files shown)


killerwife:
Make it spawn on these coords:
[16] Position: X: -10895.27 Y: -1782.626 Z: 90.55984
[16] Orientation: 3.769911

That will be futureproof.

insunaa:
Isn't spawning it directly on Barnes' coords also future-proof? Or is there a rework on the tempspawn system planned?

killerwife:
I checked the sniff of the event and my thought process is as follows. If the thingy would be spawned through spell, i.e. having CREATED_BY_SPELL unitfield then it would use that spell to spawn on his coords, however in sniff its a normal event spawn which leads me to believe its hardcoded in blizz db. So in reality we would want to spawn it like we spawn SWP with the new prototype, but thats not really necessary for one spawn, so this is a good concession with hardcoded XYZ.

PR #493, opened 2021-12-27
fix clang compiler warnings and minor logical error

insunaa:
## 🍰 Pullrequest

Fixing clang compiler warnings such as braces used as initializer. I understand why it was used in this case, because the surrounding code uses lists of enums to pass on, but since these function calls here are only for single creatures I think it makes sense to use the correct function call for single creatures

ReceiveAIEvent() usually takes an enum as function parameter, but that enum cannot contain 7166 as EventType, because that isn't defined in the AIEventType enum. Clang thus sees this comparison as always false, since the enum does not contain that value 7166 and so the parameter can also never be 7166 because it must be of type AIEventType, so clang would optimize that if statement out because at compiletime it always evaluates to false.

Minor datatype fixes.

Corrected typo in the PHASE/STATE enum used. (The original referred to an Ahn'Qiraji event and while it *could* have evaluated to the same uint32 it might also not, so I changed it to the explicit ScourgeInvasion State.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
@@ -146,7 +146,7 @@ void DespawnEventDoodads(Creature* shard)
         pDoodad->ForcedDespawn();
 
     std::list<Creature*> finderList;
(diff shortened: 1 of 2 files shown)


insunaa:
Fixed windows and gcc compile

PR #494, opened 2022-01-03
Add fixes to Invasions for the Scourge Invasion

insunaa:
## 🍰 Pullrequest

This PR fixes a few issues with the Zone Invasions, such as the Shadows of Doom never spawning, NPCs and Objects randomly respawning and the worldstate not being saved.

This PR requires https://github.com/cmangos/tbc-db/pull/1005

### Todo / Checklist

- [ ] Maybe find a less hacky way to make sure that the Necropolis Proxy NPC respawns mid-air without immediately falling to the ground
- [ ] Squash the commits once the OK is given
- [x] Remove superfluous includes with the squash


src/game/World/WorldState.cpp
```diff
@@ -1860,7 +1864,7 @@ void WorldState::HandleActiveZone(uint32 attackTimeVar, uint32 zoneId, uint32 re
 
     // Calculate the next possible attack between ZONE_ATTACK_TIMER_MIN and ZONE_ATTACK_TIMER_MAX.
     uint32 zoneAttackTimer = urand(ZONE_ATTACK_TIMER_MIN, ZONE_ATTACK_TIMER_MAX);
-    TimePoint next_attack = now + std::chrono::milliseconds(zoneAttackTimer);
+    TimePoint next_attack = now + std::chrono::milliseconds(zoneAttackTimer * 1000);
```
insunaa, line 1867: This should technically be
```cpp
TimePoint next_attack = now + std::chrono::seconds(zoneAttackTimer);
```
But I was getting some weird feedback in the "next attack in %d minutes" message, so I changed it back. Turns out that weird feedback is still present, so I guess this is superfluous

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
```diff
@@ -37,8 +37,8 @@ Unit* SelectRandomFlameshockerSpawnTarget(Creature* unit, Unit* except, float ra
 {
     std::list<Unit*> targets;
 
-    MaNGOS::AnyUnfriendlyUnitInObjectRangeCheck u_check(unit, radius);
-    MaNGOS::UnitListSearcher<MaNGOS::AnyUnfriendlyUnitInObjectRangeCheck> searcher(targets, u_check);
+    MaNGOS::AnyUnitInObjectRangeCheck u_check(unit, radius);
```
insunaa, line 40: UnfriendlyUnit has an LOS check, which we don't want because the summons are supposed to occur city-wide

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
```diff
@@ -945,35 +1055,58 @@ struct ScourgeMinion : public CombatAI
 struct PallidHorrorAI : public CombatAI
 {
     std::set<ObjectGuid> m_flameshockers;
+    std::unordered_map<ObjectGuid, uint32> m_flameshockFollowers;
 
     PallidHorrorAI(Creature* creature) : CombatAI(creature, 999)
     {
+        AddCustomAction(25, 2500u, [&]() // Hacky solution for flameshockers to remain in formation
```
insunaa, line 1062: Very hacky solution to keep the followers in a circle formation. Doesn't really work, too because their walk-speed is too low. I hope the `JoinCreatureGroup` function can take care of this in the future

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
```diff
@@ -945,35 +1055,58 @@ struct ScourgeMinion : public CombatAI
...
+                    m_flameshockFollowers.insert(std::pair<ObjectGuid, uint32>(pFlameshocker->GetObjectGuid(), i));
                 }
             }   
         }
         m_creature->SetCorpseDelay(10); // Corpse despawns 10 seconds after a crystal spawns.
         AddCombatAction(EVENT_PALLID_RANDOM_YELL, 5000u);
         AddCombatAction(EVENT_PALLID_SPELL_DAMAGE_VS_GUARDS, 5000u);
         AddCombatAction(EVENT_PALLID_SUMMON_FLAMESHOCKER, 5000u);
+        AddCombatAction(25, 5000u);
+        m_creature->GetMap()->SetWeather(m_creature->GetZoneId(), WEATHER_TYPE_STORM, 0.25f, true);
```
insunaa, line 1101: I'm not sure this is the right place to put it, but when the capitals get invaded the weather definitely changes to a very dark sky

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
```diff
@@ -1004,26 +1137,31 @@ struct PallidHorrorAI : public CombatAI
...
         uint32 cityAttackTimer = urand(CITY_ATTACK_TIMER_MIN, CITY_ATTACK_TIMER_MAX);
-        TimePoint nextAttack = now + std::chrono::milliseconds(cityAttackTimer);
-        uint64 timeToNextAttack = (nextAttack - now).count();
+        TimePoint nextAttack = now + std::chrono::seconds(cityAttackTimer);
+        uint64 timeToNextAttack = std::chrono::duration_cast<std::chrono::minutes>(nextAttack - now).count();
         SITimers index = m_creature->GetZoneId() == ZONEID_UNDERCITY_A ? SI_TIMER_UNDERCITY : SI_TIMER_STORMWIND;
         sWorldState.SetSITimer(index, nextAttack);
         sWorldState.SetPallidGuid(index, ObjectGuid());
-        sLog.outBasic("[Scourge Invasion Event] The Scourge has been defeated in %s, next attack starting in %d minutes", m_creature->GetZoneId() == ZONEID_UNDERCITY_A ? "Undercity" : "Stormwind", uint32(timeToNextAttack / 60));
+        m_creature->GetMap()->SetWeather(m_creature->GetZoneId(), WEATHER_TYPE_RAIN, 0.0f, false);
```
insunaa, line 1145: Same as above

src/game/World/WorldState.cpp
```diff
@@ -1753,8 +1757,8 @@ void WorldState::StartNewCityAttack(uint32 zoneId)
     if (m_siData.m_pendingPallids.find(zoneId) != m_siData.m_pendingPallids.end())
         return;
 
-    if (!m_siData.m_attackPoints[zoneId].pallidGuid.IsEmpty())
-        return;
+    /*if (!m_siData.m_attackPoints[zoneId].pallidGuid.IsEmpty())
```
insunaa, line 1760: I'm not sure what this is meant to do, but it was stopping City Invasions from happening, because after the first city invasion pallidGuid is *never* empty

insunaa:
I will stop changing this PR now until I get feedback, because I don't really know what else I would do. There's still the work of fixing the quest associations, but I don't know how and that's a database issue. I think SD2-wise the scourge invasion is feature complete now.

Feel free to test this anyone, and don't forget to use the accompanying modifications to the Database https://github.com/cmangos/tbc-db/pull/1005

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
```diff
@@ -469,7 +535,14 @@ struct NecropolisProxyAI : public ScriptedAI
             m_creature->ForcedDespawn();
     }
 
-    void UpdateAI(uint32 const diff) override {}
+    void UpdateAI(uint32 const diff) override
```
insunaa, line 538: I could make these CustomActions, but I don't know for what reason UpdateAI() was overridden to do nothing, so I kept it here instead. Might make sense to not run this every tick tho and instead use a timer.

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
al3xc1985: Maybe cyberium formation system cannhelp you now with this

al3xc1985:
> I will stop changing this PR now until I get feedback, because I don't really know what else I would do. There's still the work of fixing the quest associations, but I don't know how and that's a database issue. I think SD2-wise the scourge invasion is feature complete now.
> 
> Feel free to test this anyone, and don't forget to use the accompanying modifications to the Database [cmangos/tbc-db#1005](https://github.com/cmangos/tbc-db/pull/1005)

I am sure @killerwife will drop an eye on this as soon as he got time and give u a hand

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
```diff
@@ -362,6 +374,52 @@ struct NecropolisHealthAI : public ScriptedAI
 {
     NecropolisHealthAI(Creature* creature) : ScriptedAI(creature)
     {
+        AddCustomAction(0, 5000u, [&]()
+        {
+            if (m_creature->GetHealthPercent() > 99.f)
+            {
+                CreatureList proxies;
+                GetCreatureListWithEntryInGrid(proxies, m_creature, NPC_NECROPOLIS_PROXY, 200.f);
```
insunaa, line 382: I need to get better values for the distance. This value works, but I'm sure it can be optimized.

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
```diff
@@ -362,6 +374,52 @@ struct NecropolisHealthAI : public ScriptedAI
...
+                {
+                    if (proxy && (proxy->IsDespawned() || !proxy->IsAlive()))
+                    {
+                        proxy->SetRespawnTime(0);
+                        proxy->Respawn();
+                    }
+                }
+
+                GameObjectList circles;
+                GetGameObjectListWithEntryInGrid(circles, m_creature, GOBJ_SUMMON_CIRCLE, 1000.f);
```
insunaa, line 394: The range of 1000.f is probably too far, I will also need to measure what the distance between the furthest circle is to the necropolis, so we don't accidentally count circles that belong to other Necropoli. This particular part of the code is intended to be run after server restarts when an invasion is resumed. The number of defeated necropoli in an area is saved, but not the number of defeated Necrotic Shards, so if the Health Indicator for a necropolis is at 100% health it means it requires 3 necrotic shards to be defeated, if we spawned any less or some remain despawned because of the restart, this zone would not be defeatable

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
```diff
@@ -525,6 +605,18 @@ struct NecroticShard : public ScriptedAI
         {
             HandleShardMinionSpawnerSmall();
         });
+        AddCustomAction(11, 15000u, [&]() // Check if there are a summoning circle and a necropolis nearby, otherwise despawn
+        {
+            if (!m_creature)
+                return;
+            if (!GetClosestGameObjectWithEntry(m_creature, GOBJ_SUMMON_CIRCLE, 2.f) || !GetClosestCreatureWithEntry(m_creature, NPC_NECROPOLIS_HEALTH, 1000.f))
```
insunaa, line 612: Again need to find a suitable range instead of 1000.f
This code is here because of how "Battles won" work on server restart. The invasion resumes, and if one necropolis was defeated prior to restart but the others weren't, it means that now 1 or 2 necropoli, depending on the zone, will respawn in a random necropolis location. GameObjects like the summoning circle or creatures like the Necrotic Shard however may either still be despawned from that previous victory or remain spawned at a place that has no necropolis. This checks for this after 15 seconds to leave enough time to load all gameobjects nearby and then every 60 seconds. If no player is in a zone, gameobject load status is kinda undefined, so this is there to make sure that variance is managed

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
killerwife: Biggest grid search is 533, and I strongly advise against never using a value higher than 100. This needs to be done differently I am afraid.

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
killerwife: You do not need to fret about it that much. If server restarts, thats on the person restarting them. We just need to reload in a roughly equal state without forgetting wins. Server should generally not restart more than once a week with these events active, and the saving is mainly there to avoid detrimental effects of crashes.

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
insunaa: Do you have any suggestion that would work stateless?
The problem is that these GameObjects aren't summoned, but they're spawned from database, so I have no innate handle on them per necropolis. I can add these objects to a map or list when they spawn, but then I have no real way of associating them with a particular necropolis

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
insunaa: It's true that server restarts shouldn't be a major point of concern, but I don't really feel comfortable restarting a server into an undefined state (well even more undefined than usually) particularly because some of these states can leave the attack unwinnable, requiring manual intervention from GMs or above. Of course, that's their problem, but not handling this cleanly also reflects on us imo

insunaa:
Made some more drastic changes to the entire file so I put them in a separate branch
Are any of these changes acceptable?
https://github.com/insunaa/mangos-tbc/commit/4dfbc71b75038aaf3f4fb13f7f34fa0cdad537f8

src/game/AI/ScriptDevAI/scripts/world/scourge_invasion.cpp
insunaa: Can be reduced to 100.f

PR #495, opened 2022-01-06
Rewrite of Furbolg Script Event

insunaa:
## 🍰 Pullrequest

This PR rewrites the furbolg purification event from quest 4763 where you place a bowl of purified water on the ground to which the furbolgs run, drink and then after 5 seconds turn friendly and walk a way.

### Proof

- None

### Issues

- fixes https://github.com/cmangos/issues/issues/2584
I couldn't reproduce the crash that was originally reported in this issue, so it may also not fix the issue. There was some bad code in the original eventscript however, which under some circumstances might crash the server.

### How2Test

- .quest add 4763
- .additem 12341
- .additem 12342
- .additem 12343
- .additem 12347
- go cr 2168
- Find the bonfire and use the cleansing water bowl thing from your inventory

### ToDo
 - [ ] I could rewrite everything into CombatAI (with SpellLists), but I'm not sure if I have the patience

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
@@ -1000,7 +1000,7 @@ bool ProcessEventId_event_purify_food(uint32 /*eventId*/, Object* source, Object
         // Activate the purification sequence for all hostile furbolgs (prevent attracting already purified ones)
         for (auto& furbolg : furbolgs)
         {
-            if (furbolg->IsInCombat() || !furbolg->IsAlive() || !((Unit*)source)->IsEnemy(furbolg))
+            if (furbolg->IsInCombat() || !furbolg->IsAlive() || !furbolg->CanAttack((Unit*)source))
                 continue;
 
             if (auto* furbolgAI = dynamic_cast<npc_corrupted_furbolgAI*>(furbolg->AI()))
@@ -1022,7 +1022,7 @@ bool GOUse_go_furbolg_food(Player* player, GameObject* go)
     // Search for a nearby Blackwood Warrior and make it attack
     if (Creature* warrior = GetClosestCreatureWithEntry(go, NPC_BLACKWOOD_WARRIOR, 20.0f))
     {
-        if (!warrior->IsInCombat() && warrior->IsAlive() && player->IsEnemy(warrior))
+        if (!warrior->IsInCombat() && warrior->IsAlive() && warrior->CanAttack(player))
             warrior->AI()->AttackStart(player);
     }
     // Else spawn a new one (and make it attack)
--- a/src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
@@ -773,23 +773,24 @@ UnitAI* GetAI_npc_rabid_bear(Creature* pCreature)
 ######*/
 
 enum {
-    NPC_BLACKWOOD_WARRIOR = 2168,
-    NPC_BLACKWOOD_TOTEMIC = 2169,
+    NPC_BLACKWOOD_WARRIOR    = 2168,
+    NPC_BLACKWOOD_TOTEMIC    = 2169,
 
-    GO_PURIFIED_FOOD      = 175336,
+    GO_PURIFIED_FOOD         = 175336,
 
-    FACTION_BLACKWOOD     = 35,         // Faction guessed
+    FACTION_BLACKWOOD        = 35,         // Faction guessed
 
-    EMOTE_LURED           = -1010027,
-    EMOTE_PURIFIED        = -1010028,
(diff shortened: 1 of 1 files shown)


src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
```diff
@@ -1007,15 +1003,15 @@ bool GOUse_go_furbolg_food(Player* player, GameObject* go)
     // Search for a nearby Blackwood Warrior and make it attack
     if (Creature* warrior = GetClosestCreatureWithEntry(go, NPC_BLACKWOOD_WARRIOR, 20.0f))
     {
-        if (warrior->IsInCombat() || !warrior->IsAlive() || !player->IsEnemy(warrior))
+        if (!warrior->IsInCombat() && warrior->IsAlive() && player->IsEnemy(warrior))
```
insunaa, line 1006: My *guess* is that this is what caused the crash mentioned in the issue, since I couldn't reproduce it.
My speculation is that if the closest creature with that entry was dead&despawned that either it has no initialized AI() or AttackStart() fails at some point because the creature doesn't exist. Pure speculation tho.

insunaa:
Current version is 100% accurate as far as I can tell. The one thing that isn't accurate is something I don't know how to change (the demon that is summoned by the bowl is supposed to be summoned a few seconds earlier). It's hidden somewhere in the database, but everything I found in the DB about this quest was nebulous enough that I completely lost the thread, so if possible I'd request one of the DB gurus to take a quick look at it :)

Separate but related:  the food barrel things for this quest sparkle, but they shouldn't. No idea what flag causes this and how to disable it. If any DB expert is willing to take a look at it (they're supposed to be somewhat lootable, as in you get a castbar and then at the end of it you get an item, so the barrel thing needs to be clickable/usable, but it shouldn't sparkle)

I will not be pushing to this PR again so that it can be tested/evaluated in peace :)

AnonXS:
```
SELECT * FROM dbscripts_on_event where id IN (3938);
```
change delay from 20sec to whatever you think is right?

insunaa:
Thanks for looking into it for me!
After timing it precisely on the reference that also has a 20 second delay, and it also has a 5 second delay from the furbolgs reaching the bowl to becoming friendly. I think the confusion on my part came from the fact that on the reference the furbolgs don't immediately start running to the bowl when it appears, instead it takes a short but varying amount of time.

I'll add an additional commit that includes this functionality

src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
```diff
@@ -827,47 +865,30 @@ struct npc_corrupted_furbolgAI : public ScriptedAI
             return;
 
         m_creature->GetMotionMaster()->MoveIdle();
-        m_isMovementActive  = false;
-        m_purifyTimer = 5 * IN_MILLISECONDS;
+        float angle = m_creature->GetAngle(bowlCoords.GetPositionX(), bowlCoords.GetPositionY());
+        m_creature->SetOrientation(angle);
+        m_creature->SendHeartBeat();
```
insunaa, line 870: This might be unnecessary as they walk straight from where they were towards the nearest place on the bounding radius of the bowl, so they don't need to turn around. Sometimes Orientation gets weird in unexpected ways tho, so I'm not sure about it.
SendHeartBeat(); is required for some reason for the creature's orientation to be sent to clients/players, otherwise the orientation can be undefined clientside, even if well defined serverside

cala:
Assigning myself on this nice PR. I should review and merge it very soon.

src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
cala: That's a fair assumption and would it prove to be true, I would then be the culprit here. It seems that at some point I changed my mind about the logic when adding this safety check but that I did not changed the following code accordingly: asking a creature to attack when it is possibly dead or not an enemy of its target makes absolutely no sense.

killerwife: I suggest using CanAttack instead of IsEnemy, because the former is more than the latter.

insunaa: Done. The reason why IsEnemy was chosen was because the quest script temporarily turns them into a friendly faction, so IsEnemy() is less computationally intensive and achieves the desired effect

killerwife: I understand, but isenemy has a hidden caveat you prolly dont know. Factions arent symmetrical, I can be your enemy, but you can be my friend. CanAttack checks both in the correct manner. This was actually an issue with brutallus last year, because he was neutral for players, but players were hostile to him and some spells werent functioning properly until I adjusted it. All of these functions come directly from client and B basically uses CanAttack everywhere always.

cala:
Great job, thank you very much. 👍🏻

PR #497, opened 2022-01-21
MySQL 5.5/5.7 Support

MantisLord:
Ref https://github.com/cmangos/issues/issues/2620

I'm told these changes are all that's needed to support these lower versions, but I haven't confirmed it myself.

If that's the case and these are truly the only changes required, I don't see why we shouldn't support 5.5 & 5.7 still.


cyberium:
Thank you its correct, too bad it mean we will have to stop support for 5.7 in some futur

PR #501, opened 2022-02-04
Remap __FILE__ macro to relative paths

insunaa:
## 🍰 Pullrequest

This PR remaps the `__FILE__` macro to relative paths when using GCC 8.0 and up, or Clang 10.0 and up.
This should be a step towards reproducibility as mentioned in https://github.com/cmangos/issues/issues/1966

### Issues

- https://github.com/cmangos/issues/issues/1966


killerwife:
Only remark I have is that its also on windows.

insunaa:
__FILE__ cannot be changed in MSVC. I'm also not sure if it's required on Windows, because on Windows there's no real need to package CMaNGOS, except maybe for repacks but they can also just drop the uses of `__FILE__` and replace them with static strings.
Right now the only thing that's actually calling `__FILE__` is g3dlite and mangos assert for some weird reason. Even in Release Mode, when these things only really make sense in Debug mode. Technically we can remove those too and be rid of `__FILE__` altogether (in Release mode)

cyberium:
Better is to keep far from those very specific exception and find a more general answer to this. I think KW [approach](https://github.com/cmangos/issues/issues/1966#issuecomment-1024999797) is better on this. I see __FILE__ is used only on assertion code in different place from the core. Making it relative to source folder should fix that. And if its not modifiable on windows then it should not be used.

Thank you for this PR

insunaa:
Yeah I tried KW's approach and it just made compilation explode, hence my move to this approach.

PR #503, opened 2022-02-06
Add option for Link-Time Optimizations

insunaa:
## 🍰 Pullrequest

This PR adds the option to invoke Link-Time Optimizations through the CMake flag `-DEXPERIMENTAL_LTO`
LTO is especially useful for large project with lots of interaction between different source-files, which I think is very much the case with CMaNGOS. I haven't been able to get performance metering working as of yet, but I'm working on it and I'll update this PR if I have any notable results.

### Proof

- [Implementation and Evaluation of Link Time Optimization with libFirm (Paper by Mark Weinreuter)](https://pp.ipd.kit.edu/uploads/publikationen/weinreuter16bachelorarbeit.pdf)

### Issues

- None

### How2Test

- Recompile CMaNGOS with the CMake flag `-DEXPERIMENTAL_LTO=1`. Only works with GCC and Clang because MSVC doesn't see fit to implement LTO. Test performance metrics.

### Todo / Checklist

- [ ] Update the PR when I can finally get those dang metrics to work

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/black_temple/illidari_council.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/black_temple/illidari_council.cpp
@@ -373,7 +373,7 @@ struct boss_gathios_the_shattererAI : public boss_illidari_councilAI
     {
         AddCombatAction(GATHIOS_ACTION_JUDGEMENT, true);
         AddCombatAction(GATHIOS_ACTION_SEAL, 0u);
-        AddCombatAction(GATHIOS_ACTION_AURA, 90000, 90000);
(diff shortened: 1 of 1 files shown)


deiteris:
MSVC has LTO too.
https://docs.microsoft.com/en-us/cpp/build/reference/ltcg-link-time-code-generation?view=msvc-170

Or it probably also can enabled by `INTERPROCEDURAL_OPTIMIZATION` cmake flag for all compilers.
https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html

killerwife:
Can you test it using the cmake flag @insunaa ? If it works on your end, I will test it using MSVC

insunaa:
why is this closed?

PR #505, opened 2022-02-07
Fix alignment issues with clang optimizations on ARM32

insunaa:
## 🍰 Pullrequest

This PR allows CMaNGOS to be compiled with `clang` on a 32 bit ARM CPU.

### How2Test

- Clone this PR to a device with a 32-bit ARM CPU and compile cmangos with clang.

### Todo / Checklist

- [x] Confirm GCC compatibility

### DIFF
--- a/src/shared/Database/SQLStorageImpl.h
+++ b/src/shared/Database/SQLStorageImpl.h
@@ -27,14 +27,6 @@ template<class DerivedLoader, class StorageClass>
 template<class S, class D>                                  // S source-type, D destination-type
 void SQLStorageLoaderBase<DerivedLoader, StorageClass>::convert(uint32 /*field_pos*/, S src, D& dst)
 {
-#if defined(__arm__)
-    if (((unsigned) &dst) % sizeof(D)) {
-       //The address is not aligned. Use memcpy to avoid ARM unaligned trap
-       D converted(src);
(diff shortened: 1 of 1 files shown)


src/shared/Database/SQLStorageImpl.h
```diff
@@ -333,7 +425,7 @@ void SQLStorageLoaderBase<DerivedLoader, StorageClass>::Load(StorageClass& store
                 case FT_INT:    storeValue((uint32)fields[y].GetUInt32(), store, record, x, offset);      ++x; break;
                 case FT_FLOAT:  storeValue((float)fields[y].GetFloat(), store, record, x, offset);        ++x; break;
                 case FT_STRING: storeValue((char const*)fields[y].GetString(), store, record, x, offset); ++x; break;
-                case FT_64BITINT: storeValue(fields[y].GetUInt64(), store, record, x, offset);            ++x; break;
+                case FT_64BITINT: storeValue((uint64)fields[y].GetUInt64(), store, record, x, offset);            ++x; break;
```
killerwife, line 428: Why is this cast necessary if I might ask? I know it is harmful but still

insunaa: It wasn't necessary, I just didn't like that everything else gets cast and this doesn't. It's pretty much purely stylistic

PR #506, opened 2022-02-10
enable env based override of log file locations

mikehackworth:
## 🍰 Pullrequest
Allow for log locations to be overridden by environment variables when extracting resources. This is useful when using the optional output dir added in [this commit](https://github.com/cmangos/mangos-tbc/commit/1a421d936e5f4b812a96676902cb79aeec4629de) as it allows the log to be placed in the output dir, or elsewhere, rather than left in the /bin/tools dir in this scenario.

### How2Test
Default behaviour unchanged:

    ./ExtractResources.sh /client-dir /extracted-resources

But when you set LOG_FILE or DETAIL_LOG_FILE vars in your env before running script the logs will be output to their respective locations:

    LOG_FILE=/extracted-resources/extractor.log DETAIL_LOG_FILE=/extracted-resources/detailed.log \
    ./ExtractResources.sh /client-dir /extracted-resources


insunaa:
Are you sure behaviour will be unchanged? To me it looks like if the env variables aren't set the log-files will now start with a `-` in front of their name. What am I missing?

killerwife:
This is a bit of an obscure implementation of what you want to achieve, mainly because it requires previous knowledge of the script, maybe it would be better to pass another parameter "log directory" instead so it conforms with the other customizability?

I am in favor of specifying log dir but this way it kinda goes against how the rest of it works.

mikehackworth:
Hi @killerwife, thanks for taking the time to review my PR.

I would argue that currently the positional arguments also require knowledge of how the script works under the hood since there is no way to discover them. 

Take for example a scenario using some of the current available parameters:
`./ExtractResources.sh /path/to/client /path/to/extracted resources`

There is no way to know that these parameters are available, without previous knowledge, as far as I can tell. This is fine, since if someone is not going to look at the source of ExtractResources.sh they are probably following the steps from the [wiki](https://github.com/cmangos/issues/wiki/Installation-Instructions) which just cover using the script without any parameters.

Every additional positional parameter passed to the script adds additional complexity, since the script now has to handle the various allowed opts for each position see the if block in script [here](https://github.com/cmangos/mangos-tbc/blob/master/contrib/extractor_scripts/ExtractResources.sh#L83) for an example of what I mean. 

Allowing for variables to be overridden at runtime via environment variable adds very little extra complexity, and follows the way other popular tools such as Makefiles, Dockerfiles work.

killerwife:
Yes but the point is to be consistent. Either have everything using arguments or everything using environment variables, not both. I will leave this to @cyberium as he is the granddaddy of our scripts and he has the last say in this.

cyberium:
Thank you very much for this PR, however we prefer to use same output folder than the one provided in the argument if there is one.

PR #508, opened 2022-02-13
Fix Razorgore's summons calling a non-existent AI

insunaa:
### Issues

- fixes https://github.com/cmangos/issues/issues/2677

### Todo / Checklist

- [ ] Check what is causing the Anti-Cheat issues

### DIFF
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -2099,7 +2099,8 @@ void Creature::CallAssistance(Unit* enemy)
         float radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS);
         if (GetCreatureInfo()->CallForHelp > 0)
             radius = GetCreatureInfo()->CallForHelp;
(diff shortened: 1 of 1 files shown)


src/game/Entities/Creature.cpp
```diff
@@ -2099,7 +2099,8 @@ void Creature::CallAssistance(Unit* enemy)
         float radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS);
         if (GetCreatureInfo()->CallForHelp > 0)
             radius = GetCreatureInfo()->CallForHelp;
-        AI()->SendAIEventAround(AI_EVENT_CALL_ASSISTANCE, target, sWorld.getConfig(CONFIG_UINT32_CREATURE_FAMILY_ASSISTANCE_DELAY), radius);
+        if(AI())
+            AI()->SendAIEventAround(AI_EVENT_CALL_ASSISTANCE, target, sWorld.getConfig(CONFIG_UINT32_CREATURE_FAMILY_ASSISTANCE_DELAY), radius);
```
insunaa, line 2103: This PR fixes the issue of course, but more interesting is the question: Why does the creature summoned by the instance not have an AI yet when CallAssistance() is called?
I couldn't figure it out, so this is the only way I could think of fixing it

PR #511, opened 2022-02-23
Adjust Out of Combat Health Regeneration for Player controlled npcs

AnonXS:
## 🍰 Pullrequest
This targets to decrease player controlled mob health regeneration out of combat, they by default regenerate like a normal mob that is evading home in 33% ticks

This makes enslaveable mobs highly abuseable as they regenerate very fast after every drop of combat.

Searing Infernal 6073 level 29 with maxhealth 950 regens 123hp every 5secs ooc

### Proof
Based on https://mega.nz/file/CnQkxZAL#5lSze4bAZK-iJUmx4NQGb-9_ILK_vH_phnexHv3G4OI - classic ptr

grafik
grafik
grafik

### Issues

- Tick Timer should 5secs, not 2secs as currently

### DIFF
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -810,7 +810,7 @@ void Creature::RegenerateAll(uint32 update_diff)
 
     RegeneratePower(2.f);
 
-    m_regenTimer = REGEN_TIME_FULL;
+    m_regenTimer = REGEN_TIME_FULL_UNIT;
 }
 
 void Creature::RegeneratePower(float timerMultiplier)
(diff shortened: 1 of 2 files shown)


Nytanath:
> Tick Timer should 5secs, not 2secs as currently

Tangentially related to this: on this video the enslaved demon regens health 3 seconds after evade, then every 5 seconds after that.

I tested that behavior on normal creatures and got the same: first tick after 3 secs, then after 5 secs until full health.

The current 2 seconds timer seems wrong, then.

https://user-images.githubusercontent.com/99603810/155373581-e4f77554-ecfc-4386-8ad7-80f52932118d.mp4

AnonXS:
Question is also, when they evade back home and regen hp, do when do they become attackable again? as soon as they do random movement at home location?

Nytanath:
From what I noticed, creatures start to regen hp as soon as they begin evading, and become attackable as soon as they exit evade state (so on home reached).

AnonXS:
Ok so only thing wrong now is our initial tick timer, which is the same concurrent 5secs, which should start ticking on evade start.

Nytanath:
Yeah, first regen tick should happen 3 seconds after evade, then every 5 seconds until full health.

PR #512, opened 2022-02-25
[SD] Make ChoRush friendly and Sit on Dungeon Create & King Gordok done

AnonXS:
## 🍰 Pullrequest
Make ChoRush Sit on Dungeon Create when King Gordok event is done

### Proof
https://youtu.be/RnQ7BEswhts?t=1262

### Issues
Else you can run out and back in, and he will be attackable again, eventhough only King Gordok was killed that "id"

### How2Test
.go c id 14324
kill King Gordok
leave dungeon
.go c id 14324
he is attackable again and does not sit


insunaa:
He should sit, but he should not be friendly.
He should not be attackable however.  When you exit DM north and enter again all ogres are hostile(ish) again, but many can't be attacked. there are exceptions: The summoned demons as well as the demonic eyes will remain attackable always. Same with respawning ogres, they will also be attackable for people who re-enter (but not for people who stayed inside, for those they should all be friendly)

PR #514, opened 2022-03-04
[Spell] Add script for s.8555 Left for Dead -> s.8359

AnonXS:
Add support for smartest spell ever. https://tbc.wowhead.com/npc=4422/agathelos-the-raging#abilities;mode:normal

## 🍰 Pullrequest
Adds SPELL_ATTR_TRIGGERS_A_SPELL component of the spell 8555

### Proof
https://www.wowhead.com/npc=4422/agathelos-the-raging#comments:id=1071607
https://wowwiki-archive.fandom.com/wiki/Agathelos_the_Raging

### How2Test
.go c id 4422 with two clients, watch one getting feign deathed, its seemingly not even a debuff so you can just break it, but not by pushing normal spells/abilities.

### DIFF
--- a/sql/scriptdev2/spell.sql
+++ b/sql/scriptdev2/spell.sql
@@ -27,6 +27,7 @@ INSERT INTO spell_scripts(Id, ScriptName) VALUES
 (4132,'spell_banish_exile'),
 (6467,'spell_unarmed_woodcutter'),
 (7054,'spell_forsaken_skill'),
+(8555,'spell_left_for_dead'),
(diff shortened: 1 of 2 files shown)


al3xc1985:
well, this is a nice catch and something I NEVER seen fixed on ANY private server

AnonXS:
Do you know by chance when this is actually used? currently he uses it on a non tank as soon as he gets in range to do it.

From comments it seems he is able to reset the encounter if tryed to get soloed.

PR #515, opened 2022-03-10
[SD] Add OnStealthAlert Check for Klaven Mortwake 7053

AnonXS:
## 🍰 Pullrequest
Remove ACID script which does not work as intended

https://github.com/cmangos/tbc-db/commit/f3b45ce0b04a5eeabd1fa32d8812ffe167b03973

### Proof

https://www.youtube.com/watch?v=d_2-8ql5pZE&t=65s

### Issues
- NA

Edit: What i found to be pretty frustrating is that detect range is pretty much the same as detect + aggro range, so maybe there is still something wrong with detection formulars.

#2 Npcs dont reset to their default orientation after a set amount of time after doing a stealth detection. (probably linked to why patrol groups dont all continue walking after they were distracted currently)

### How2Test
.go c id 7053


killerwife:
Had to adjust the new script function to the new style. https://github.com/cmangos/mangos-tbc/commit/0974b26626b08e0c502c41d8aeaed67ff25ef3fa

PR #517, opened 2022-03-11
Update min/max values of SKILL_RANGE_MONO skills on character levelup

Nytanath:
## 🍰 Pullrequest

Currently, SKILL_RANGE_MONO have their min/max values updated when a character logs in. This is correct. However, they are not updated when the character levels up, which is incorrect.

Valid for Vanilla, TBC and WotLK.

### Proof

- None

### Issues

- None

### How2Test

- Create a Draenei.
- Check the tooltip for Gift of the Naaru, see that it mentions 50 health points.
- .levelup 10
- Check the tooltip, see that it still mentions 50 health points.
- This is fixed when logging out and back in, but it should be updated on levelup too.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -5783,7 +5783,8 @@ void Player::UpdateSkillsForLevel(bool maximize/* = false*/)
         if (!pSkill)
             continue;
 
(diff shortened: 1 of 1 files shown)


killerwife:
This should be applicable to vanilla as well I think.

Nytanath:
Yep, and WotLK too.

PR #521, opened 2022-03-18
Port breadcrumb quest system from TrinityCore

Nytanath:
## 🍰 Pullrequest

Port of the core-side part of https://github.com/TrinityCore/TrinityCore/commit/5ed77113b637c49e99a610b00f385c51da67c3bf. The DB side will take a bit longer, but the functionality itself works.

### Proof

- None

### Issues

https://github.com/cmangos/issues/issues/2745 (does not close it yet)

### How2Test

```sql
-- Rude Awakening (363) -> The Mindless Ones (364)
UPDATE `quest_template` SET `BreadcrumbForQuestId`=364 WHERE `entry`=363;
UPDATE `quest_template` SET `PrevQuestId`=0 WHERE `entry`=364;

```

Create an Undead.
- .go creature id 1569
- Spawn creature 1568 nearby, for ease of access.
- Accept quest from Shadow Priest Sarvis, see that you can't take Undertaker Mordo's quest. Do the opposite thing and see you can't accept the other's quest. Notice that Undertaker Mordo's quest is entirely optional. They behave as a breadcrumb and target quest should do.

### DIFF
--- a/sql/updates/mangos/s2444_01_mangos_quest_template.sql
+++ b/sql/updates/mangos/s2444_01_mangos_quest_template.sql
@@ -1,0 +1,2 @@
+ALTER TABLE `quest_template` ADD COLUMN `BreadcrumbForQuestId` MEDIUMINT(9) UNSIGNED DEFAULT 0 NOT NULL AFTER `ExclusiveGroup`; 
+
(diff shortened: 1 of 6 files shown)


killerwife:
Yeah, I am aware this is doable but in retrospect, its very annoying to add a condition when there is a ton of quests in the breadcrumb list. In wotlk there are several where 8 breadcrumbs exist for same quest. Imagine addition conditions for that.

Nytanath:
It's kind of like ExclusiveGroup - a functionality that could be replicated via conditions, but is common enough that it's easier to handle in a more... graceful? way.

PR #522, opened 2022-03-23
Add .clang-format file

insunaa:
## 🍰 Pullrequest

This PR adds a .clang-format file. IDEs which integrate clang-format can use this to automatically format their code to CMaNGOS code-style as defined by https://github.com/cmangos/issues/wiki/Coding-Standards

### Todo / Checklist

- [ ] Make sure that code-style is completely coherent

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,178 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Microsoft
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignArrayOfStructures: Left
+AlignConsecutiveMacros: AcrossEmptyLinesAndComments
+AlignConsecutiveAssignments: AcrossEmptyLinesAndComments
+AlignConsecutiveBitFields: AcrossEmptyLinesAndComments
+AlignConsecutiveDeclarations: None
+AlignEscapedNewlines: Right
+AlignOperands:   Align
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortEnumsOnASingleLine: false
+AllowShortBlocksOnASingleLine: Always
+AllowShortCaseLabelsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: All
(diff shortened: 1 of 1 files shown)


insunaa:
This format file uses the line-length of 80 prescribed by the style-guide, but imo this creates pretty cramped looking code. Imo a higher value, like 120 should be used

Fabi:
> This format file uses the line-length of 80 prescribed by the style-guide, but imo this creates pretty cramped looking code. Imo a higher value, like 120 should be used

more like 180, 220 or 240

insunaa:
My editor with a pretty small font is ~ 230 columns wide, I'd be on board with 180, but I think more than that would be a bit overkill and then there might as well be no line-length restriction

insunaa:
https://gist.github.com/insunaa/e6d9f505877f314e9c5c329dbec3c536 this Gist has diff files between the original `Creature.cpp` and `Creature.cpp` treated with clang-format at 80, 120 and 180 line-length

killerwife:
My personal rule of thumb is 120 but i have made exceptions where I have made up to 180 cos of code simplicity over readability. So I would go with 180, because we should already add line breaks for sake of readability.

killerwife:
And also we have a verticality problem in many files, one of the reasons i dont like stacking constructor member init one per line. Especially in scripts.

insunaa:
There has been no further discussion happening. Do we want a standardized format file or not? If so I would request we merge this PR. If not let me know so I can clean the cluttered PR page.

killerwife:
@cyberium please check it and merge it. @Warlockbugs maybe speak your opinion here.

Warlockbugs:
Hopefully, this wont cause enforced formatting on entire file on any change made to that file. That would make submitting and merging PRs miserable, as well complicate git history quite a bit.

insunaa:
It does not *enforce* formatting, but if you select the whole file and use the auto-formatting feature it will obviously format the whole file. Generally you're supposed to select whatever you want to format and then use the auto-format feature for only that area. I don't know if some IDEs enforce formatting the entire file, tho.

insunaa:
I don't know if the syntax is even still valid, but the discussion about this addition should not just stay dead, imo.
Maybe we should also add a `.clang-tidy` file, which allows automatic highlighting of various problems, such as going against the Cpp core guidelines

PR #526, opened 2022-04-04
Implement ITEM_EXTRA_IGNORE_QUEST_STATUS to allow items that start a quest to drop for players even if they have the related quest already accepted/completed/rewarded

Nytanath:
## 🍰 Pullrequest

This is my attempt to fix [Arena Master](https://tbc.wowhead.com/item=18706/arena-master) and items like that (although the number of items with this behavior I know of, other than this, is 0) without having to add a specific case for this item directly in core.

Items that start a quest, and have ITEM_EXTRA_IGNORE_QUEST_STATUS, will be lootable regardless of the related quest's status.

I had a hard time trying to find alternative ways to fix this, because this item is special in that, as long as you "win" the event (by opening the chest) you can obtain as many of this trinket as you want. But there's also a quest (the next step of the quest started by the trinket) that requires 12 of those.

I removed the ```!player->HasQuestForItem(itemId)``` part because it only affects this item (git blame bought me to the SVN import, but I'm 99% sure this was only added with this item in mind), but it's not relevant for this change so I can put it back if anyone wants it.

### Proof

- None

### Issues

- None

### How2Test

- .gobj add 179697
- Loot the Arena Master trinket.
- Right-click it, accept the quest.
- .gobj add 179697
- Open the chest, notice that there's no Arena Master trinket in the new chest.

Execute this query:
```sql
UPDATE `item_template` SET `ExtraFlags`=`ExtraFlags`|2 WHERE `entry`=18706;

```

Repeat above steps, notice how now you're able to loot the trinket again.

### DIFF
--- a/src/game/Entities/ItemPrototype.h
+++ b/src/game/Entities/ItemPrototype.h
@@ -448,8 +448,9 @@ inline uint8 ItemSubClassToDurabilityMultiplierId(uint32 ItemClass, uint32 ItemS
 
 enum ItemExtraFlags
 {
-    ITEM_EXTRA_REAL_TIME_DURATION = 0x01,                   // if set and have Duration time, then offline time included in counting, if not set then counted only in game time
-    ITEM_EXTRA_ALL                = 0x01                    // all used flags, used for check DB data (mask all above flags)
+    ITEM_EXTRA_REAL_TIME_DURATION  = 0x01, // if set and have Duration time, then offline time included in counting, if not set then counted only in game time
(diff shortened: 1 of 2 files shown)


killerwife:
Name it like here to keep consistent with at least vmangos https://github.com/vmangos/core/commit/a5655fee95deb52e6379eac3b288486d9b005df9#diff-d7ee65bf53881bf43393460d7969c3cfcf783efa14b2f7771497b234c12dca78R397

killerwife:
TC shauren identical name almost, vmangos brotalnia identical name almost, you do you :D

killerwife:
Hmmm, have you checked there are no quests which require item starting quest items also in wotlk? Like requires lets say 2 instead of just one.

PR #533, opened 2022-04-22
Disable MANGOS_ASSERT when defining NDEBUG

insunaa:
## 🍰 Pullrequest

This PR disables the MANGOS_ASSERT macro when a release build is selected and `NDEBUG` is defined.

Asserts are a debugging tool and have no place in release builds https://en.wikipedia.org/wiki/Assertion_(software_development)#Assertions_in_production_environment

### Issues

- fixes https://github.com/cmangos/issues/issues/1966

### How2Test

- use `cmake .. -DCMAKE_BUILD_TYPE=Release -DNDEBUG` when building

### DIFF
--- a/src/game/AI/ScriptDevAI/system/ScriptLoader.cpp
+++ b/src/game/AI/ScriptDevAI/system/ScriptLoader.cpp
@@ -371,7 +371,7 @@ void AddScripts()
     AddSC_battleground();
 
     // custom
-
+MANGOS_ASSERT(false);
     // examples
     AddSC_example_creature();
     AddSC_example_escort();
--- a/src/game/AI/ScriptDevAI/system/ScriptLoader.cpp
+++ b/src/game/AI/ScriptDevAI/system/ScriptLoader.cpp
@@ -371,7 +371,7 @@ void AddScripts()
     AddSC_battleground();
 
     // custom
-MANGOS_ASSERT(false);
+
     // examples
     AddSC_example_creature();
     AddSC_example_escort();
--- a/src/shared/Errors.h
+++ b/src/shared/Errors.h
@@ -21,27 +21,23 @@
 
 #include "Common.h"
 
-#ifndef NDEBUG
 // Normal assert.
 #  define WPError(CONDITION) \
 if (!(CONDITION)) \
 { \
-    printf("%s:%i: Error: Assertion in %s failed: %s\n", \
-        __FILE__, __LINE__, __FUNCTION__, STRINGIZE(CONDITION)); \
-    fflush(stdout); \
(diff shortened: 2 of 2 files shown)


killerwife:
Well, that would require us to restructure the whole codebase to deal differently with unrecoverable errors.

There are two major development styles:

A - you encounter an unrecoverable situation that should never occur and you bring down situation immediately

B - you try to repair it with exceptions or different approaches.

Either way, we use asserts all over the core in situations that should never happen, so disabling it you will be potentially corrupting your core and pushing incorrect db data at some point. This is further highlighted by the fact that we added like 15 asserts in the last 5 years. Hiding it will not do anyone any good. Do you have a situation where the asserts frequently crash the core?

insunaa:
No, but the underlying `assert()` is already disabled by NDEBUG anyway...
If we really want to do what you say we should use `abort()` instead...

for reference, this is where `assert()` is defined in `<cassert>`
```cpp
#ifdef NDEBUG
#define assert(condition) ((void)0)
#else
#define assert(condition) /*implementation defined*/
#endif
```

insunaa:
I updated the PR so that now both of our goals are met.
With the code as it is in my PR we are no longer using `__FILE__` in release builds, but now we're also actually halting execution when a `MANGOS_ASSERT()` evaluates to false.

killerwife:
C++ exceptions are not used due to performance reasons. We use them inside networking code in a completely valid way but even there most is handled through error codes instead. Read this for reference: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2544r0.html

The assert thing, @cyberium please read through this and resolve this if you can :) I do not know enough about the topic.

insunaa:
The point of Exceptions is to not need them. I'm not saying we should start inserting *throws* everywhere for no reason, that's pointless. What I mean is rather coding with exceptions in mind / exception proof coding. The fact that we have all of this assert-stuff going on in production code means that for some reason impossible scenarios can occur in our code. writing code exception-safe would mean that every object we create or function we call operates in such a way that in case of a failure it doesn't actually change the state of anything else and keeps running as if it was never called (while logging an error ofc)

Things like this, to prevent issues like the Razorgore issue where we were trying to set an AI for summoned creatures into combat before the AI was initialized...

killerwife:
Well that one occurs because we are doing shenanigans that alter global state within constructor, which is a no-no in C++ to begin with (virtual functions dont work in constructor). And that could lead to so many side effects not even a seasoned programmer wouldnt know where to look.

insunaa:
Please re-evaluate this PR.
It now actually does what you thought it did, when it didn't.

PR #534, opened 2022-04-24
Update The Blackwood Corrupted script to CombatAI

insunaa:
## 🍰 Pullrequest

This PR updates the Blackwood Corrupted quest script to CombatAI and removes use of a global variable.

### Issues

- fixes https://github.com/cmangos/issues/issues/2887

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
@@ -21,6 +21,7 @@ SDComment: Quest support: 731, 945, 994, 995, 2078, 2118, 4763, 5321
 SDCategory: Darkshore
 EndScriptData */
 
+#include "AI/ScriptDevAI/base/CombatAI.h"
 #include "AI/ScriptDevAI/include/sc_common.h"
 #include "Grids/GridNotifiers.h"
 #include "Grids/CellImpl.h"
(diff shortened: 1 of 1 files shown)


src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
```diff
@@ -831,28 +840,49 @@ struct npc_corrupted_furbolgAI : public ScriptedAI
...
+        AddCombatAction(ACTION_HEALING_WARD, true);
     }
 
     void Reset() override
     {
+        CombatAI::Reset();
+        
+        switch (m_creature->GetEntry())
+        {
+            case NPC_BLACKWOOD_WARRIOR: ResetIfNotStarted(ACTION_THUNDERCLAP, GetSubsequentActionTimer(ACTION_THUNDERCLAP)); DoCastSpellIfCan(m_creature, SPELL_BATTLE_STANCE); break;
```
killerwife, line 861: Instead of doing this, and doing AddCombatAction(ACTION_THUNDERCLAP, true);

do AddCombatAction(ACTION_THUNDERCLAP, initialMin, initialMax); and each reset CombatAI::Reset will do it for you :) I can change it when merging but making sure you know.

insunaa: Wouldn't both types of furbolg have the same spellset?
Only totemics are supposed to use healing wards and only warriors are supposed to use thunderclap

killerwife: Well if you add a condition to only call add for one or the other then they wont have both.

killerwife: Well if you wanna be extra modern, add the spells to creature_spell_list :D And you dont have this problem to begin with.

PR #536, opened 2022-05-07
Implement PTR Paymaster ScriptDev

insunaa:
## 🍰 Pullrequest

This PR implements the ScriptDev script for the PTR Paymaster NPC.

### Todo / Checklist

- [ ] Add SQL Part

```sql
DELETE FROM `npc_text_broadcast_text` WHERE `Id` IN (39288);
INSERT INTO `npc_text_broadcast_text` (`Id`, `Prob0`, `BroadcastTextId0`) VALUES
(39288, 1, 39288);

UPDATE `creature_template` SET `GossipMenuId`=26751 WHERE `Entry`=26075;
UPDATE `creature_template` SET `ScriptName`="npc_paymaster" WHERE `Entry`=26075;
```


src/game/AI/ScriptDevAI/scripts/world/npcs_special.cpp
```diff
@@ -2805,6 +2805,206 @@ bool GossipHello_npc_gossip_npc(Player* player, Creature* creature)
...
+            if (pPlayer->GetFreePrimaryProfessionPoints() > 0)
+            {
+                pPlayer->learnSpellHighRank(28897);
+                pPlayer->SetSkill(755, 375, 375);
+                LearnSkillRecipesHelper(pPlayer, 755);
+            }
+            pPlayer->CLOSE_GOSSIP_MENU();
+            return true;
+        case GOSSIP_ACTION_INFO_DEF + 6:
+            if (pPlayer->GetFreePrimaryProfessionPoints() > 0)
```
killerwife, line 2912: Shouldnt this check be instead around actually showing the options rather than selecting them?

src/game/AI/ScriptDevAI/scripts/world/npcs_special.cpp
```diff
@@ -2805,6 +2805,206 @@ bool GossipHello_npc_gossip_npc(Player* player, Creature* creature)
     return true;
 }
 
+struct PTRPaymasterAI : public ScriptedAI
```
killerwife, line 2808: This seems useless.

insunaa: Might be. I'm not sure if I can have the gossip functions hooked without an AI. If I can then I can throw this out

src/game/AI/ScriptDevAI/scripts/world/npcs_special.cpp
insunaa: Logically yes, on the reference however the options always show and just don't allow you to learn more professions if you already have 2

miraco:
After learning an Profession skill you also learn all of the recipts including sunwell. Is this wanted?
Current PTR only has T6 items (no ZA no Sunwell) and afaik on PTR you only learn the profession spells for example Alchemy(Master) and get 375 skill points.

insunaa:
> After learning an Profession skill you also learn all of the recipts including sunwell. Is this wanted? Current PTR only has T6 items (no ZA no Sunwell) and afaik on PTR you only learn the profession spells for example Alchemy(Master) and get 375 skill points.

Learning Sunwell recipes is not intended, on PTR you only learn up to BT recipes, but I don't know how to exclude sunwell recipes aside from using a static list of ids.

miraco:
on what ptr you get all recipes from paymaster?
Or is my char buged? This video is from current tbc ptr.
https://youtu.be/slgw10MNx_I 
im fine with having all recipes up to bt and i think a static list is still better then having all sunwell recipes

insunaa:
You're correct, I am now also only receiving the profession and no recipes. Either they changed this in the meantime or I misremember this, tho I did record it on video. I'll check that when I get home

src/game/AI/ScriptDevAI/scripts/world/npcs_special.cpp
```diff
@@ -2820,6 +2820,110 @@ bool GossipHello_npc_gossip_npc(Player* player, Creature* creature)
     return true;
 }
 
+bool GossipHello_npc_paymaster(Player* pPlayer, Creature* pCreature)
```
killerwife, line 2823: I do not understand why the hello part is not in DB.

insunaa: I don't know if there are broadcast texts for it etc.

killerwife:
Ok I will sum up merging the PR. Basically, we have broadcast texts that handle all translations and if we add it like it is right now, the translations wont be handled by this npc. Whenever I can be bothered, or before you do, we need to remove the hello part  and move it to db and the select part needs to be somehow better integrated into using broadcast texts. The PR as-is is on par with old stuff but there is no reason why we cant make a better future for this clearly needed custom paradigm in 2022. And since you added BossAI which is exactly a step in that direction I bet you will agree with me. I will sit on it and think about how this can be better integrated.

PR #540, opened 2022-05-25
Update auras when applying temporary enchantments

insunaa:
## 🍰 Pullrequest

This PR updates the status of all Auras for the client when a temporary enchantment occurs.
This will cause the `UNIT_AURA` API-Event to fire on the client, notifying all Addons that a change has taken place and that there are values that can be updated.

### How2Test

- Cast `Windfury Weapon` repeatedly
- use `/etrace start`
- Cast `Windfury Weapon` repeatedly, again
- Notice that the `UNIT_AURA` API-Event fires.

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -11817,6 +11817,9 @@ void Player::ApplyEnchantment(Item* item, EnchantmentSlot slot, bool apply, bool
                         }
                         else
                             RemoveAurasDueToItemSpell(item, enchant_spell_id);
+
+                        if (SpellAuraHolder* holder = GetSpellAuraHolder(enchant_spell_id))
+                            holder->UpdateAuraDuration();
                     }
                     break;
                 }
--- a/src/game/Spells/SpellEffects.cpp
+++ b/src/game/Spells/SpellEffects.cpp
@@ -5267,6 +5267,7 @@ void Spell::EffectEnchantItemTmp(SpellEffectIndex eff_idx)
 
     // add new enchanting if equipped
     item_owner->ApplyEnchantment(itemTarget, TEMP_ENCHANTMENT_SLOT, true);
+    item_owner->UpdateClientAuras();
 }
 
 void Spell::EffectTameCreature(SpellEffectIndex /*eff_idx*/)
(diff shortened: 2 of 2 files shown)


killerwife:
Why not notify about only the unit aura again and not about all of them, you are basically spamming client with possibly hundreds of packets.

insunaa:
Because temporary enchantments don't come with Auras, as far as I can tell. (At least I haven't been able to extract a spellauraholder in any part of the process of applying a temporary enchantment)

killerwife:
Incorrect, SpellItemEnchantment.dbc, 283 - Windfury 1 - spell Id 33757

insunaa:
OK? So how do I get to that spell/aura ID from the `m_spellInfo`?
I don't want to hard-code the spell IDs if I don't need to

killerwife:
We already apply them on enchant apply, Player::ApplyEnchantment check ITEM_ENCHANTMENT_TYPE_EQUIP_SPELL

insunaa:
added a change that should do what you suggest, but it doesn't work

killerwife:
Well then compare sniff with packetlog, and see whats being sent, debug packetlog 1, and parse it with wpp

killerwife:
And set packetlog in mangosd conf

insunaa:
There's nothing being sent because the enchant_spell_id is not the spell id that causes the aura you linked earlier, so it doesn't find an auraholder for it

insunaa:
nvm that was incorrect, it *is* being sent, but it doesn't cause the correct client response. let me check

insunaa:
ah yes, I'm dumb. this here is stopping it from being sent:
image

insunaa:
can't get it to work even if I remove this requirement entirely. the aura is both passive and outside of the max aura slot.
The packetlog config doesn't seem to work at all, but from what I gather the aura duration update is still either never received or discarded for some reason.
In any case the `UNIT_AURA` event never triggers for Windfury enchants for example

killerwife:
TBC parsing isnt perfect, you can easily fix the parsing in WPP and make a pull request there as you need to investigate some issue. Do investigate it more :P

PR #541, opened 2022-06-04
Make Kalecgos stand up when any spell is cast on him

insunaa:
## 🍰 Pullrequest

This PR makes sure Kalecgos leaves his sleep state when any spell is cast on him

### Proof

- [Kalecgos standing with only hunter's mark](https://www.youtube.com/watch?v=y2q76zQP6r0)
- [Kalecgos standing up with hunter's mark before aggro](https://youtu.be/yvsT58mOSiQ?t=15)
- [Kalecgos standing pre-pull](https://www.youtube.com/watch?v=PQS2QimEU_A)
- [Kalecgos also standing pre-pull](https://www.youtube.com/watch?v=KfQqrK3KoZE)
- [Kalecgos standing up with hunter's mark before aggro](https://www.youtube.com/watch?v=YSNxDeyXmlo)

### How2Test

- Make a hunter
- Level Hunter to 70
- Go into SWP
- Stand before Kalecgos
- Cast Hunter's Mark


insunaa:
Further investigation has just shown that Kalecgos does not stand up from Mind Vision on Classic servers, so it doesn't seem to be every spell. Doing further testing.

insunaa:
OK, further investigation reveals that Hunter's Mark always makes enemies stand up. I'll change the location of this to the Hunter.cpp spellaura scripts

insunaa:
This PR is accurate to the original, but comes at a minor performance cost, even when disabled in config. Please evaluate so we can get rid of a few PRs

henhouse:
Does this need to be a config option? I can't imagine this being custom enough to want to disable.

insunaa:
It doesn't need to be, but KW hates it so this was the compromise

insunaa:
I'll remove the config option then to reduce the workload for this PR

PR #553, opened 2022-07-04
Fix GitHub daily release not building with extractors

insunaa:
## 🍰 Pullrequest

This PR makes sure that the Extractors are once again included in the `with-all` Release zip


cyberium:
Thank you merged ;)

PR #543, opened 2022-06-05
Modernize script for wailing caverns escort

miraco:
## 🍰 Pullrequest
Use waypoint_path instead of script_waypoint
use broadcast_text instead of script_text
correct spawns and let creatures move to escort npc
adjust some timers and rp scripts
add some todos

### Proof

Based on TBC PTR sniffs 
https://www.youtube.com/watch?v=QmDV7ald1RU&feature=youtu.be
### Issues

https://github.com/cmangos/issues/issues/2833
It should now always be doable to kill endboss.
### How2Test

- None

### Todo / Checklist

Only some cosmetics that i have in my head. 
- Deviate Moccasine and Nightmare ectoplasm waves have multiple spawn and waypoints (will have to do it multiple times on retail to get all points)
- Nightmare ectoplasm seems to be in formation (spawn) groups after spawn (https://youtu.be/QmDV7ald1RU?t=388)
- After killing the boss, both npcs transfor into birds and fly to instance entrance, could get some cosmetic updates

PR #556, opened 2022-07-09
[SD] Modernize script for wailing caverns escort

AnonXS:
Use waypoint_path instead of script_waypoint
use broadcast_text instead of script_text
correct spawns and let creatures move to escort npc
adjust some timers and rp scripts
add some todos

## 🍰 Pullrequest
https://github.com/cmangos/mangos-tbc/pull/543
https://github.com/cmangos/tbc-db/tree/wailingcaverns -> https://github.com/cmangos/tbc-db/pull/1079


miraco:
Just as an info for my self so i dont forget it

started looking at this again and updating it with new spawn_group techniques and did a new test on ptr cause i wanted to know what happens with the spawned npcs when they killed the escort npc.

- noticed that the escort continues on wait timer not when spawned npcs are dead
- spawned npcs keep doing their waypoints from spawn point near escort npc

AnonXS:
Except for endboss event which cant start without him being alive? all the "wave mobs" still spawn?

PR #558, opened 2022-07-11
Fix riding skill scaling with level

insunaa:
## 🍰 Pullrequest

This PR stops the riding skill from increasing when leveling up

### How2Test

- Learn riding at the trainer
- use `.level`
- notice that the riding skill does not change.

### DIFF
--- a/src/game/Entities/Player.cpp
+++ b/src/game/Entities/Player.cpp
@@ -21055,7 +21055,7 @@ void Player::_LoadSkills(QueryResult* result)
                 if (!(entry->classMask & classMask))
                     continue;
 
-                if (entry->flags & SKILL_FLAG_MAXIMIZED)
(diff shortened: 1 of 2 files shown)


killerwife:
Ok I am a bit slow but I had an epiphany. It is correct that maximized maxes out lvl but riding max is skill stage max not level bound. So your fix is correct for riding but the overarching issue implies there is supposed to be a query for how the skill scales in that maximized condition. Will try to resolve today.

MilkmanMHOC:
Just finally lvled in Classic and this also occurs there as well.  Every lvl, 5 more skill points in riding.

killerwife:
I will push stuff into all cores but i do it in batches. For now your can cherry pick the commit yourself.

MilkmanMHOC:
yes, just thought i would let you know since this was all labeled TBC from what i could tell.  Thanks again!

PR #562, opened 2022-07-14
Add SPAWN_GROUP_FORMATION_OPTION_DONT_FOLLOW_LEADER

miraco:
## 🍰 Pullrequest

This adds an simple formation option to spawn_groups preventing groups getting in SpawnGroupFormationType.

### Proof

179057202-3431c521-cbda-4484-8e06-31c96631704c

https://youtu.be/l_bPZ-7Fkms?t=472

This group for example stands like this after respawning, before they start moving as SPAWN_GROUP_FORMATION_TYPE_SINGLE_FILE.

We can then use dbscript SCRIPT_COMMAND_SPAWN_GROUP (51) to change formation options before they start moving.

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Maps/SpawnGroup.cpp
+++ b/src/game/Maps/SpawnGroup.cpp
@@ -1258,6 +1258,9 @@ void FormationData::FixSlotsPositions()
         ++totalMembers;
     }
 
+    if (HaveOption(SPAWN_GROUP_FORMATION_OPTION_DONT_FOLLOW_LEADER))
+        return; 
+
     if (totalMembers <= 1)
         return;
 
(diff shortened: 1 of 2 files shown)


killerwife:
I believe what anon is saying should actually be the case. There are many npcs like that, who even do so mid formation script. Perhaps @cyberium can answer that per what his design is for this in mind. I am sure @Tobschinski can explain that to him :P

cyberium:
No that is wrong. Its what i tried to explain many time to you. Movement is not static for any objects in the world. If a creature need some movement in the better world it should be assigned by a script. That way we would be able to assign any movement to any object.

So as soon as we assign a formation to a group it will behave like a formation, to stop that behaviour we have to remove that group from that formation. If we accept this PR it will introduce some more complexity by having some creature in formation without formation behaviour.

The only correct way for me is to assign the formation by a spawn, script that will wait for x second then add formation to that group.
At some point we will have to stop all of those static link that we are doing in db that should have be dynamic from the start.
By dynamic i mean controlled by script.

killerwife:
Ok then how do you propose to resolve this one. I have no problem with it either way, you know more than I do.

cyberium:
I forgot to say a big thank you @miraco for this PR 👍 

Will check alternatives when i'll be on my desktop.

miraco:
> I believe what anon is saying should actually be the case. There are many npcs like that, who even do so mid formation script. Perhaps @cyberium can answer that per what his design is for this in mind. I am sure @Tobschinski can explain that to him :P

Yeah so far I noticed it mostly on spawn, and when reaching a waypoint.

PR #571, opened 2022-09-06
[Spell]Curse of Tongues now has a 12 second duration when used on PvP Targets.

miraco:
## 🍰 Pullrequest

Adds 12 seconds duration to Curse of Tongues when used on Players.

### Proof

[Patch 2.1:  ](https://wowwiki-archive.fandom.com/wiki/Patch_2.1.0#Warlocks)

> Curse of Tongues now has a 12 second duration when used on PvP targets.

https://wowwiki-archive.fandom.com/wiki/Curse_of_Tongues?oldid=1428403
### Issues

- None

### How2Test

- None

### DIFF
--- a/sql/scriptdev2/spell.sql
+++ b/sql/scriptdev2/spell.sql
@@ -642,7 +642,9 @@ INSERT INTO spell_scripts(Id, ScriptName) VALUES
 (19734,'spell_devour_magic'),
 (19736,'spell_devour_magic'),
 (27276,'spell_devour_magic'),
-(27277,'spell_devour_magic');
+(27277,'spell_devour_magic'),
+(1714,'spell_curse_of_tongues'),
+(11719,'spell_curse_of_tongues');
 
 -- Pet Scaling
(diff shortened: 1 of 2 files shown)


killerwife:
This wont work because it needs to work in tandem with diminishing returns. Unit::ApplyDiminishingToDuration is the function for it. Not scriptable in a nice script atm.

miraco:
How should this apply then?
if i remember correct dr on pvp was 10s, 5s, 2,5s (100%, 50%, 75%)

How would this look on 12s? 12s->6s->3s ?

Does dr even affect curse of tongues?

killerwife:
I was just told blizz has a GetDuration spellscript hook. I will add it once I figure out a sensible place to put it. Basically you will be able to override the duration for spell if in pvp and it will respect all calc.

AnonXS:
https://github.com/cmangos/mangos-wotlk/commit/efa16a06e23f6cc5d38b5c425c7e2e82b7ddc5d4

PR #573, opened 2022-09-17
Remove GM aura if not GM on log in

henhouse:
## 🍰 Pullrequest
Discovered an issue where if a player is no longer a GM and logs in, they will have the permanent transparency debuff. This checks, on log in, if they're supposed to still have it or not.

### Proof

- None

### Issues

- None

### How2Test

- Log in as GM, get transparency debuff
- Logout and remove GM access
- Log in again, and notice you still have the debuff

### Todo / Checklist

- [X] None


killerwife:
Perhaps we should instead never save it to db? Cos if the config changes, then it will not work. You can add an ID to IsSaveToDbHolder.

PR #574, opened 2022-09-17
Fix Devour Magic self cast on debuffs

henhouse:
## 🍰 Pullrequest
Fixes https://github.com/cmangos/issues/issues/3064.

tl;dr: Devour Magic not possible to self-cast or on friendly ally, but is supposed to be. Check was wrong.

### Proof

- See [spell tooltip](https://tbc.wowhead.com/spell=27277/devour-magic)

### Issues

- [3064](https://github.com/cmangos/issues/issues/3064)

### How2Test

- Reproduction steps in issue linked above.

### Todo / Checklist

- [X] None


killerwife:
CanAssist is the check you are looking for rather than IsFriend

PR #582, opened 2022-11-29
Fix large GOs visibility override map visibility

celguar:
## 🍰 Pullrequest
Fix "Large" Gameobjects visibility set to Large (200y) on maps with higher visibility, e.g. Battlegrounds

### Issues
- "Large" Gameobjects in Battleground (e.g. Flagstands) are set to 200y and ignore that map visibility is higher

### How2Test
- Go away from flag in BG to >200y, it disappears. Everything else, e.g. players, is still visible from 500+y

### DIFF
--- a/src/game/Entities/GameObject.cpp
+++ b/src/game/Entities/GameObject.cpp
@@ -292,8 +292,8 @@ bool GameObject::Create(uint32 dbGuid, uint32 guidlow, uint32 name_id, Map* map,
     if (InstanceData* iData = map->GetInstanceData())
         iData->OnObjectCreate(this);
 
-    // Check if GameObject is Large
-    if (GetGOInfo()->IsLargeGameObject())
+    // Check if GameObject is Large, skip if map has same or better visibility (e.g. Battleground)
+    if (GetGOInfo()->IsLargeGameObject() && GetVisibilityData().GetVisibilityDistance() < VISIBILITY_DISTANCE_LARGE)
         GetVisibilityData().SetVisibilityDistanceOverride(VisibilityDistanceType::Large);
 
     if (GetEntry() == 187039) // Smuggled Mana Cell - only GO in phase in TBC


celguar:
https://youtu.be/arsMCx9Topw?t=218
I'm pretty sure players and flag in the background is not 100y, not 200y but more

celguar:
if looking at player capturing the flag in AB and not seeing the flag itself or its "beam effect" (ally/horde/contested) is blizzlike then sure is fine. But I have some doubts

celguar:
in vmangos Visibility.Distance.BG             = 180,
cmangos is 533
What is correct, I dunno

Kittnz:
Do some AV's on wrath classic for example, you can see the end bosses from far away, even flags. But if you keep your target on someone and let him move far away you will see him disapear. 533 is just grid size, i don't know what the correct view distance is for players really, but it's limited.

celguar:
This PR is not about players, it's about fixing the issue on cmangos that on maps with visibility distance set to > 200y, for example all battlegrounds in cmangos, every object except for GOs flagged with large is visible properly at max distance. That is clearly not intended that flag large should work as limiter in that scenario

Kittnz:
i know this is weird on cmangos to have a huge range for everything and have these go's like flags and what not only have 200y visibility, however the visibily data for these objects never change. Having 533 on cmangos for everything is not correct, only the objects or npcs that have a specific visibility range closest to the player should see them. Not everything from the map should be seen by a player. Range for the players is set (prolly 200y) in a battleground, and normal change for objects should be 100y except for specific gobs (flags) or npcs (bosses) that have large flag set.

celguar:
In vmangos these gameobjects are always set ton 533 visibility btw
https://github.com/vmangos/core/blob/e776319fc455e68eb707bc197bfb42f49f235276/src/game/Objects/GameObjectDefines.h#L656

celguar:
I'm not trying to fix blizzlike ranges, this PR is to fix "weird on cmangos to have a huge range for everything and have these go's like flags and what not only have 200y visibility" and only that specific thing. If I set my visibility on my server to 500 or whatever else I expect it to work. Default is 533 for BGs in cmangos and it should work. Otherwise these configs are ambiguous or not fully working.

Kittnz:
summoning @killerwife

al3xc1985:
U can't do a normal summon for @killerwife, u need to use a ritual :P

PR #584, opened 2022-11-30
Fix VMapManager multi-threading issues.

evil-at-wow:
## 🍰 Pullrequest
I started fixing more data races we have in the core(s) because of API being (mis)used from multiple threads. Because these changes are not as straightforward as the handful of changes I've been pushing lately, I prefer to run them through a PR for review and additional testing by others before I merge/push them.

This PR is about data races/crashes I've run into in VMapFactor and the VMAP manager. For details, please see the commit description.

### DIFF
--- a/src/game/vmap/VMapFactory.cpp
+++ b/src/game/vmap/VMapFactory.cpp
@@ -18,8 +18,16 @@
 
 #include "VMapFactory.h"
 #include "VMapManager2.h"
+#include <atomic>
+#include <mutex>
 
 using namespace G3D;
+
+namespace
+{
+    std::mutex gVMapMutex;
+    std::atomic<VMAP::IVMapManager*> gVMapManager { nullptr };
+}
 
 namespace VMAP
 {
@@ -51,8 +59,6 @@ namespace VMAP
         }
     }
 
-    std::mutex m_vmapMutex;
-    IVMapManager* gVMapManager = nullptr;
 
     //===============================================
     // result false, if no more id are found
@@ -79,26 +85,32 @@ namespace VMAP
         return result;
     }
 
-    //===============================================
-    // just return the instance
-    IVMapManager* VMapFactory::createOrGetVMapManager()
+    /*
+     * Return the instance.
+     */
+    IVMapManager& VMapFactory::GetVMapManager()
     {
-        if (!gVMapManager)
+        IVMapManager* instance = gVMapManager.load(std::memory_order_acquire);
+        if (!instance)
         {
-            std::lock_guard<std::mutex> lock(m_vmapMutex);
-            if (!gVMapManager)
-                gVMapManager = new VMapManager2();              // should be taken from config ... Please change if you like :-)
+            std::lock_guard<std::mutex> lock(gVMapMutex);
+            instance = gVMapManager.load(std::memory_order_relaxed);
+            if (!instance)
+            {
+                instance = new VMapManager2();
+                gVMapManager.store(instance, std::memory_order_release);
+            }
         }
-        return gVMapManager;
+        return *instance;
     }
 
-    //===============================================
-    // delete all internal data structures
+    /*
+     * Delete all internal data structures.
+     */
     void VMapFactory::clear()
     {
-        std::lock_guard<std::mutex> lock(m_vmapMutex);
-
-        delete gVMapManager;
-        gVMapManager = nullptr;
+        // No lock required because no other thread should still be
+        // accessing the VMapManager when we're deleting it!
+        delete gVMapManager.load();
     }
 }
--- a/src/game/vmap/VMapManager2.h
+++ b/src/game/vmap/VMapManager2.h
@@ -66,7 +66,14 @@ namespace VMAP
     class VMapManager2 : public IVMapManager
     {
         private:
-            std::mutex m_vmStaticMapMutex;
+            StaticMapTree* GetInstanceTree(uint32 mapId) const
+            {
+                std::lock_guard<std::mutex> lock(m_vmStaticMapMutex);
+                auto iterator = iInstanceMapTrees.find(mapId);
+                return (iterator != iInstanceMapTrees.end() ? iterator->second : nullptr);
+            }
+
+            mutable std::mutex m_vmStaticMapMutex;
             std::mutex m_vmModelMutex;
 
         protected:
--- a/src/game/Maps/GridMap.cpp
+++ b/src/game/Maps/GridMap.cpp
@@ -24,6 +24,7 @@
 #include "Server/DBCStores.h"
 #include "Maps/GridMap.h"
 #include "VMapFactory.h"
+#include "IVMapManager.h"
 #include "MotionGenerators/MoveMap.h"
 #include "World/World.h"
 #include "Policies/Singleton.h"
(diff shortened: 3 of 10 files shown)


src/game/vmap/VMapFactory.cpp
```diff
@@ -79,26 +85,32 @@ namespace VMAP
         return result;
     }
 
-    //===============================================
-    // just return the instance
-    IVMapManager* VMapFactory::createOrGetVMapManager()
+    /*
+     * Return the instance.
+     */
+    IVMapManager& VMapFactory::GetVMapManager()
```
killerwife, line 91: This is more of a philosophical remark, rather than having issue with your code. I wonder if we arent't shooting ourselves in the foot with these dynamic loading procedures and whether we should not just load it on app start in a synchronous manner. This is because if I remember correctly, MMAP code has a similar possible race condition. Just food for thought.

src/game/vmap/VMapManager2.h
```diff
@@ -66,7 +66,14 @@ namespace VMAP
     class VMapManager2 : public IVMapManager
     {
         private:
-            std::mutex m_vmStaticMapMutex;
+            StaticMapTree* GetInstanceTree(uint32 mapId) const
```
killerwife, line 69: I am not entirely sure how quick are mutexes in a nonblocking scenario, but I feel like this will absolutely obliterate performance. Liquid and area checks are being computed very often (which is incorrect, blizz caches them and TC has it prototyped now and its also visible on official servers) and this will prolly call the mutex hundreds of  times. I do however see where you are going with this. The "real" fix I feel like would be to change the ownership of vmaps and mmaps (which I reckon have the same exact issue) to Map ownership. Essentially in  the current threading model there is World context and Map context, and I have been slowly moving some reloadable DB data that could be vulnerable to data races to Map data storage (more memory consumption but safe).  This would however be a massive change and even with this performance impact here, its better to have this safeguard and document it well than not have it. I would suggest trying to compare all grids loaded 0 1 530 maps before and after this change to see the impact on perf and at least know what we are getting ourselves into.

src/game/vmap/VMapFactory.cpp
evil-at-wow: You're correct, I think MMap code has similar issues indeed. It's also on my radar for a future change...

src/game/vmap/VMapManager2.h
evil-at-wow: As far as I've measured and seen (on Linux and Windows, I'm not sure about other operating systems but I have no reason to believe they're off much worse), mutexes aren't really that slow. Unless - there's always a but, right? - there's a lot of contention. Which of course opens a debate on what's a lot of contention and if we're in such a case here, and so on. I don't think this will or should obliterate performance, because we're also not exactly (un)loading parts of maps all the time. Sure, it happens regularly, but it's not like that is the majority of the work the core does.

That said, it's obviously not helping performance either, so if we can reasonably avoid the mutexes that would be a bonus. For that reason I have considered other options as well, and I'm glad to see you've come to the same conclusion as I have: a better fix would be to have vmaps/mmaps per map indeed. It makes total sense when you think about it: the ideal situation is not to share any data between threads, simply because not sharing data means there's no need for synchronization (atomics, locks, etc, you name it). So we certainly agree on the long term solution, but when I started going that route... well, as you say, it seems a pretty big change indeed and I ran into some things that made it less straightforward for me than I hoped it would be (for example, the vmap manager is also doing stuff with (World)Models, and these can be shared between maps (things like braziers or whatever that are present on multiple maps) so that again required synchronization, etc). Long story short, I went for the more straightforward short term solution for now, also because there's about a dozen other issues on my radar that I'd like to tackle first. But I'd certainly be willing to work towards the longer term solution at one point... :smiley: 

Also, fair point about trying to assess the impact. I'll try to measure the impact of this on loading grids on continents and report back.

killerwife: Yep, thats fair. Also I am having a crash for like a year now that originates in vmap code and happens way more frequently on wotlk core, so its entirely possible its due to this code here (or the mmap one). So I am definitely up for merging this without issue, just perhaps after merging creating an issue on the issue repo where the ideal long term solution is detailed in short. Since we don't delete issues willy nilly its a perfect way of documenting it without forgetting about it.

evil-at-wow: In the meantime I did some basic tests. What I did was locally adjust ```MapManager::CreateContinents()``` to have some timing around the map creation, as follows:

```
    ...
    auto start = std::chrono::high_resolution_clock::now();
    for (auto id : continents)
    {
        // code not modified
    }

    for (auto& futurItr : futures)
        futurItr.wait();
    auto end = std::chrono::high_resolution_clock::now();
    std::cout << "Time to create continents = " << std::chrono::duration_cast<std::chrono::microseconds>(end - start).count() << "µs.\n";
    ...
```

I built the server in release mode of course, and then started the server 6 times - configured with fairly default server settings, MMAPs enabled - both without and with this PR, and gathered the resulting timings. The first run was considered a 'warming up' run to guarantee all needed files would be cached in memory and its result was discarded, leaving results of 5 runs. This was done on a Xeon E5-1620v4, a ~6 year old CPU but with decent clock speeds, to see what results I'd get on an average CPU these days (my assumption being that more modern CPUs won't be worse, but also that most people won't be running the server on a state-of-the-art 64-core EPYC either). This is what I got:

|         | Linux, without PR | Linux, with PR | Windows 10, without PR | Windows 10, with PR  |
| --- | ---: | ---: | ---: | ---: |
| run 1 | 220293 µs | 332594 µs | 512925 µs | 512118 µs |
| run 2 | 220969 µs | 263592 µs | 524910 µs | 538870 µs |
| run 3 | 235373 µs | 211585 µs | 514172 µs | 542832 µs |
| run 4 | 219793 µs | 315052 µs | 512783 µs | 571856 µs |
| run 5 | 215032 µs | 216665 µs | 590699 µs | 527100 µs |
| Avg | 222292 µs | 267897.6 µs | 531097.8 µs | 538555.2 µs |

I can't explain why the Linux results are varying much more compared to the Windows results (ironically, the host runs Linux and Windows 10 runs in a VM). I'm also not sure yet why ```CreateContinents()``` takes much longer on Windows (file I/O is typically slower in a VM, but there was like no disk access after the 'warming up' run caused everything to be cached in memory).

So purely going with these averages, we could say this PR makes the Windows build about 1.5% slower, and the Linux build about 20% slower. But as can be seen in the table, even with this PR some Linux timings are faster than the timings without the PR, and results seem to vary widely anyway regardless of this PR.

And even if one is willing to accept this being 20% slower, we're talking about something that takes less than a quarter of a second here to load 3 full continents, so I doubt we'll see any noticable effect when loading or unloading a single grid now and then...

killerwife: Silly q, did you build linux with -o2?

evil-at-wow: No, I used the default CMake Release target, which uses -O3, as expected I would add. I used the defaults because I don't think many people override compiler flags when they build the server and the point of the exercise was trying to figure out what people would get after this PR.

I don't think -O3 or -O2 is explaining the result, nor do I really believe this PR is making that part on Linux 20% slower. The main issue here is that I got too much variation in the timings, making it hard to draw conclusions IMHO. I'll have to profile to see if I can understand what's going on.

killerwife: Alright, i was just throwing ideas at wall to see what sticks :D

killerwife:
Reminder that you prolly want to merge this yourself if it has marinated in your head enough @evil-at-wow :D

evil-at-wow:
Thanks for reminding me, although it was still on my list and I didn't forget. I was still thinking about looking into the unexpected timing variations first (see discussion).

In addition, I started a little effort/experiment with MMaps because, as you rightfully remarked before, these have a similar issue, and it seems like a nice idea to have them handled by the map. Depending on the outcome of that experiment I might change course here and do something similar for the VMaps...

I'll mark it as draft for now, we can still decide later to merge this after all.

xhyanhao9:
The extraction of MMAP files is incomplete, only 000.mmap and 001.mmap. How can we solve this problem? thanks！

killerwife:
@evil-at-wow resolved it by asking around after forgetting about it. TC guys had it solved without locks, so I opted for that solution after reviewing our usage.

killerwife:
MMAPs still to go.

evil-at-wow:
I didn't forget about it, it just happened to end up low on a long priority list (which I admit is practically the same for the outside world) :wink: 

Mostly because I've been using this patch and a similar patch for MMAPs for like two years now, so my problem was kind of fixed and not the most urgent any more. We all know how it goes...

killerwife:
I absolutely do :D Either way, I have a plan for mmaps as well, so it will finally be sorted out.

PR #585, opened 2022-12-03
Ice Barrier fix in SWP

maikale:
UPDATE `gameobject_template` SET `data0`='0' WHERE  `entry`= 188119;

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/sunwell_plateau/sunwell_plateau.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/sunwell_plateau/sunwell_plateau.cpp
@@ -277,7 +277,11 @@ void instance_sunwell_plateau::OnObjectCreate(GameObject* go)
         case GO_FORCEFIELD:
         case GO_BOSS_COLLISION_1:
         case GO_BOSS_COLLISION_2:
+            break;
         case GO_ICE_BARRIER:
+            if (m_auiEncounter[TYPE_BRUTALLUS] == IN_PROGRESS)
+                go->GetVisibilityData().SetVisibilityDistanceOverride(VisibilityDistanceType::Gigantic);
+            go->SetGoState(GO_STATE_ACTIVE);
             break;
         case GO_FIRE_BARRIER:
             if (m_auiEncounter[TYPE_FELMYST] == DONE)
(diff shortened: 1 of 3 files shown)


Grz3s:
official data:
```code
ServerToClient: SMSG_QUERY_GAME_OBJECT_RESPONSE (0x005F) Number: 331
Entry: 188119
Type: 0 (Door)
Display ID: 7704
[0] Name: Doodad_Sunwell_Ice_Barrier01
[1] Name: 
[2] Name: 
[3] Name: 
Icon Name: 
Cast Caption: 
Unk String: 
[0] Data: 1
[1] Data: 0
[2] Data: 0
[3] Data: 0
[4] Data: 0
[5] Data: 0
[6] Data: 0
[7] Data: 0
[8] Data: 0
[9] Data: 0
[10] Data: 0
[11] Data: 0
[12] Data: 0
[13] Data: 0
[14] Data: 0
[15] Data: 0
[16] Data: 0
[17] Data: 0
[18] Data: 0
[19] Data: 0
[20] Data: 0
[21] Data: 0
[22] Data: 0
[23] Data: 0
Size: 1
[0] Quest Item: 0 (0)
[1] Quest Item: 0 (0)
[2] Quest Item: 0 (0)
[3] Quest Item: 0 (0)
[4] Quest Item: 0 (0)
[5] Quest Item: 0 (0)
```

maikale:
I don't think it's right to change the data ->  data0='1'

killerwife:
I will explain the thought process to everyone considering there was some discussion on discord.

The issue are STATIONARY creatures and (stationary, but thats all of them) gameobjects. Creatures that move, broadcast their position in creature->player direction, which means that mobs like Fel reaver (who has motion) broadcasts his position but Doom Lord Kazzak (who is stationary) does not despite being the same visibility distance wise.

There is only one solution really: Broadcast stationary creature/go position for those who have visibility > default (100yd in world maps).

The fix will need to be twofold:  
1. fix infinite mobs to be added immediately to visibility lists on player map join or creature map join
2. fix all other mobs to periodically (1 second for large, 2 second for gigantic) broadcast their positions to players

Currently we broadcast visibility on each server tick, this is not only resource intensive but also wrong. As such, EVEN if we do not fix it this way, we need to do an intermediary fix that does not prevent the final correct state.

Should someone ask about this, please show them this explanation.

PR #586, opened 2022-12-16
Improve ByteBuffer API.

evil-at-wow:
## 🍰 Pullrequest

Here's another change I'll move through review first before pushing it.

See the commit for all details. The main motivation here is that this is kind of a no-brainer for me in a C++11 and later world, and that a movable ByteBuffer will, after a next almost trivial change, also make WorldPacket movable (which may open a few other opportunities). There may even be other improvements where we now (implicitly) copy ByteBuffers where this may help, but I have no examples in mind.

A somewhat unrelated question: would you prefer I keep doing this for changes like this, that are hopefully not controversial and are just making better use of C++, but use some of your time to review it, or are you like "you can commit and you seem to know what you're doing in C++, so just push it"?

### DIFF
--- a/src/shared/ByteBuffer.h
+++ b/src/shared/ByteBuffer.h
@@ -49,22 +49,38 @@ struct Unused
 class ByteBuffer
 {
     public:
-        const static size_t DEFAULT_SIZE = 0x1000;
 
-        // constructor
-        ByteBuffer(): _rpos(0), _wpos(0)
+        explicit ByteBuffer(size_t reservedSize = s_defaultSize): _rpos(0), _wpos(0)
         {
-            _storage.reserve(DEFAULT_SIZE);
+            _storage.reserve(reservedSize);
         }
 
-        // constructor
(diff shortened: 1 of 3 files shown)


src/shared/ByteBuffer.h
```diff
@@ -436,9 +452,10 @@ class ByteBuffer
             append((uint8*)&value, sizeof(value));
         }
 
-    protected:
         size_t _rpos, _wpos;
         std::vector<uint8> _storage;
+
+        static constexpr size_t defaultSize = 0x1000;
```
evil-at-wow, line 458: Code style question: I know we use ```m_someThing``` for members (at least in new code), and of course technically this is a member of the class as well. Now I personally find it rather handy to name statics differently so that you easily realize you're not just messing with that one object but potentially with all of them at the same time. Would people mind introducing something like, say, ```s_someThing``` for statics? Like we already use ```g_someThing``` in a few places for (typically) global stuff?

killerwife: I am down with s_

PR #587, opened 2022-12-16
Fix wrong AuthSocket idle detection timer.

evil-at-wow:
## 🍰 Pullrequest
I stumbled upon this somewhat unfortunate implementation that kind of defeats the purpose of having a timeout in the first place. Details are in the commit.

One question that remains, and which I didn't address at all, is whether we only want a timeout for the first client command? So after my change we close sockets if the client takes more than 30s to send a first reply, but we cancel the timer rather than reset it (see the ```m_timeoutTimer.cancel()``` in ```AuthSocket::ProcessIncomingData()```) which, well, cancels any asynchronous operations waiting on the timer. So technically speaking after the first client reply we don't bother checking any more...

### How2Test
* Start a server, and in particular realmd, putting a breakpoint in the timeout handler (or adding some output or whatever).
* Wait at least 30s before any client logs in.
* Notice how the accepting socket's timeout expires before a client even attempted to make a connection...


al3xc1985:
This fixes an ancient bug indeed!

PR #590, opened 2023-01-05
prevent gizletone  spawn_group going into normal movement on point 19…

miraco:
… when doing southside quest.

## 🍰 Pullrequest

Southside Quest should also be doable now.
### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


AnonXS:
Test with this branch changes: https://github.com/cmangos/issues/issues/3170#issuecomment-1374989654

PR #592, opened 2023-01-29
Replace "DoOrSimulateScriptTextForThisInstance" with "DoBroadcastText…

fangshun2004:
…". Used to support multiple languages

## 🍰 Pullrequest
Replace “DoOrSimulateScriptTextForThisInstance” with "DoBroadcastText" to support npc shouting multilingual text display

### Proof

- None

### Issues

- None

### How2Test

- [x] None


insunaa:
Good work!
Tho ideally you wouldn't override the Broadcast Text in the function call and instead you would adjust the broadcast text in the DB instead.
Lot's of BCTs in DB don't have the sound entries or correct chat type, but that can be changed. If all you want is to get things working, your way should work, but if you want to do it right you should probably create a separate PR in the DB repository where you update the broadcast text parameters in SQL

killerwife:
This has an issue however. The old one was used for cases when the target of the text isnt loaded into the instance. So the npc either has to be active or we need a mechanism for always loading grids for certain npcs :) Make sure you test that. For example KT in Naxx is def a problem cos he is too far from the instance.

killerwife:
DoBroadcastText and DoScriptText are identical in their implementation except of defaults. So they have the same limitation. Dont get me wrong, DoOrSimulate should be removed in all cases, but noone took time to remove this limitation :)

AnonXS:
I'd say as starting point we should 1:1 port all information from script_texts to the respective BCT, but beware that some texts, especially short ones, can come in as multiple BCT text ids, so you've to choose the right one.

AnonXS:
yes that is perfectly fine to do. emote -> [EmoteID1](https://github.com/cmangos/issues/wiki/broadcast_text#EmoteID1)
``` 
EmotesID	Unused ```

PR #596, opened 2023-02-25
Problems accessing the server.

Ostring:
The last commit seems to have caused a problem when trying to access the server, to be more exact when entering the worldserver it says that the server is full.

Only accounts whit gm can enter.

-- sorry for my bad eng....


Niam5:
This should not be a pull request, this should be in the issues repository or asked on discord. Eluna will not be part of the main project as it does not align with the goal.

PR #598, opened 2023-03-02
fix honorable kills update field name

celguar:


killerwife:
You marked it as a draft so i had to make my own commit.

celguar:
Thank you

PR #600, opened 2023-03-11
[MMAPs] Fix Path Finding for .go c 5540016

AnonXS:
befor: npcs being stuck ontop unable to path down. 

after: no such issue.

We might want to think about tweaking the default param for walkable height and climb a bit, just so much that npcs cant go over fences.


AnonXS:
Map 540:

Befor:

grafik

After:

grafik

PR #605, opened 2023-04-29
[SD] AQ40: Update Flesh Tentacle 15802 Summon Locations

AnonXS:


killerwife:
We will have to move these one day to db but for now its fine.

PR #609, opened 2023-05-19
[Spell] Moss Hide s.8852 should not be removed on Evade

AnonXS:
everyone has it, never removed, also not on low hp while fleeing. (except Mosshide Brute 1012)


AnonXS:
Legit change, all except  "mosshide 1012" have this and it is a aura passive.

PR #613, opened 2023-06-04
Add disabled check in UpdateSpellLists

Foritus:
## 🍰 Pullrequest
The creature "Infernal Attacker" in Shadowmoon valley seems to cause segfaults in the server when starting up. Its spell list is disabled but contains a single spell with spellId=2, which returns nullptr when we try to run `sSpellTemplate.LookupEntry<SpellEntry>(spell.SpellId);`.
I'm not sure if this is from the spell slot -> spell list migration (looking in the sql migration templates), but at least with this fix my server doesn't bomb out on startup anymore. Unsure if this will break some other behaviour I'm not familiar with, still quite new to the codebase 🙂

### Proof
Crash site:
image
Calling method, UpdateSpellLists, note that spellInfo is nullptr
image

### Issues

I can't find any existing issues for this

### How2Test

I think the repro for this is to park a character in Shadowmoon valley so that the area gets populated, then restart the server. I'll try it on a fresh instance though.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/BaseAI/UnitAI.cpp
+++ b/src/game/AI/BaseAI/UnitAI.cpp
@@ -1192,6 +1192,11 @@ void UnitAI::UpdateSpellLists()
 
     CreatureSpellList const& spells = GetSpellList();
 
+    if (spells.Disabled)
+    {
+        return;
+    }
+
     // when probability is 0 for all spells, they will use priority based on positions
     std::vector<std::tuple<uint32, uint32, uint32, Unit*>> eligibleSpells;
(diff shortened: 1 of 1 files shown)


AnonXS:
yea seems `SELECT * FROM creature_template_spells where entry IN (21419);` was not removed. as it should not be a controllable spell.
spellid does not exist.
grafik

killerwife:
Actually its meant to be there. This is melee attack and when you control a mob its not always there. I guess im just missing a safeguard somewhere.

PR #614, opened 2023-06-21
[CreatureEventAI] add inCombat Check to EVENT_T_AURA and EVENT_T_MISS…

miraco:
…ING_AURA

## 🍰 Pullrequest

This adds a simple inCombat check to EVENT_T_AURA and EVENT_T_MISSING_AURA

### Proof

https://youtu.be/pZOSVT0y_cY 
You can see from wotlk classic, that Baroness Dorothea Millstipe casts her Shadowform Aura only inCombat, thats where this would be needed. 
https://youtu.be/9S7-fpVjypI
Here she casted it ~1 seconds into the fight.

Sniffs shown that she uses this spell between 1500ms and 8000 ms into the fight, if she doesnt have the aura already. 

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -506,9 +506,36 @@ bool CreatureEventAI::CheckEvent(CreatureEventAIHolder& holder, Unit* actionInvo
             break;
         case EVENT_T_AURA:
         {
-            SpellAuraHolder* auraHolder = m_creature->GetSpellAuraHolder(event.buffed.spellId);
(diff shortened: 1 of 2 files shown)


killerwife:
Stupid question, but if its just in combat, then why arent you doing this with normal spell cast? There is already a case for this. I assume she casts it on herself.

killerwife:
CAST_AURA_NOT_PRESENT

miraco:
Thats a good question, and tbh i dont rly know why i didnt used it like this.

PR #636, opened 2023-11-04
[SD] Add Pummel to npc_dashel_stonefist c.4961

AnonXS:


killerwife:
How about you add spell lists and i rewrite the AI so that it supports them instead of doing this.

AnonXS:
sure. :D

PR #640, opened 2023-11-29
SpawnManager: Fix self-invalidating iterator in Update loop

insunaa:
## 🍰 Pullrequest

This PR fixes an operation in the update loop that invalidates its parent iterator

### DIFF
--- a/src/game/Maps/SpawnManager.cpp
+++ b/src/game/Maps/SpawnManager.cpp
@@ -78,15 +78,13 @@ void SpawnManager::Initialize()
 void SpawnManager::AddCreature(uint32 dbguid)
 {
     time_t respawnTime = m_map.GetPersistentState()->GetCreatureRespawnTime(dbguid);
-    m_spawns.emplace_back(TimePoint(std::chrono::seconds(respawnTime)), dbguid, HIGHGUID_UNIT);
-    std::sort(m_spawns.begin(), m_spawns.end());
+    m_spawnInserts.emplace_back(TimePoint(std::chrono::seconds(respawnTime)), dbguid, HIGHGUID_UNIT);
 }
 
 void SpawnManager::AddGameObject(uint32 dbguid)
 {
     time_t respawnTime = m_map.GetPersistentState()->GetGORespawnTime(dbguid);
-    m_spawns.emplace_back(TimePoint(std::chrono::seconds(respawnTime)), dbguid, HIGHGUID_GAMEOBJECT);
-    std::sort(m_spawns.begin(), m_spawns.end());
+    m_spawnInserts.emplace_back(TimePoint(std::chrono::seconds(respawnTime)), dbguid, HIGHGUID_GAMEOBJECT);
 }
 
 void SpawnManager::RespawnCreature(uint32 dbguid, uint32 respawnDelay)
@@ -160,6 +158,9 @@ void SpawnManager::RespawnAll()
 
 void SpawnManager::Update()
 {
+    std::move(m_spawnInserts.begin(), m_spawnInserts.end(), std::back_inserter(m_spawns));
+    m_spawnInserts.clear();
+    std::sort(m_spawns.begin(), m_spawns.end());
     auto now = m_map.GetCurrentClockTime();
     for (auto itr = m_spawns.begin(); itr != m_spawns.end();)
     {
--- a/src/game/Maps/SpawnManager.h
+++ b/src/game/Maps/SpawnManager.h
@@ -74,6 +74,7 @@ class SpawnManager
         Map& m_map;
 
         std::vector<SpawnInfo> m_spawns; // must only be erased from in Update
+        std::vector<SpawnInfo> m_spawnInserts;
         std::map<uint32, SpawnGroup*> m_spawnGroups;
 };
 


src/game/Maps/SpawnManager.cpp
```diff
@@ -160,6 +158,9 @@ void SpawnManager::RespawnAll()
 
 void SpawnManager::Update()
 {
+    std::move(m_spawnInserts.begin(), m_spawnInserts.end(), std::back_inserter(m_spawns));
+    m_spawnInserts.clear();
+    std::sort(m_spawns.begin(), m_spawns.end());
```
killerwife, line 163: Yeah this solution wont float. This is going to be extremely expensive once we port everything to spawn groups.

killerwife: dynguid*

insunaa: it's apparently been causing other issues too, but I can't see a solution that doesn't do this *and* preserves `AddCreature()` in LoadFromDB

killerwife:
The issue is caused by my recent commit:

https://github.com/cmangos/mangos-tbc/commit/fd90b989fe6cc94a8dd508244c7f87cda18ce131#diff-b92ffcf851f438f2744429e37cfea1e7f7ad6a7c089b1bab347aece4c5a7a4e7R1648

I tried to be cute and treated the symptom instead of doing the really hard fix of detecting it on map init due to how persistent state works. Basically need to stop adding them permanently to grid and on map init load them properly from respawn information.

killerwife:
> That's an interesting discovery indeed! But I think this patch isn't fully fixing the issue given API like `RespawnCreature`, `RespawnGameObject` and `RespawnAll` (1) also iterate the same `m_spawns` vector while they can be called from a different thread running `Update` if I see it correctly. And even if that's not the case today, it's only one patch away to reintroduce a subtle and hard to track issue.
> 
> If we need to do calls on the same object from multiple threads - which we clearly do here - then making things properly synchronized is the only good fix IMHO. Otherwise it's just going to be an endless stream of hard to debug misery.
> 
> (1) `RespawnAll` is apparently even buggy because it's not iterating at all...

Not cross threaded. Invalidated within a single call stack.

evil-at-wow:
> Not cross threaded. Invalidated within a single call stack.

In the case that triggered this PR. But I see for example that ```ChatHandler::HandleRespawnCommand``` can also call ```RespawnCreature```, and there are others, and I'm not sure we're guaranteed all such calls will happen in the same thread that calls ```Update```. I'm not sure - it's extremely hard to track what can be done from which thread in this core - but I have my doubts all API is only used from the same thread. In which case my points still stands: we have other similar issues that are only going to happen once in a blue moon and will be very hard to figure out.

Resizing a vector can also invalidate iterators, so again will not be enough IF other API like ```RespawnCreature``` can be used from another thread (note the IF).

insunaa:
yeah, but it would invalidate the iterators before new iterators are assigned and used, which is why I put it forth as an option :D
either way at some point an issue occurs where AddCreature() is called until all guids are used up, at least I just experienced that but the program closed before I could get a callstack

killerwife:
> > Not cross threaded. Invalidated within a single call stack.
> 
> In the case that triggered this PR. But I see for example that `ChatHandler::HandleRespawnCommand` can also call `RespawnCreature`, and there are others, and I'm not sure we're guaranteed all such calls will happen in the same thread that calls `Update`. I'm not sure - it's extremely hard to track what can be done from which thread in this core - but I have my doubts all API is only used from the same thread. In which case my points still stands: we have other similar issues that are only going to happen once in a blue moon and will be very hard to figure out.
> 
> Resizing a vector can also invalidate iterators, so again will not be enough IF other API like `RespawnCreature` can be used from another thread (note the IF).

Chat commands are run in world thread when no other map thread is running. So still, single threaded stack.

killerwife:
Linked a commit, check if that does the trick for you.

insunaa:
doesn't compile for me, some constructor seems to be missing

killerwife:
fkin time_point is not noexcept, working on it

insunaa:
m_spawns.insert(m_spawns.end(), m_deferredSpawns.begin(), m_deferredSpawns.end());
not an option?

killerwife:
Fixed. While that may seem like the simple fix, I wanted to avoid copying. Hence why this compilation issue occured.

insunaa:
something like this
```cpp
    if (!m_deferredSpawns.empty()) // cannot insert during update
    {
        m_spawns.insert(m_spawns.end(), std::make_move_iterator(m_deferredSpawns.begin()), std::make_move_iterator(m_deferredSpawns.end()));
        m_deferredSpawns.clear();
    }
```
should also avoid copying

killerwife:
That will do more allocations.

evil-at-wow:
> Chat commands are run in world thread when no other map thread is running. So still, single threaded stack.

Except that's not how it works. Even if no map thread is running at the time, the simple fact the calls are done from different threads means access need to be properly synchronized, otherwise there is zero guarantee that one thread will see the effects caused by the other, which is known as data races. For example, one thread could literally be run an a different CPU (like in a different socket on the motherboard) and whatever data it updated might only have reached its own cache and not another CPU that (later) started to run the other function. That's why you need proper synchronization: to make sure that cache coherency does its job and CPUs/cores see the values you intended.

Sure, practically speaking you can get away with a lot of them, but you will be bitten sooner or later. But I suppose I'm already talking about another class of issues lurking than the one that's being solved here.

insunaa:
from my understanding this would resize the m_spawns vector to fit both vectors (moving it in the process) and then basically just concat the 2 vectors
the backinserter does the same thing-ish, but it does it element by element, instead of insert which can *in theory* move larger chunks of memory at once

insunaa:
also there's nevertheless still the issue of exponential growth:
image

killerwife:
Yeah, thats the result of that other issue I linked with my commit above.

PR #643, opened 2023-12-08
Fix bots not drinking or eating because of dirty movement flags after facing something

Scarabol:
## 🍰 Pullrequest

I just debugged into the non-drinking/eating bots. The problem is that the spell is aborted, because the bot thinks it's moving all the time. You can see that by targeting your bot and calling ".debug moveflag" with gm permissions.
image
The flags are there, because MoveSplineInit::Launch() sets them, when it is called by [Unit::SetFacingTo](https://github.com/cmangos/mangos-tbc/blob/1789a95a27de389ce5ed771046db98c7d68592da/src/game/Entities/Unit.cpp#L4647) in the FollowMovementGenerator::_setOrientation call.

### Proof

- None

### Issues

- fixes #2830
- fixes #3309

### How2Test

- Log into game, add bot, make bot go out of mana or low on health
- Error: Bot spams chat, but does not drink or eat
- Apply fix provided here
- Solution: Bot drinks and eats as expected

### Todo / Checklist

- [x] None

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -4647,7 +4647,7 @@ void Unit::SetFacingTo(float ori)
     init.Launch();
     // orientation change is in-place
     UpdateSplinePosition();
-    movespline->_Finalize();
+    DisableSpline();
 }
 
 void Unit::SetFacingToObject(WorldObject* object)
@@ -4661,7 +4661,7 @@ void Unit::SetFacingToObject(WorldObject* object)
     init.Launch();
     // orientation change is in-place
     UpdateSplinePosition();
-    movespline->_Finalize();
+    DisableSpline();
 }
 
 bool Unit::isInAccessablePlaceFor(Unit const* unit) const
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -192,8 +192,6 @@ void CreatureEventAI::InitAI()
         processMap(creatureEvent);
     }
 
-    if (m_creature->GetDbGuid() == 5026)
-        printf("");
     auto creatureEventsGuidItr = m_creature->GetMap()->GetMapDataContainer().GetCreatureEventGuidAIMap()->find(m_creature->GetDbGuid());
     if (creatureEventsGuidItr != m_creature->GetMap()->GetMapDataContainer().GetCreatureEventGuidAIMap()->end())
     {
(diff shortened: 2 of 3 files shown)


obi23lipnik:
Tested out on my server, and it works 💯

f0n1x:
I ported this fix to classic, and it does indeed resolve the drinking/eating issue.

obi23lipnik:
@killerwife want to call attention to this PR; have been running with this change on my server with heavy bot use without issues.. It's getting a little annoying to have to cherry-pick the two commits every rebase 😓 

TL;DR: 
- [ ] Approved
- [ ] Merged

killerwife:
Sorry, christmas rushes at work have kept me away from wow. I will write a comment on the problematic line of code.

src/game/Entities/Unit.cpp
```diff
@@ -4647,7 +4647,7 @@ void Unit::SetFacingTo(float ori)
     init.Launch();
     // orientation change is in-place
     UpdateSplinePosition();
-    movespline->_Finalize();
+    DisableSpline();
```
killerwife, line 4650: Disable calls _Interrupt, I need to call _Finalize. The way this works is that on blizz end, orientation SMSG_MONSTER_MOVE is immediate and is the reason why its this way. I understand that prolly the "MOVEFLAG_SPLINE_ENABLED | MOVEFLAG_FORWARD" is required, so a new function is in order called EndSpline. Please test if a new func with _Finalize and the moveflag change does the trick.

obi23lipnik: @killerwife  so, like: https://github.com/cmangos/mangos-tbc/commit/cc5eacc000d3ab9bc0d9235ab4a5cb804f683c9a ?

I haven't yet been able to test this on my live server, but it's the gist, right?

obi23lipnik: aye, does do the trick ☝️

killerwife: Amazing.

killerwife: Add the commit to the PR then.

obi23lipnik: I have no edit permissions on this PR 😞 

Opened a new one: https://github.com/cmangos/mangos-tbc/pull/647

killerwife:
Merged in the amended one

joltcan:
The fix works great! Nice find!

PR #655, opened 2024-01-23
Playerbot: shaman updates and fixes

mikehackworth:
## 🍰 Pullrequest
Update playerbot logic for shaman to: 

- Prioritize Flame Shock uptime over using Stormstrike.
- Cast Water shield if available.
- If bot has combat orders to assist, use more appropriate fire totems instead of flametongue (effect does not stack with Windfury). Searing totem for single target and magma totem for multi target.
- Fix Weapon imbues (Flametongue and Windfury) so that they are actually cast as expected. Also add logic to cast Windfury weapon on off-hand weapon when available.
- Save Shamanistic rage for when mana is < 20%

### How2Test
Add a bot which is a lv70 shaman, deep enchancement spec, set combat orders to assist. Use DPS meter etc to compare before and after.

### Todo / Checklist

- [X] None


insunaa:
Awful timing... Bots are being deprecated and replaced in the near future.

mikehackworth:
> Awful timing... Bots are being deprecated and replaced in the near future.

@insunaa I see, that is unfortunate. Can you share any details about the intended replacement?

insunaa:
https://github.com/cmangos/mangos-classic/pull/528

evil-at-wow:
> Bots are being deprecated and replaced in the near future.

That was news for me as well. This may not be the best place to ask, but anything you can share or point to about advantages/improvements, or is this a "it's totally different but so much better" situation?

cpevors:
> > Bots are being deprecated and replaced in the near future.
> 
> That was news for me as well. This may not be the best place to ask, but anything you can share or point to about advantages/improvements, or is this a "it's totally different but so much better" situation?

They are mangosbot, based on ike3 bots. You can find relevant discussion on the CMaNGOS Discord, here: https://discord.com/channels/362206349339262976/1195445025862074408

insunaa:
> > Bots are being deprecated and replaced in the near future.
> 
> That was news for me as well. This may not be the best place to ask, but anything you can share or point to about advantages/improvements, or is this a "it's totally different but so much better" situation?

Sure!
Unlike our current playerbots which have unfortunately not been maintained for quite some time the mangosbots are being actively maintained and developed. They are also not just party-bots but also random world population bots, which roam the world doing various things looking as if they were actually doing something. You can invite them to party up for quest etc.
They also support full 40-man raids unlike our playerbots which are limited to 9 due to the account character limit. They also do PvP in battlegrounds and use the LFG system

PR #659, opened 2024-02-06
Clang: Fix compiler warnings

insunaa:
## 🍰 Pullrequest

See https://github.com/cmangos/mangos-wotlk/pull/493

(Please go over this PR once since it came from a dirty tree and I'm not 100% sure if I got everything incorrect out. It compiles ofc. but I don't want to accidentally sneak something in)


evil-at-wow (approved):
This looks consistent with the previous one you did in WotLK, and I didn't see anything unexpected with a quick glance.

The GridMap change is new, but that one makes sense and is fine here. The difference is because apparently in the WotLK core ```m_holes``` still is a raw pointer with some 'manual' (de)allocation, so that check probably makes sense there, while ```m_holes``` is an array in the TBC core.

insunaa:
THank you very much for going over it :heart:

PR #665, opened 2024-03-07
Implementation of Autobroadcast & XP Rate Command

Diabloxx:
## 🍰 Pullrequest
Added .xp command to the core with options to choose xp rates, rates are set in the database with the included sql in custom/xpBoost.sql

### Proof
[Images](https://imgur.com/a/ek5FsvT)

### Issues
None, works like a charm

### How2Test
Autobroadcast:
Import custom/mangos/autobroadcast.sql to your world database
Add any text you want in autobroadcast table in database.
Enable with a new line in config.

**AutoBroadcast = 60** Where 60 equals seconds where the autobroadcast should post a message in chat.

XP Part:
Import custom/mangos/xpcommand.sql to your world database
Import custom/character/xpBoost.sql to your character table. 
Find experience_brakcet_cap, add 2 new lines with the desired information
Min/Max - Equals what level this rate will be available to
Team = 0 Alliance - 1 Horde
Value = Maximum XP Rate

### DIFF
--- a/sql/custom/characters/xpBoost.sql
+++ b/sql/custom/characters/xpBoost.sql
@@ -1,0 +1,16 @@
+DROP TABLE IF EXISTS `experience_bracket_cap`;
+CREATE TABLE `experience_bracket_cap` (
+  `low` int(11) unsigned NOT NULL DEFAULT '0' COMMENT 'Lower bracket bound',
+  `high` int(11) unsigned NOT NULL DEFAULT '0' COMMENT 'Upper bracket bound',
+  `team` int(11) unsigned NOT NULL DEFAULT '0' COMMENT 'Bracket team',
+  `value` int(11) unsigned NOT NULL DEFAULT '0' COMMENT 'Bracket XP cap',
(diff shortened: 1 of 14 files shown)


insunaa:
but why

Diabloxx:
Neat feature to have, in my opinion

al3xc1985:
yes, very good feature

killerwife:
Okay, so I am closing this and here is why:

The PR contains my code, without giving authorship. Thats fine. - https://github.com/killerwife/mangos-tbc/commit/c17a424e677e9aed4463ad905bb311ac22345723 (actual code is as old as balls, maybe 2018|

But there is a reason why this code is not merged here and it is available on my public github.

I am fine with someone maintaining a fork with custom features of cmangos but this does not line up with the project goals. The reason is that if its merged, we have to support it. Its on my github for anyone to take, for that same reason, I do not have to support it.

We arent really making server software here (although it can be used that way), and anyone who wants can push it to their own forks as they desire.

Will happily merge blizzlike code in the future.

PR #676, opened 2024-05-19
Fix Lady Vashj not correctly resetting Timers on Wipe resulting in sp…

miraco:
…awning adds while being ooc.

## 🍰 Pullrequest

Lady Vashj and prob. other Bosses dont correctly reset timers on wipe resulting in constantly spawning adds, even while being ooc or being in phase 1.

Prob. came with this change: https://github.com/cmangos/mangos-tbc/commit/e512fc1a6e3e764ddf88e429c2a6714e0365aea3

### Proof

- None

### Issues

- None
https://github.com/cmangos/issues/issues/3589

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/serpent_shrine/boss_lady_vashj.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/serpent_shrine/boss_lady_vashj.cpp
@@ -435,6 +435,8 @@ struct boss_lady_vashjAI : public CombatAI
         if (m_instance)
             m_instance->SetData(TYPE_LADYVASHJ_EVENT, FAIL);
 
+        ResetAllTimers();
+
         DespawnGuids(m_spawns);
     }
 


killerwife:
While this addresses the issue itself, this should never be needed, due to vashj despawning on wipe. I will need to debug this.

insunaa:
@killerwife 
as far as I can tell the root cause is
```patch
diff --git a/src/game/AI/ScriptDevAI/base/TimerAI.cpp b/src/game/AI/ScriptDevAI/base/TimerAI.cpp
index 2913ab63b..67fbcc330 100644
--- a/src/game/AI/ScriptDevAI/base/TimerAI.cpp
+++ b/src/game/AI/ScriptDevAI/base/TimerAI.cpp
@@ -152,8 +152,7 @@ void TimerManager::ResetAllTimers()
 void TimerManager::ResetTimersOnEvade()
 {
     for (auto& data : m_timers)
-        if (data.second.combatSetting != TIMER_ALWAYS)
-            data.second.ResetTimer();
+        data.second.ResetTimer();
 }
 
 void TimerManager::GetAIInformation(ChatHandler& reader)
```
that we check whether or not the timer is TIMER_ALWAYS and don't reset/disable the timer again if it is.

imo we should always reset Timers. I can't think of any timer in any scenario that needs to keep running from where it was at while/after a boss has evaded

PR #678, opened 2024-06-18
Hackfix: add target check for Dance Vibe spell in Karazhan to silence the combatlog

miraco:
## 🍰 Pullrequest

This is prob. the hardest hackfix i ever did, but i still wanted to share it here, for people who are Interested.

It should drasticly reduce the combatlog spam in karazhan.

### Proof

- None

### Issues

- None
This is related to this discussion:
https://github.com/cmangos/issues/issues/3323

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/karazhanScripts.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/karazhanScripts.cpp
@@ -590,6 +590,36 @@ struct WhipFrenzyHorses : public SpellScript
     }
 };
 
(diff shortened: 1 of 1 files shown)


insunaa:
I think you can simplify that quite a bit. In this context `target` can't be nullptr, and both `target` and `caster` *have to be* creatures, if it's EFFECT_APPLY_AREA_AURA_PARTY, so that doesn't have to be checked either. Then I don't think you need to do the InCombat checks either, because afaik area auras don't just disappear when enemies enter combat, or that'd make them pretty pointless, and if the aura is cast on spawn then they shouldn't ever cast it fresh while in combat regardless

PR #681, opened 2024-06-23
Completly remove boss the maker script, and add it to acid and spell_…

miraco:
…list

## 🍰 Pullrequest

This will completly remove the script for this Boss. Its a simple tank and spank encounter and doesnt need a full script. Acid and spell_list system can handle all of this.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


killerwife:
Fyi, the only thing wrong with the commit was the missing BF: prefix in the message, otherwise I have no problem with this.

PR #691, opened 2024-08-15
EAI: add ACTION_T_INFORM_PET

miraco:
## 🍰 Pullrequest

With this we have a better handling for Summoning NPCs that dont despawn when their owner dies.
https://github.com/cmangos/issues/issues/3755#issuecomment-2287088657

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AI/EventAI/CreatureEventAI.cpp
+++ b/src/game/AI/EventAI/CreatureEventAI.cpp
@@ -1189,6 +1189,18 @@ bool CreatureEventAI::ProcessAction(CreatureEventAI_Action const& action, uint32
             SendAIEventAround(AIEventType(action.throwEvent.eventType), target, 0, action.throwEvent.radius);
             break;
         }
+        case ACTION_T_INFORM_PET:
+        {
+            Pet* pet = m_creature->FindGuardianWithEntry(action.informPetEvent.creatureId);
+            if (!pet || !pet->IsAlive())
+            {
+                if (failedTargetSelection)
+                    sLog.outErrorEventAI("Event %d attempt to start relay script but Target == nullptr. Creature %d", eventId, m_creature->GetEntry());
+                return false;
+            }
+            SendAIEvent(AIEventType(action.informPetEvent.eventType), m_creature, pet);
+            break;
+        }
         case ACTION_T_SET_THROW_MASK:
         {
             m_throwAIEventMask = action.setThrowMask.eventTypeMask;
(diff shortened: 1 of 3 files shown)


killerwife:
Why not integrate it into send ai event to send to pet?

miraco:
I tried adding TARGET_T_PET first so we can use it for more things
devenv_vgHNRmEbeb
but this never worked, target was always Target == nullptr

then i tried looking at boss scripts where this could get used and i found it working with this at mgt boss fight
devenv_JZJ3AvWkMw

and ACTION_T_THROW_AI_EVENT already has all 3 params in use so i thought its better using a new action for this

Imo the best way is so somehow add a new target like, TARGET_T_PET

killerwife:
Problem is guardian != pet. Either way looking into it.

killerwife:
Ok and why does ACTION_T_DESPAWN_GUARDIANS not help here? In the issue it was about despawning them on death right?

Not saying it fixes the whole situation.

killerwife:
Okay, situation is as follows ->

ACTION_T_THROW_AI_EVENT from EAI - throws event around AI user and target is only used as action invoker, it doesnt really actually send the event to the guardian since you cant rly target one singular creature

SCRIPT_COMMAND_SEND_AI_EVENT - enables sending it to multiple targets but the system has no shorthand way of targeting guardians, the only way would be to assign these specific guardians a string id and target with those but then you could end up despawning spawns of a different master

Problem with targeting guardians is that its 1:N, and all of targeting in EAI is inherently 1:1 as it currently stands. I can add a SCRIPT_COMMAND_SEND_AI_EVENT equivalent to EAI, which will target a singular creature, however then we still have to solve the fact guardian targeting is 1:N. With the advent of string id, dbscripts have this issue inherently solved, since they can do 1:N.

So give me a rundown of what you want to do beyond despawn guardians, and I will model it upon that testcase. I get that you guys need a new possibility added here.

miraco:
> Ok and why does ACTION_T_DESPAWN_GUARDIANS not help here? In the issue it was about despawning them on death right?
> 
> Not saying it fixes the whole situation.

No they shouldn’t instantly despawn  when their master dies.
They will only despawn on Reached Home (after Evade)

killerwife:
Okay but thats default guardian behaviour. What spell are they summoned through?

miraco:
> Okay but thats default guardian behaviour. What spell are they summoned through?

Yes it should be default guard behaviour, but this doesn’t seem to be working.

https://www.wowhead.com/classic/spell=12746/

killerwife:
Now we are getting somewhere, will get back to you.

miraco:
If you are you sure that this is default pet behaviour, we could prob just add a check here
https://github.com/cmangos/mangos-tbc/blob/27f97760bfa57d3774086b771593de08c46b2428/src/game/MotionGenerators/HomeMovementGenerator.cpp#L101

Something like 
if(!owner-IsAlive())
m_creature->despawn;

Or is there a flag that handle this?

killerwife:
Check commits I made.

miraco:
Works perfectly fine, ty

PR #700, opened 2024-10-09
Battle Shout used from various tbc trash npcs should not stack

miraco:
## 🍰 Pullrequest

[Battle Shout](https://www.wowhead.com/spell=31403/battle-shout) should not be stackable.

### Issues

Using Ethereal Crypt Raider as example here:
This npcs can use Battle Shout on their self. Pulling multiple npcs of this type can result in overstacking Battle Shout.
Wow_eM1jdiM3z9

Tests on retail showed that they use it on their self even if they have the Buff (refreshing timer from 1min to 2 minutes)

### How2Test

- None

### Todo / Checklist

- [X] None


killerwife:
Can you pls try adding SPELL_ATTR_EX_AURA_UNIQUE to the spell and see if that fixes the issue.

miraco:
Yep this seems to also work, should i update the pr or are you gonna do it yourself?

killerwife:
Update PR, that way i can close this very quickly. Dont have dev open rn.

miraco:
Done

PR #701, opened 2024-10-14
Missing bot-specific pathfinder modifications to allow for travel node generation.

mostlikely4r:
## 🍰 Pullrequest
This PR (re-)implements some adjustments to the existing pathfinder logic to allow for pathfinding without a source unit. This is required for generating travelnodes used in long-range bot pathfinding.
This PR also includes a fix in using the proper instanceId when using get/set area which allows navmesh manipulation by bots in instances.

These changes should have been implemented during the original bot-code merge but where forgotten or have been broken since then.

Note without these changes bots will continue to work fine however we will be unable to re-generate the nodemap to keep up with future map-changes or fix existing minor issues (such as missing zeppelin paths in tbc).

### Proof
An example can be found in the current pathfinder:  https://github.com/cmangos/mangos-tbc/blob/1605ff06b177df6cdb4b4d72b7038ea273ca70d5/src/game/MotionGenerators/PathFinder.cpp#L127

This simply stops any calculation without a m_sourceUnit. This functionality is however essential to generate paths during server start. 

Here you can see the original bot implementation lacks that check: https://github.com/celguar/mangos-tbc/blob/63b2e8f43fb1a10e4a7b55de2f0697fbd46e8979/src/game/MotionGenerators/PathFinder.cpp#L119

If you look at the changes made they are limited to bypassing/implementing some m_sourceUnit checks and only effect builds that include playerbots. Also some extra instanceId handling specifically for get/setArea. Normal pathfinding should not be affected.

### Issues
-Unreported: Bots in tbc no longer use zeppelins to travel between continents.
-Unreported: Bots manipulating the navmesh to avoid mobs crash the server when in an instance.

### How2Test
Testing travelnode generation can be tested by clearing the ai_playerbot_travelnode table. Doing this without this fix will result in 800 travelnodes remaining mostly in instances. With this fix 2500+ nodes will remain including proper paths in the overworld.

### Todo / Checklist
-If this fix is implemented botcode should be adjusted so bots are allowed to manipulate the navmesh inside instances again.

### DIFF
--- a/src/game/MotionGenerators/PathFinder.cpp
+++ b/src/game/MotionGenerators/PathFinder.cpp
@@ -38,6 +38,9 @@ PathFinder::PathFinder(const Unit* owner, bool ignoreNormalization) :
     m_cachedPoints(m_pointPathLimit * VERTEX_SIZE), m_pathPolyRefs(m_pointPathLimit), m_polyLength(0),
     m_smoothPathPolyRefs(m_pointPathLimit), m_sourceUnit(owner), m_navMesh(nullptr), m_navMeshQuery(nullptr),
     m_defaultMapId(m_sourceUnit->GetMapId()), m_ignoreNormalization(ignoreNormalization)
+#ifdef ENABLE_PLAYERBOTS
+    , m_defaultInstanceId(m_sourceUnit->GetInstanceId())
+#endif
 {
     DEBUG_FILTER_LOG(LOG_FILTER_PATHFINDING, "++ PathFinder::PathInfo for %u \n", m_sourceUnit->GetGUIDLow());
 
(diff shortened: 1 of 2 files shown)


src/game/MotionGenerators/PathFinder.cpp
```diff
@@ -175,7 +186,7 @@ void PathFinder::setArea(uint32 mapId, float x, float y, float z, uint32 area, f
 
     MMAP::MMapManager* mmap = MMAP::MMapFactory::createOrGetMMapManager();
 
-    dtNavMeshQuery const* query = mmap->GetNavMeshQuery(mapId, 0);
```
killerwife, line 178: I have not watched this code when it was initially merged since my codepath doesnt use setArea. If I am correct however, this is a massive race condition.

killerwife: Even in the new version

mostlikely4r: > Even in the new version

Could you please elaborate? My assumption was that these lines are not much different from: https://github.com/cmangos/mangos-tbc/blob/1605ff06b177df6cdb4b4d72b7038ea273ca70d5/src/game/MotionGenerators/PathFinder.cpp#L46 which is called for bot-movement in similar fashion.

Do you have some advice how you'd prefer this functionality be implemented?

killerwife: That one is fine. But imagine this, a map, has multiple instances. Lets say BG. You have 3 wsgs happening side by side, with different instance ids, BUT the bots inside the bgs all access default instance 0. Its safe to access a mesh from different threads, its not safe to write to a mesh from 3 different threads (we can assume each map instance is own thread)

killerwife: Ok, if its only used during generation stage thats fine. Where is it first generated however? I assume whole map is generated at once right?

mostlikely4r: > Ok, if its only used during generation stage thats fine. Where/when is it first generated however? I assume whole map is generated at once right?

Correct. The initial generation is done here:
https://github.com/cmangos/playerbots/blob/8aa5c3a903a41c6eec91bd0720c234ef82bf6a45/playerbot/TravelMgr.cpp#L942
This creates 1 thread for each mapId and stops the server until all calculations are done.


This is 'only' called when the server's ai_playerbot_travelnode table is empty. However this table is never empty for normal use since we distribute the content: https://github.com/cmangos/playerbots/blob/8aa5c3a903a41c6eec91bd0720c234ef82bf6a45/sql/world/tbc/ai_playerbot_travel_nodes.sql

So the generation stage is something only a developer will run to refresh/update the information of that .sql file.

killerwife: Okay then, disregard my concerns.

killerwife:
Should I merge this or are you not yet

PR #712, opened 2024-12-05
Core/Map: Fix crash when battlegrounds ends

d3athbl0w:

### DIFF
--- a/src/game/Maps/Map.cpp
+++ b/src/game/Maps/Map.cpp
@@ -2148,7 +2148,10 @@ void BattleGroundMap::Update(const uint32& diff)
 {
     Map::Update(diff);
 
-    if (!m_bg->GetPlayersSize())
+    if (!m_bg)
+        return;
+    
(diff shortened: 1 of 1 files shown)


d3athbl0w:
I have made this fix and it works, the bgs no longer crash at end, but I am not sure if this fix complies with the CMaNGOS fix development standards. It would be my first fix contributed to the CMaNGOS community. If you can correct me, I will gladly listen to you and I learn.

mostlikely4r:
See also https://github.com/cmangos/mangos-wotlk/pull/574
This was closed because initially I was able to trivially reproduce this. 
When you look at the method it's easy to see m_bg being set to nullptr can lead to this issue in the next call. The question is however why it doesn't happen more often and why the mechanism that normally prevents it fails here.

killerwife:
I put it in vanilla, I put it in wotlk, why didnt i put it in tbc?

killerwife:
Also if (m_bg->GetPlayersSize() == 0) is identical to the other statement.

PR #717, opened 2024-12-31
Fixed: Core update including routine managment not crashing

dvdvideo1234:
## 🍰 Pullrequest
When doing core updates user will bet the following error
```
>>> ERROR 1370 (42000) at line 24: alter routine command denied to
  user 'mangos'@'localhost' for routine 'tbccharacters.uint32toint32'
```
This happens because there is no routine privilege.

### Proof

[27.2.2 Stored Routines and MySQL Privileges](https://dev.mysql.com/doc/refman/8.4/en/stored-routines-privileges.html)

### Issues

- [X] None


killerwife:
We do not use routines in cmangos and until someone shows me the core is doing that, not merging additional priviledges we do not need

dvdvideo1234:
@killerwife 

It may does not, however the update SQL certainly does. It includes a routine being dropped but the user does not have access to make the change.

killerwife:
Yes but is that any SQL we wrote? I dont discourage anyone doing whatever they want on their own server. Just official cmangos ones have to reflect what we do.

dvdvideo1234:
So the update will continue crashing

miraco:
What update will crash? Can you link specific sql file?

dvdvideo1234:
> What update will crash? Can you link specific sql file?

It is located in the attached ticket.

killerwife:
That update function was a mistake. Use this instead https://github.com/cmangos/mangos-tbc/tree/master/contrib/iteminstance_converter

PR #719, opened 2025-01-01
Proc: initialize canProc with true

insunaa:
## 🍰 Pullrequest

This PR fixes the `canProc` variable being uninitialized making proc behaviour of all spells unpredictable due to memory layout issues

### Issues
- fixes https://github.com/cmangos/issues/issues/3787
- fixes https://github.com/cmangos/issues/issues/3795
- fixes https://github.com/cmangos/issues/issues/3816
- fixes https://github.com/cmangos/issues/issues/3784
- fixes https://github.com/cmangos/issues/issues/3800


insunaa:
Because `MAX_EFFECT_INDEX` is a macro, it could in theory be redefined to something else that doesn't make sense and thus break this code. I'm 99.999999% sure we'll never redefine `MAX_EFFECT_INDEX` but in case we want to protect against that we could add something like this
image

PR #721, opened 2025-01-14
SSC: Remove Boss-Requirements from Vashj (2.4.0)

insunaa:
## 🍰 Pullrequest

This PR removes the requirement to kill all bosses in SSC prior to attempting Lady Vashj, as per Patchnotes of 2.4.0

### Proof

- [Patch 2.4.0](https://warcraft.wiki.gg/wiki/Patch_2.4.0#Dungeons_and_raids)

### How2Test

- Go to Lady Vashj's bridge console. Click on it immediately

### Todo / Checklist

- [ ] Maybe gate the "removal" behind PRENERF macro, to ensure both types of behaviour can be observed if desired.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/serpent_shrine/serpent_shrine.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/serpent_shrine/serpent_shrine.cpp
@@ -118,6 +118,9 @@ void instance_serpentshrine_cavern::OnObjectCreate(GameObject* pGo)
             m_goEntryGuidStore[pGo->GetEntry()] = pGo->GetObjectGuid();
             break;
         case GO_CONSOLE_VASHJ:
+#ifdef PRENERF_2_3
+            pGo->SetFlag(GAMEOBJECT_FLAGS, GO_FLAG_NO_INTERACT);
+#endif
             m_goEntryGuidStore[pGo->GetEntry()] = pGo->GetObjectGuid();
             EngageBridgeConsole(pGo);
             break;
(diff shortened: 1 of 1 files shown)


killerwife:
PRENERF_2_3

insunaa:
Yeah, I know. I'll probably rework it as I said in ToDo

cpevors:
Can confirm -- I have tested both the 2.4 and pre-2.4 behaviour. This commit seems to accomplish both, and the raid instance is working as intended.

insunaa:
Actually having just tested the bug report I couldn't actually reproduce it even without these fixes applied. Everything just worked. Some more investigation is needed, I'll return this to a draft PR.

cpevors:
This commit did restore the correct behaviour. https://github.com/cmangos/tbc-db/commit/8d940e43f3764683426d06ae3e9ccb725d101d1f

insunaa:
The PR is still valid as it now also allows pre-2.4.0 behaviour if enabled. The appropriate flags still need to be investigated, as the flags applied by the DB change did include `GO_FLAG_LOCKED` for all objects involved, which we don't currently use in DB.

src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/serpent_shrine/serpent_shrine.cpp
```diff
@@ -375,10 +383,15 @@ void instance_serpentshrine_cavern::EngageBossConsole(uint32 uiType, GameObject*
         }
         console = instance->GetGameObject(m_goEntryGuidStore[goEntry]);
     }
+    #ifdef PRENERF_2_3
```
killerwife, line 386: ifdefs are generally formatted at the complete left, not inline ill fix it myself tho

PR #722, opened 2025-01-17
[SD] boss_thermapluggAI: Adjust KnockAwayTimer & Add missing spells

AnonXS:


killerwife:
Do i need to do smth here, sniff, or find it or smth?

PR #731, opened 2025-02-27
Fix missing template instantiation in RelWithDebInfo builds

miranda:
## 🍰 Pullrequest
Fixes a linking issue on Linux for `VisitAllObjects` in `RelWithDebInfo` (`-O2 -g`) and ASan (`-fsanitize=address`) builds by ensuring `CellImpl.h` is included in `TargetedMovementGenerator.cpp`.  

### Proof
- **Build tested successfully on Linux (`RelWithDebInfo`, `Debug`, `Release`).**
- **Confirmed that missing template instantiation was the cause of the issue.**  

### Issues
- No existing issue reports for this specific problem.

### How2Test
1. Compile with `-DCMAKE_BUILD_TYPE=RelWithDebInfo -fsanitize=address` on Linux.
2. Before this fix: Linker fails with **undefined reference to `VisitAllObjects`**.
3. After this fix: Compiles and links successfully.

### Todo / Checklist
- [X] Ensure `VisitAllObjects` is correctly instantiated.
- [X] Fix Linux linker error.
- [X] Tested across different build types.

### DIFF
--- a/src/game/MotionGenerators/TargetedMovementGenerator.cpp
+++ b/src/game/MotionGenerators/TargetedMovementGenerator.cpp
@@ -28,6 +28,7 @@
 #include "Grids/GridNotifiersImpl.h"
 #include "Entities/Transports.h"
 #include "Maps/SpawnGroup.h"
+#include "Grids/CellImpl.h"
 
 #define IGNORE_M2 true // simple define for avoiding bugs due to different setting across movechase
 


insunaa:
What is the problem exactly?
Is this just a PCH=Off build issue?

miranda:
On Linux?  I tried many variations and they fail linking.  For example:

    cmake ../mangos-tbc \
        -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_TBB_PAR_BACKEND=1 -ltbb -fsanitize=address -fno-omit-frame-pointer" \
        -DCMAKE_C_FLAGS="-fsanitize=address -fno-omit-frame-pointer" \
        -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address" \
        -DCMAKE_INSTALL_PREFIX=~/cmangos/run \
        -DPCH=1 \
        -DDEBUG=0 \
        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
        -DBUILD_PLAYERBOTS=ON \
        -DBUILD_AHBOT=ON \
        -DFETCHCONTENT_FULLY_DISCONNECTED=ON

the result is:

[100%] Linking CXX executable mangosd
cd /home/serqetry/projects/cmangos-dev/public/build/src/mangosd && /usr/bin/cmake -E cmake_link_script CMakeFiles/mangosd.dir/link.txt --verbose=1
/usr/bin/c++ -D_GLIBCXX_USE_TBB_PAR_BACKEND=1 -ltbb -fsanitize=address -fno-omit-frame-pointer -O1 -Wno-unused-result -Wno-psabi -g3 -O2 -g -DNDEBUG -fsanitize=address -pthread -rdynamic -Wl,--dependency-file=CMakeFiles/mangosd.dir/link.d CMakeFiles/mangosd.dir/CliRunnable.cpp.o CMakeFiles/mangosd.dir/MaNGOSsoap.cpp.o CMakeFiles/mangosd.dir/Main.cpp.o CMakeFiles/mangosd.dir/Master.cpp.o CMakeFiles/mangosd.dir/RASocket.cpp.o CMakeFiles/mangosd.dir/WorldRunnable.cpp.o CMakeFiles/mangosd.dir/soapC.cpp.o CMakeFiles/mangosd.dir/soapServer.cpp.o -o mangosd  -Wl,-rpath,:::::: ../shared/libshared.a ../game/libgame.a ../../dep/src/gsoap/libgsoap.a ../modules/PlayerBots/libplayerbots.a ../shared/libshared.a ../framework/libframework.a ../shared/libshared.a ../framework/libframework.a -ldl /usr/lib/libboost_system.a /usr/lib/libboost_program_options.a /usr/lib/libboost_thread.a /usr/lib/libboost_regex.a /usr/lib/libboost_serialization.a /usr/lib/libboost_filesystem.a /usr/lib/libboost_chrono.a /usr/lib/libboost_atomic.a /usr/lib/libssl.so /usr/lib/libcrypto.so /usr/lib/libmariadb.so ../../dep/g3dlite/libg3dlite.a ../../dep/recastnavigation/Detour/libDetour.a /usr/lib/libz.so /usr/lib/libboost_json.a /usr/lib/libboost_container.a
/usr/bin/ld: ../game/libgame.a(TargetedMovementGenerator.cpp.o): in function ChaseMovementGenerator::FanOut(Unit&)':
/home/serqetry/projects/cmangos-dev/public/mangos-tbc/src/game/MotionGenerators/TargetedMovementGenerator.cpp:442:(.text+0x1f031): undefined reference to void Cell::VisitAllObjects<MaNGOS::UnitSearcher<MaNGOS::AnyUnitFulfillingConditionInRangeCheck> >(WorldObject const*, MaNGOS::UnitSearcher<MaNGOS::AnyUnitFulfillingConditionInRangeCheck>&, float, bool)'
collect2: error: ld returned 1 exit status
make[2]: *** [src/mangosd/CMakeFiles/mangosd.dir/build.make:237: src/mangosd/mangosd] Error 1
make[2]: Leaving directory '/home/serqetry/projects/cmangos-dev/public/build'
make[1]: *** [CMakeFiles/Makefile2:656: src/mangosd/CMakeFiles/mangosd.dir/all] Error 2
make[1]: Leaving directory '/home/serqetry/projects/cmangos-dev/public/build'
make: *** [Makefile:136: all] Error 2

insunaa:
Why are you using tbb? Also why are you setting DEBUG to 0?

miranda:
I could not get cmangos to build at all in my Garuda Linux environment w/ i7-13700K CPU.  Is there a reason I shouldn't use tbb?

I am using DDEBUG=0 because it runs much faster but I still get address sanitize and stack trace with RelWithDebInfo.

And can you explain why #include "CellImpl.h" should not be added? If it’s required for proper linkage in some configurations, shouldn’t it be included explicitly?

insunaa:
Well we don't officially support tbb and we've had several bug reports involving CMaNGOS compiled against tbb in the discord. I can't speak to how valid they are, however.
As for including `CellImpl.h` in genuinely can't reproduce your issue. I don't mind it getting included I guess, but I don't see why it's fixing your linking issue.

miranda:
Ok thanks, that makes sense about tbb, but like I said I need it to be able to build.  This is only the case on my 13700K machine, it builds fine without tbb on my vanilla Arch Linux Ryzen 5800X... not sure why this is the case, but it was a solution for my Intel machine.

The function Cell::VisitAllObjects<T> is declared in Cell.h but only defined in CellImpl.h. If CellImpl.h is not explicitly included, the function template isn’t instantiated correctly, leading to an undefined reference error at link time.  Seems like the issue may not appear for everyone because PCH can mask missing includes by precompiling headers from other parts of the codebase... but fixing it anyway would be nice so there aren't any configurations that won't link.

evil-at-wow:
There seem to be two different things at play here. 

First, the use of tbb. Using tbb is not entirely a user choice any more these days. GCC uses it for the parallel algorithm these days (because it was a contribution by Intel). See for example the documentation on https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017, and in particular note 3 there:
> Note 3: The Parallel Algorithms have an external dependency on Intel TBB 2018 or later. If the <execution> header is included then -ltbb must be used to link to TBB.
As far as I'm aware, we're not using tbb nor the parallel algorithms directly, so there's no need for tbb from our side, but I vaguely remember someone mentioning an issue where something included <execution> indirectly in a build causing issues when not linking with libtbb. I forgot what it was, but my guess would be a boost library like asio or something like that. I don't know Garuda Linux, but maybe the story is similar there (or perhaps -D_GLIBCXX_USE_TBB_PAR_BACKEND=1 triggers it unconditionally?).
Either way, I don't see your use of tbb as problematic; you are also passing '-ltbb' as the documentation says you should in those cases, and your issue is not that you get linker errors because of unresolved tbb symbols. 

Then there's the actual undefined reference you get. A missing include leading to a missing template instantiation makes sense. I can't immediately see why 'release vs debug' makes a difference (I'll check because I'm genuinely curious now), but I do agree that necessary and missing includes should be added.

miranda:
Yes, I'm using -D_GLIBCXX_USE_TBB_PAR_BACKEND=1 -ltbb

What you are saying about GCC sounds right.  I think it was boost-related linking problems, but it was months ago that I figured out I needed to adjust my cmake options to use tbb in order to get it to link on this machine... and haven't had a problem since.  But it was never problem on my Ryzen machine, it links everything fine with or without tbb, with the same GCC and libraries.  It's strange.

killerwife:
So whats the verdict. I maintain a linux installation of forked mangos-tbc and never had this issue.

insunaa:
@killerwife seems to have been caused by an errant `#include` in playerbots code

evil-at-wow:
I finally looked into this, and my conclusion is I'm going to disagree this was caused by a playerbots include, and I agree with @miranda's findings and PR.

First, let's see the relevant ```VisitAllObjects```, that's called exactly once in TargetedMovementGenerator.cpp (in ```ChaseMovementGenerator::FanOut```):

```
struct Cell
{
    ...
    template<class T> static void VisitAllObjects(const WorldObject* obj, T& visitor, float radius, bool dont_load = true);
    ...
}
```

We already knew from the PR and dicussion this was a template, and now it's clear this template is instantiated with the visitor type as template parameter. Its definition is also not in Cell as you can see, it's in CellImpl.h.

Now if you search through the cmangos-tbc code base, you'll notice ```VisitAllObjects``` is called in various source files, and every single one of them - except for TargetedMovementGenerator.cpp - includes CellImpl.h. If you now check who's including CellImpl.h, then you'll notice this is also only included by source files. So TargetedMovementGenerator.cpp also does not get the CellImpl.h include because of some indirect include chain - the compiler doesn't get to see the definition at all. That alone makes this PR sensible.

Now, why do (most) other people building on Linux, like myself, not get this linker error issue then? Well, because our linker finds the necessary symbol! Remember the ```VisitAllObjects``` template is instantiated per visitor type. In TargetedMovementGenerator this is ```MaNGOS::UnitSearcher<MaNGOS::AnyUnitFulfillingConditionInRangeCheck>```. If you search through the code base, it turns out there is exactly one other ```VisitAllObjects``` call using the same visitor type (in GameObject.cpp, the first one in ```GameObject::Update()```, currently at line 478)! And there, in GameObject.cpp, the compiler does see the template definition thanks to the CellImpl.h include and instantiates it, so when linking the linker finds the necessary code anyway and there is no problem. This theory is also very easy to check: comment out that one ```VisitAllObjects``` line in GameObject.cpp and rebuild, and bingo (even with a minimal debug build without anything playerbots):

```
/usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: ../game/libgame.a(TargetedMovementGenerator.cpp.o): in function `ChaseMovementGenerator::FanOut(Unit&)':
/mnt/cmangos-tbc/src/game/MotionGenerators/TargetedMovementGenerator.cpp:442:(.text+0x25f3): undefined reference to `void Cell::VisitAllObjects<MaNGOS::UnitSearcher<MaNGOS::AnyUnitFulfillingConditionInRangeCheck> >(WorldObject const*, MaNGOS::UnitSearcher<MaNGOS::AnyUnitFulfillingConditionInRangeCheck>&, float, bool)'
collect2: error: ld returned 1 exit status
```

The last remaining mystery is why a RelWithDebInfo build still triggers a linker error for @miranda. The answer can only be that in that case the symbol isn't there for the linker to find after all, despite GameObject.cpp using/needing it as well. Not having reproduced the issue myself, I can't be sure, but my educated guess is the compiler decided to inline the template code in her case. Release builds typically have much more optimizations going, and it's a pretty simple function template, so that wouldn't surprise me one bit.

In conclusion, I think there really is an issue in the code which this PR found and fixes. So unless someone comes with very good arguments against it, I'd like to merge the PR after all, say this weekend.

evil-at-wow:
And now practically merged with cmangos/mangos-tbc@6325dda255f9f9f4d7b974cc63d6a213bb5d35c3.

[In case you wonder: the original branch was force-pushed in the meantime, so I couldn't easily merge the PR with GitHub cooperating and seeing it as a merged commit, but the only drawback is GH thinking this PR still has unmerged commits.]

PR #736, opened 2025-04-25
Darshore: rework both Volcor escort quests

miraco:
## 🍰 Pullrequest

Rework Volcor escort quest Escape Through Force and Escape Through Stealth

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


src/game/AI/ScriptDevAI/scripts/kalimdor/darkshore.cpp
```diff
@@ -486,107 +461,59 @@ struct npc_volcorAI : public npc_escortAI
...
-                if (Player* pPlayer = GetPlayerForEscort())
-                    DoScriptText(SAY_ESCAPE, m_creature, pPlayer);
-                break;
-            case 18:
-                DoCastSpellIfCan(m_creature, SPELL_MOONSTALKER_FORM);
-                break;
-            case 24:
                 if (Player* pPlayer = GetPlayerForEscort())
                     pPlayer->RewardPlayerAndGroupAtEventExplored(QUEST_ESCAPE_THROUGH_STEALTH, m_creature);
+                m_creature->GetMotionMaster()->Clear(false, true);
```
killerwife, line 504: Also call move idle after this otherwise that might lead to problems i think

miraco: Never had problem while testing but i added it for safty

PR #741, opened 2025-05-16
Port Recast Demo mod to linux

LeFroid:
## 🍰 Pullrequest
This PR will allow linux users to test the recast demo mod

### Proof
- Shared some screenshots in Discord, can provide other proof as needed

### Issues
- Couldn't build previously, was disabled in CMake

### How2Test
- Run cmake with build flag -DBUILD_RECASTDEMOMOD=ON
- make -jX install 
- cd into install folder/bin/tools and run `./RecastDemoMod --dataDir <path to maps,mmaps,vmaps>`

### Todo / Checklist
- Test build again on Windows
- Test on another linux machine

### DIFF
--- a/contrib/recastdemomod/Source/Filelist.cpp
+++ b/contrib/recastdemomod/Source/Filelist.cpp
@@ -26,6 +26,53 @@
 #	include <dirent.h>
 #endif
 
+int is_match(char* line, char* pattern)
+{
+  int wildcard = 0;
+
+  do
+  {
+    if ((*pattern == *line) || (*pattern == '?'))
+    {
+      line++;
+      pattern++;
+    }
+    else if (*pattern == '*')
+    {
+      if (*(++pattern) == '\0')
+      {
+        return 1;
+      }
+      wildcard = 1;
+    }
+    else if (wildcard)
+    {
+      if (*line == *pattern)
+      {
+        wildcard = 0;
+        line++;
+        pattern++;
+      }
+      else
+      {
+        line++;
+      }
+    }
+    else
+    {
+      return 0;
+    }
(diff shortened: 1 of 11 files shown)


contrib/recastdemomod/Source/Filelist.cpp
```diff
@@ -55,10 +102,15 @@ void scanDirectory(const char* path, const char* ext, FileList& list)
 	while ((current = readdir(dp)) != 0)
 	{
 		int len = strlen(current->d_name);
-		if (len > 4 && strncmp(current->d_name+len-4, ext, 4) == 0) // TODO:: may not work due to joker(*) added
```
killerwife, line 58: This change stands out to me, what is its purpose, and have you validated it works on windows

killerwife:
I will test on windows, but likely not before sunday

contrib/recastdemomod/Source/Filelist.cpp
LeFroid: This is under a platform-specific block for *nix, so it shouldn't have an impact.
In CMaNGOS_Map::ScanFoldersForMaps, this function is called like this:
`scanDirectory(m_ctx->getMapFolder(), "*.map", m_MapFiles);`
The issue is that we are doing a string comparison on the file suffix (ex: "0011246.map" -> ".map") and a pattern ("*.map"). Needed to offset pattern by 1 to compare ".map" to ".map" instead of "*.map"

Separately, added an else condition for more complicated pattern matching inside CMaNGOS_Map::scanFoldersForTiles() where we need to load tiles where the file has both prefix and suffix (ex: "001*.mmtile")

PR #743, opened 2025-05-29
EventAI: add a simple InCombat check to EVENT_T_RECEIVE_AI_EVENT

miraco:
## 🍰 Pullrequest

This change will help recreating some OpenWorld RPs. 
As an example i will use the RP from Warp-Raider Nesaad with his 2 Zaxxis Stalker.

https://youtu.be/EY_ur_8Sf4U 

The first part of the Video where he turns to one of the Zaxxis Stalkers and says some text with emote, will happen even if the Stalker is Dead.
The 2nd Part where the Zaxxis Stalker will teleport away, should only happen if they are alive and not in combat.
AFAIK we currently don't support any inCombat checks for dbscripts .

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


al3xc1985:
well there are many open world rp's that can use  this feature. I can confirm it

PR #751, opened 2025-07-01
Fix: avoid invalid spline velocity after slow application

maxxbox81:
### Proof

- None

### Issues

- None

### How2Test

- None
spline error Screenshot 2025-06-30 112501

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Movement/MoveSplineInit.cpp
+++ b/src/game/Movement/MoveSplineInit.cpp
@@ -88,12 +88,24 @@ namespace Movement
 
         moveFlags |= (MOVEFLAG_SPLINE_ENABLED | MOVEFLAG_FORWARD);
 
-        if (args.velocity == 0.f) // ignore swim speed and flight speed because its not used in generic scripting - always possible to override
-        {
-            args.velocity = unit.GetSpeed(MovementInfo::GetSpeedType(MovementFlags(moveFlags & ~(MOVEFLAG_FLYING | MOVEFLAG_SWIMMING))));
-            if (args.slowed != 0.f) // when set always > 0.5
-                args.velocity *= args.slowed;
-        }
+        if (args.velocity <= 0.f)
+{
+    args.velocity = unit.GetSpeed(MovementInfo::GetSpeedType(MovementFlags(moveFlags & ~(MOVEFLAG_FLYING | MOVEFLAG_SWIMMING))));
+
+    // Sicherheitscheck: slowed muss im Bereich > 0.0 und = 1.0 liegen
+    if (args.slowed <= 0.f || args.slowed > 1.0f)
+        args.slowed = 1.0f;
+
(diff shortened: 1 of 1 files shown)


maxxbox81:
fixed Errors

src/game/Movement/MoveSplineInit.cpp
```diff
@@ -88,12 +88,24 @@ namespace Movement
...
         moveFlags |= (MOVEFLAG_SPLINE_ENABLED | MOVEFLAG_FORWARD);
 
-        if (args.velocity == 0.f) // ignore swim speed and flight speed because its not used in generic scripting - always possible to override
-        {
-            args.velocity = unit.GetSpeed(MovementInfo::GetSpeedType(MovementFlags(moveFlags & ~(MOVEFLAG_FLYING | MOVEFLAG_SWIMMING))));
-            if (args.slowed != 0.f) // when set always > 0.5
-                args.velocity *= args.slowed;
-        }
+        if (args.velocity <= 0.f)
+{
```
cpevors, line 92: It's a nitpick, but this block of code is less indented than the surrounding code.

src/game/Movement/MoveSplineInit.cpp
```diff
@@ -88,12 +88,24 @@ namespace Movement
...
+    args.velocity = unit.GetSpeed(MovementInfo::GetSpeedType(MovementFlags(moveFlags & ~(MOVEFLAG_FLYING | MOVEFLAG_SWIMMING))));
+
+    // Sicherheitscheck: slowed muss im Bereich > 0.0 und = 1.0 liegen
```
cpevors, line 95: I'd also recommend that this comment be translated to English, for consistency with the rest of the code base: "Safety check: slowed must be in the range > 0.0 and ≤ 1.0"

src/game/Movement/MoveSplineInit.cpp
```diff
@@ -88,12 +88,24 @@ namespace Movement
...
+    if (args.slowed <= 0.f || args.slowed > 1.0f)
```
cpevors, line 96: Would it not, in some way, make sense to replace Lines 96-101 with something similar to: 

```c++
            // Safety check: if unit is rooted, it shouldn't be slowed
            if (unit.IsRooted())
            {
                args.slowed = 1.0f;
            }
            else if (args.slowed <= 0.f || args.slowed > 1.0f)
            {
                args.slowed = 1.0f;
            }
```
            
In this way, rooted units bypass the slowdown logic entirely, which would make sense since, if a unit is rooted, they are already immobile.

maxxbox81:
I'm sorry, I'm completely new to github. You could also limit the error messages so that they only appear once per npc id and only in debug mode, then at least they wouldn't disappear in the background, as a suggestion

killerwife:
I believe i resolved this by fixing my low hp slow impl

PR #752, opened 2025-07-02
Fix: Improve LoadHelper stability with null check and debug logging

maxxbox81:
Added obj = nullptr to ensure default initialization and avoid undefined behavior.

Combined null and load failure check (if (!obj || !obj->LoadFromDB(...))) to prevent invalid access.

Moved IsInWorld() assertion to a conditional debug log using #ifdef MANGOS_DEBUG.

Replaced C-style cast with static_cast<T*> for improved type safety.

Improved overall code robustness and error tolerance without changing functionality.

### Proof

- None

### Issues

- None

### How2Test

- None
error 100
error 101

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Grids/ObjectGridLoader.cpp
+++ b/src/game/Grids/ObjectGridLoader.cpp
@@ -118,47 +118,61 @@ void LoadHelper(CellGuidSet const& guid_set, CellPair& cell, GridRefManager<T>& 
 
     for (uint32 guid : guid_set)
     {
-        T* obj;
+        T* obj = nullptr;
         uint32 newGuid = guid;
         if constexpr (std::is_same_v<T, GameObject>)
         {
             GameObjectData const* data = sObjectMgr.GetGOData(guid);
             MANGOS_ASSERT(data);
-            obj = (T*)GameObject::CreateGameObject(data->id);
+            obj = static_cast<T*>(GameObject::CreateGameObject(data->id));
             if (map->GetSpawnManager().IsEventGuid(guid, HIGHGUID_GAMEOBJECT))
                 newGuid = 0;
         }
         else
         {
-            obj = new T;
+            obj = new T();
             if (map->GetSpawnManager().IsEventGuid(guid, HIGHGUID_UNIT))
                 newGuid = 0;
         }
         // sLog.outString("DEBUG: LoadHelper from table: %s for (guid: %u) Loading",table,guid);
-        if (!obj->LoadFromDB(guid, map, newGuid, 0))
+        if (!obj || !obj->LoadFromDB(guid, map, newGuid, 0))
         {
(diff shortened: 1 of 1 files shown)


src/game/Grids/ObjectGridLoader.cpp
```diff
@@ -118,47 +118,61 @@ void LoadHelper(CellGuidSet const& guid_set, CellPair& cell, GridRefManager<T>&
 
     for (uint32 guid : guid_set)
     {
-        T* obj;
+        T* obj = nullptr;
```
killerwife, line 121: This is unnecessary. Should new fail, it will fail with exception. We populate the field with every flow.

src/game/Grids/ObjectGridLoader.cpp
```diff
@@ -118,47 +118,61 @@ void LoadHelper(CellGuidSet const& guid_set, CellPair& cell, GridRefManager<T>&
...
         else
         {
-            obj = new T;
+            obj = new T();
             if (map->GetSpawnManager().IsEventGuid(guid, HIGHGUID_UNIT))
                 newGuid = 0;
         }
         // sLog.outString("DEBUG: LoadHelper from table: %s for (guid: %u) Loading",table,guid);
-        if (!obj->LoadFromDB(guid, map, newGuid, 0))
+        if (!obj || !obj->LoadFromDB(guid, map, newGuid, 0))
```
killerwife, line 138: Again, null check unnecessary. We populate the field in every case.

src/game/Grids/ObjectGridLoader.cpp
```diff
@@ -118,47 +118,61 @@ void LoadHelper(CellGuidSet const& guid_set, CellPair& cell, GridRefManager<T>&
...
+        // Set cell info (used also by creatures)
+        addUnitState(obj, cell);
 
+        // Only gameobjects are added to grid here
         if (!obj->IsCreature())
-        {
             grid.AddGridObject(obj);
+        
+        // Check whether object is correctly in world
+        if (!obj->IsInWorld())
```
killerwife, line 152: If this happens, we have a critical problem in the core and need to shut down immediately. The assert is there for a critical reason.

src/game/Grids/ObjectGridLoader.cpp
```diff
@@ -118,47 +118,61 @@ void LoadHelper(CellGuidSet const& guid_set, CellPair& cell, GridRefManager<T>&
...
-
         obj->GetViewPoint().Event_AddedToWorld(&grid);
-
         if (bg)
             bg->OnObjectDBLoad(obj);
-
         ++count;
     }
-}
+}
```
killerwife, line 175: Empty space

PR #760, opened 2025-08-20
[Spell] Add script for z level fix for [DND] TAR Pedestal npcs

AnonXS:
pedestal

without script in the texture, with script at the hight it is where it is spawned with sniffed position.
seemingly should stand perfectly ontop, this seems to not be possible due to gobject clipping.

wit further adjustments it works sometimes, but not always.

grafik
grafik


killerwife:
Your PR is perfect, you were missing this:

Movement Flags: 1536 (DisableGravity, Root)

All the npcs hidden hover on B end. InhabitType 4 fixes it

PR #766, opened 2025-09-18
Refactor: Make function for generating Bitmasks

insunaa:
## 🍰 Pullrequest

This PR refactors uses of the pattern `1 << (enum_member - 1)` into a more generic `getBitmask(enum_member)` style.

Performance shouldn't be affected too badly.


killerwife:
Not gonna say this is bad outright, but the way to go here, is to add type specific functions, like hasattribute functions. Its easier to find them and use them with modern type checking cpp and avoids issues.

 Not saying how yet, but thats the idea, cos then u cant bitmask check wrong field (which is ofc issue in old code too)

The stances field would have a func for form enum specifically. I know its more code but less mistakes will happen

PR #771, opened 2025-10-08
Fix Snake Trap / Crash

devovh:
After Hunter uses Snake Trap on the target, a Server Crash occurs.


kelbren:
I get the same server crash on WotLK, I tested this change on WotLK and snake trap works correctly now. so this would need to be fixed there as well.

killerwife:
Appreciate the contribution, sorry for the delay. Life was fairly rough.

PR #800, opened 2026-04-11
BG/WSG: Fix flag carrier HUD worldstate using wrong team index

henhouse:
Fixes issue where the HUD showing which flag was actively being carried was reversed. Alliance picking up Horde flag showed the Alliance flag was being carried by the Horde team, and vice versa.


killerwife:
Ugh, I hate these bugs. The rewrite for worldstates was necessary, but I caused too many of these in the process.

PR #805, opened 2026-04-19
Fix Ruthlessness bug with Slice and Dice

### LINKED ISSUE #3927: 🐛 [Bug Report] TBC The ability is not working correctly
### Bug Details

1 . https://www.wowhead.com/wotlk/spell=14160/ruthlessness
2 . We hit 2 combo points
3 . using  https://www.wowhead.com/wotlk/spell=5171/slice-and-dice  
4 . 2 combo points appear . It should be 1

### Steps to Reproduce

1.1 . https://www.wowhead.com/wotlk/spell=14160/ruthlessness
2 . We hit 2 combo points
3 . using  https://www.wowhead.com/wotlk/spell=5171/slice-and-dice  
4 . 2 combo points appear . It should be 1
2.
3.
4.

### Expected behavior

_No response_

### Suggested Workaround

_No response_

### Crash Log

_No response_

### Core SHA1 Commit Hash

TBC

### Database SHA1 Commit Hash

TBC

### Operating System

Win

### Client Version

2.4.3 (The Burning Crusade)

Velite12:
Fix Slice and Dice activating Ruthlessness twice. 

Slice and Dice casts on both self and target with combo points which activates the passive twice as often as other finishers and can even result in 2 combo points after cast.

This fix ensures that if Ruthlessness activates but if the target is self (the rogue casting slice and dice), it does not award the combo point. That way, only the cast on the target with existing combo points awards proc.

### Proof

- None

### Issues

fixes https://github.com/cmangos/issues/issues/3927

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Spells/UnitAuraProcHandler.cpp
+++ b/src/game/Spells/UnitAuraProcHandler.cpp
@@ -2475,7 +2475,7 @@ SpellAuraProcResult Unit::HandleProcTriggerSpellAuraProc(ProcExecutionData& data
             {
                 // Only trigger if you are not the target (Slice and Dice casts twice,
                 // ignore on you, keep on target)
-                if (pVictim->GetGUIDLow() != GetGUIDLow()) 
+                if (pVictim->GetObjectGuid() != GetObjectGuid()) 
                 {
                     spell->AddTriggeredSpell(trigger_spell_id);
                     return SPELL_AURA_PROC_OK;
--- a/src/game/Spells/UnitAuraProcHandler.cpp
+++ b/src/game/Spells/UnitAuraProcHandler.cpp
@@ -2475,7 +2475,7 @@ SpellAuraProcResult Unit::HandleProcTriggerSpellAuraProc(ProcExecutionData& data
             {
                 // Only trigger if you are not the target (Slice and Dice casts twice,
                 // ignore on you, keep on target)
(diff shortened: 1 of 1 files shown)


src/game/Spells/UnitAuraProcHandler.cpp
```diff
@@ -2473,8 +2473,16 @@ SpellAuraProcResult Unit::HandleProcTriggerSpellAuraProc(ProcExecutionData& data
             // Need add combopoint AFTER finishing move (or they get dropped in finish phase)
             if (Spell* spell = GetCurrentSpell(CURRENT_GENERIC_SPELL))
             {
-                spell->AddTriggeredSpell(trigger_spell_id);
-                return SPELL_AURA_PROC_OK;
+                // Only trigger if you are not the target (Slice and Dice casts twice,
+                // ignore on you, keep on target)
+                if (pVictim->GetName() != GetName())
```
killerwife: String comparison is very bad.

You can just compare pointers (victim and THIS)

Or object guids.

Velite12:
Alright let's do GetGUIDLow() instead.

insunaa:
Just use ObjectGuid

Velite12:
Alright let's do GetObjectGuid() instead.

PR #806, opened 2026-04-19
[Molten Core] Implement Core Hound (11671) collapse/reignite.

evil-at-wow:
## 🍰 Pullrequest

I recently noticed that we don't seem to have the Core Hound's collapse/reignite mechanic implemented in Molten Core, so currently you can easily kill these packs by just killing them individually, rather than all within 10 seconds.

This PR is my implementation of this mechanic, but given I'm not familiar with core scripts, I'd like to have this reviewed before I push this.

### Proof
See for example the description [here](https://warcraft.wiki.gg/wiki/Core_Hound_(Molten_Core)).

### How2Test
* Enter Molten Core, go to the trash packs before Lucifron.
* Try to kill a Core Hound pack.
* Notice Core Hounds now collapse when you 'kill' them, but they reignite after 10s if any of the others are still alive at that time.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/molten_core/molten_coreScripts.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/molten_core/molten_coreScripts.cpp
@@ -24,6 +24,7 @@ EndScriptData */
 /* ContentData
 EndContentData */
 
+#include "AI/EventAI/CreatureEventAI.h"
 #include "AI/ScriptDevAI/include/sc_common.h"
 #include "molten_core.h"
 
@@ -54,6 +55,94 @@ bool GOUse_go_molten_core_rune(Player* /*pPlayer*/, GameObject* pGo)
     return true;
 }
 
+struct npc_core_hound : public CreatureEventAI
+{
+    // Action IDs
+    static constexpr uint32 actionReigniteTimer = 0;
+
+    // Spells
+    static constexpr uint32 spellFullHeal = 17683;
+    static constexpr uint32 spellPlayDead = 19822;
+    static constexpr uint32 spellFireNovaVisual = 19823;
+    static constexpr uint32 spellPacifySelf = 19951;
+
+    npc_core_hound(Creature* creature) : CreatureEventAI(creature)
+    {
+        SetDeathPrevention(true);
+        AddCustomAction(actionReigniteTimer, true, [this] { HandleReigniteTimer(); });
+    }
+
+    void Reset() override
+    {
+        CreatureEventAI::Reset();
+        SetDeathPrevention(true);
+    }
+
+    // Reignite, exiting a fake death.
+    void Reignite()
+    {
+        m_creature->SetStandState(UNIT_STAND_STATE_STAND);
+        m_creature->RemoveAurasDueToSpell(spellPacifySelf);
+        m_creature->RemoveAurasDueToSpell(spellPlayDead);
+        DoCastSpellIfCan(m_creature, spellFullHeal, CAST_TRIGGERED);
+        DoResetThreat();
+        SetCombatMovement(true);
+
+        SetDeathPrevention(true);
+
+        // %s reignites from the heat of another Core Hound!
+        DoBroadcastText(7867, m_creature);
+    }
+
+    // Collapse, entering a fake death.
+    // Notes:
+    // * You can still target a collapsed Core Hound.
+    // * You can still damage a collapsed Core Hound.
+    // * They keep debuffs (Curse of the Elements, Sunders, etc).
+    // To do:
+    // * Confirm they keep combo points.
+    void Collapse()
+    {
+        // %s collapses and begins to smolder.
+        DoBroadcastText(7866, m_creature);
+
+        ResetTimer(actionReigniteTimer, 10000);
+
+        DoCastSpellIfCan(m_creature, spellPlayDead, CAST_TRIGGERED);
+        DoCastSpellIfCan(m_creature, spellPacifySelf, CAST_TRIGGERED);
+
+        m_creature->SetTarget(nullptr);
+        SetCombatMovement(false);
+        m_creature->SetStandState(UNIT_STAND_STATE_DEAD);
+    }
+
+    void JustPreventedDeath(Unit* /*attacker*/) override
+    {
+        Collapse();
+    }
+
+    void HandleReigniteTimer()
+    {
+        // Reignite if there are still other core hounds alive 10 seconds after the collapse.
+        std::list<Creature*> coreHounds;
+        GetCreatureListWithEntryInGrid(coreHounds, m_creature, 11671 /* Core Hound */, 40.0f);
+        for (Creature* coreHound : coreHounds)
+        {
+            // Ignore alive core hounds in groups not in combat with us.
+            if (coreHound && coreHound->IsInCombat() && coreHound->GetHealth() > 1)
+            {
+                Reignite();
+                DoCastSpellIfCan(m_creature, spellFireNovaVisual, CAST_TRIGGERED);
+                return;
+            }
+        }
+
+        // No other core hounds alive: really die (and despawn the corpse).
+        m_creature->Suicide();
+        m_creature->ForcedDespawn();
+    }
+};
+
 void AddSC_molten_core()
 {
     Script* pNewScript;
@@ -61,5 +150,10 @@ void AddSC_molten_core()
     pNewScript = new Script;
     pNewScript->Name = "go_molten_core_rune";
     pNewScript->pGOUse = &GOUse_go_molten_core_rune;
+    pNewScript->RegisterSelf();
+
+    pNewScript = new Script;
+    pNewScript->Name = "npc_core_hound";
+    pNewScript->GetAI = &GetNewAIInstance<npc_core_hound>;
     pNewScript->RegisterSelf();
 }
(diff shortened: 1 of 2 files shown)


src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/molten_core/molten_coreScripts.cpp
```diff
@@ -54,6 +55,111 @@ bool GOUse_go_molten_core_rune(Player* /*pPlayer*/, GameObject* pGo)
     return true;
 }
 
+struct npc_core_hound : public CreatureEventAI
+{
+    npc_core_hound(Creature* creature)
```
killerwife: So I believe the codestyle is either all in a line, or the comma after the )

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/molten_core/molten_coreScripts.cpp
```diff
@@ -54,6 +55,111 @@ bool GOUse_go_molten_core_rune(Player* /*pPlayer*/, GameObject* pGo)
...
+    }
+
+    void UpdateAI(const uint32 diff) override
+    {
+        CreatureEventAI::UpdateAI(diff);
+
+        // Reignite if there are still other core hounds alive 10 seconds after the collapse.
+        if (m_collapsed)
+        {
+            if (m_reigniteTimer < diff)
```
killerwife: This workflow usually lead to issues. Generally trying to stay away from 

void UpdateAI(const uint32 diff) override

U can do it in this case by adding smth like:

```
AddCustomAction(JULIANNE_SUMMON_ROMULO, true, [&]() { DoSummonRomulo(); });

```

In constructor and then:

```
ResetTimer(JULIANNE_SUMMON_ROMULO, 12000);


```

At the point of starting it

Which will then do all you tried to do in a more future proof manner.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/molten_core/molten_coreScripts.cpp
```diff
@@ -54,6 +55,111 @@ bool GOUse_go_molten_core_rune(Player* /*pPlayer*/, GameObject* pGo)
...
+        m_collapsed = true;
+
+        m_creature->InterruptNonMeleeSpells(false);
+        m_creature->StopMoving();
+        m_creature->SetTarget(nullptr);
+        SetCombatMovement(false);
+        SetCombatScriptStatus(true);
+        SetMeleeEnabled(false);
+        m_creature->SetStandState(UNIT_STAND_STATE_DEAD);
+        m_creature->SetFlag(UNIT_DYNAMIC_FLAGS, UNIT_DYNFLAG_DEAD);
```
killerwife: This dynflag i dont see being used in sniff. Do u have evidence for that?

Tobschinski:
doubt it matters much for visuals and can probably ignored.. but they do use SpellIDs 17683, 19951 and 19822 for their stuff..

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/molten_core/molten_coreScripts.cpp
evil-at-wow: No, it was just a flag I tried, because without that flag things didn't look just right in the client. In the sense that the Core Hound didn't look death while that was clearly the case when I played Classic, and this flag "fixed" that. At the same time the 2.4.3 client is a totally different beast compared to what we had in 2019.

But you are absolutely correct questioning things that should be visible in sniffs! Somehow I was so busy with a script that did what I wanted, that I totally forgot to check everything visible in sniffs. In the ones I have in my little archive, the 1 hp when collapsed is correct, and they also get an update that removes their target, sets UNIT_FIELD_BYTES_1 to have UNIT_STAND_STATE_DEAD (so all correct) and sets unit_field_flags to unit_flag_in_combat + unit_flag_pet_in_combat + unit_flag_immune_to_npc it seems.

I also discovered they cast spell 19951 (Pacify Self) when collapsing... so I'm going to change the implementation based on this. I haven't found the reignite case yet though, to see what spell they use to "heal", if any. [Edit: Tobschinski just provided it while I was typing this :smile:]

evil-at-wow:
@Tobschinski Thanks a lot for that! I had discovered the use of 19951 (Pacify Self) in the meantime, but I didn't find the reignite case yet in my little archive so I was still unsure about a heal spell, if any. Looks like that heal spell you list fits perfectly.

And I think it does perhaps matter for visuals. That 19822 'Play Dead' spell is most likely what's going to make them appear dead, which is not the case now unless I use that UNIT_DYNFLAG_DEAD flag killerwife was questioning.

killerwife:
Yes that aura certainly is a whole another matter. If thats used, thats all good.

evil-at-wow:
I pushed an updated version:
* The timer and collapsed members are gone; they have been replaced by a custom action as suggested during review.
* The UNIT_DYNFLAG_DEAD and UNIT_FLAG2_FEIGN_DEATH flags are gone too (at least explicitly setting them).
* This is now using spells, and this also avoids the need for explicitly calling StopMoving(), InterruptNonMeleeSpells(), SetMeleeEnabled() and so (taken care of by the Pacify Self basically).

This seems to work as intended, except for one issue: the "Play Dead" aura is visible on collapsed Core Hounds and that shouldn't be the case (duh). I still have to look into that.

I've opted to use the more modern approach rather than a 90s enum to avoid getting storage for a few "magic constants" like spells. I can of course change that again if you're absolutely against that (like because everything else in this core (ab)uses enum).

@Tobschinski What version did you see those spells in? What packet? Because in my data from version 3.0.8 I don't seem to find any trace of spell 19822 nor of the associated feign death flag. Yet it makes perfect sense because that's exactly what causes a collapsed dog to show as 'dead' in the client, so I must be looking at the wrong things...

killerwife:
Two ideas here. Isnt he meant to be unselectable? That would make the aura not visible. I also have a sniff that has unit state dead rather than feign death aura. Will send it to u tomorrow from classic.

evil-at-wow:
> Isnt he meant to be unselectable?

No, you can definitely target them when they are collapsed (also in Classic). That's how I know for sure they look dead in the client and also keep their debuffs (warlock curses, sunders, etc).

> I also have a sniff that has unit state dead rather than feign death aura.

The UNIT_STAND_STATE_DEAD is indeed sent to the client, I see that in my 3.0.8 sniffs too. I didn't see the feign death aura in that sniff yet (but admittedly the tools I wrote back in the day are seriously lacking when it comes to searching and finding stuff like this), which is why I was asking Tobschinski about that spell.

If it wasn't Tobschinksi bringing it up, and the fact the mob doesn't show in the client as dead as it should be without it, I'd drop the use of that feign death spell and call it a day. But it does fix a visual thing, so there's that.

So more research needed...

killerwife:
Check your email, i sent u a Season of Mastery sniff which is same as vanilla for this purpose

evil-at-wow:
In the meantime I had a close look at your sniff, and it's essentially matching my own 3.x data. So no sign of that 'Play Dead' spell in your data either. The only difference, which is an interesting observation by itself, is they still do threat updates while they are collapsed and people continue to damage them. That different is not really a surprise given 3.x didn't share actual threat values, but it suggests they don't reset threat when reigniting (or collapsing).

At this point I'm thinking some of this could be a difference between the early clients and 3.x and later clients (including Classic, SoM, etc, given these are based on more modern clients), so I'm going to run a few more experiments with a 3.x server/client next.

evil-at-wow:
I finally managed to run some tests with a 3.3.5 server/client, and I see the exact same thing there as in my TBC setup. So we can rule out a difference between 3.x and earlier clients related to this.

I went on to checking a bunch of videos, and several suggest that the collapsed hounds aren't really shown as dead by the client (some UIs merely seem to suggest this, kind of). In particular, and in contrast to other trash mobs and bosses killed, the word 'dead' doesn't appear in the default client UI, and neither does hovering over a collapsed hound show 'corpse'.

So I've decided I'm going to get rid of that 'Play Dead' spell again for now. Then the pack mechanic is there and everything is visually OK. If more evidence comes to light later, one way or another, we can still adjust things.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/molten_core/molten_coreScripts.cpp
```diff
@@ -54,6 +55,91 @@ bool GOUse_go_molten_core_rune(Player* /*pPlayer*/, GameObject* pGo)
...
+        m_creature->SetStandState(UNIT_STAND_STATE_STAND);
+        m_creature->RemoveAurasDueToSpell(spellPacifySelf);
+        DoCastSpellIfCan(m_creature, spellFullHeal, CAST_TRIGGERED);
+        DoResetThreat();
+        SetCombatMovement(true);
+
+        SetDeathPrevention(true);
+
+        // %s reignites from the heat of another Core Hound!
+        DoBroadcastText(7867, m_creature);
```
killerwife, line 92: Either add these as constexpr or an enum above. Generally prefer not to have text magic numbers without naming.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/molten_core/molten_coreScripts.cpp
```diff
@@ -54,6 +55,91 @@ bool GOUse_go_molten_core_rune(Player* /*pPlayer*/, GameObject* pGo)
...
+    void JustPreventedDeath(Unit* /*attacker*/) override
+    {
+        Collapse();
+    }
+
+    void HandleReigniteTimer()
+    {
+        // Reignite if there are still other core hounds alive 10 seconds after the collapse.
+        std::list<Creature*> coreHounds;
+        GetCreatureListWithEntryInGrid(coreHounds, m_creature, 11671 /* Core Hound */, 40.0f);
```
killerwife, line 125: Same here with the 11671 which i believe is npc entry

killerwife (changes requested):

PR #807, opened 2026-04-30
Add Energizing Spores Spell 35336 to positive effects list

Velite12:
## 🍰 Pullrequest

Add Energizing Spores spell 35336 to positive effects list

### Proof

https://youtu.be/FsaskRGj7pM?t=548

The spell should also be listed as a debuff and not a buff, this requires an sql change.
You can find the associated sql change in this issue here: 
https://github.com/cmangos/issues/issues/4055

It increases your stats and is therefore always a positive effect.


killerwife:
The bot code is wrong on this one. Bot developers never started accounting for new reversal discoveries, i.e.:

SPELL_ATTR_EX4_AURA_IS_BUFF

IsPositiveEffect is our old function that attempted to emulate without further knowledge. As such this commit is redundant, because you need to check the proper function, which is IsPositiveAuraEffect

(for purposes of effect, this can still be negative, as it should, since it is explicitly enemy targeted TARGET_ENUM_UNITS_ENEMY_AOE_AT_SRC_LOC, but for aura its overriden as buff)

PR #811, opened 2026-05-02
Build: Formalize ICU dependency from Boost

### LINKED ISSUE #4056: 🐛 [Bug Report] Silent libicu dependency
### Bug Details

We currently have an issue where one of our dependencies (I think it's boost) itself depends on libicu-dev or similar, but not all package managers automatically resolve this dependency. I think it would make sense to manually add a find_library or whatever in cmake to make sure cmake fails if the dependency is missing, instead of the compilation process failing in the linking stage

### Steps to Reproduce

1. Compile CMaNGOS on Ubuntu 26.04 without libicu-dev manually installed

### Expected behavior

CMake should fail when the dependency is not found and present a meaningful error message.

### Suggested Workaround

_No response_

### Crash Log

_No response_

### Core SHA1 Commit Hash

6f58c89fad771097198b6f629f8d2fd335ef1a69

### Database SHA1 Commit Hash

6272cd8eb85d7f77fd60c786b2e461e1d3d0497e

### Operating System

Ubuntu 26.04

### Client Version

2.4.3 (The Burning Crusade)

insunaa:
## 🍰 Pullrequest

This PR makes sure the ICU dependency we inherit from Boost::Regex is formally added in CMake, so error messages when the dependency is missing are easy to understand

Depends on commit https://github.com/cmangos/mangos-tbc/commit/cec7e209319fcf6fc7c469c74e750aa1a80ac925

### Issues

- fixes https://github.com/cmangos/issues/issues/4056

### How2Test

- Compile CMaNGOS on a Linux or MacOS system without libicu installed

### Todo / Checklist

- [X] None

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,7 +206,7 @@ endif()
 
 # Find needed packages and if necessery abort if something important is missing
 find_package(Boost 1.70.0 REQUIRED COMPONENTS program_options thread regex serialization filesystem)
-if (APPLE)
+if (UNIX OR APPLE)
     find_package(ICU COMPONENTS uc i18n data REQUIRED)
     add_library(icuuc ALIAS ICU::uc)
     add_library(icui18n ALIAS ICU::i18n)


insunaa:
Alternatively we can probably remove `regex` from the boost required components list. I don't see us including the boost regex library anywhere and if I remove it from CMakeLists.txt cmangos still builds.

Edit: Don't seem to need `serialization` component, either.

Proposal to change it to:
```cmake
find_package(Boost 1.70.0 REQUIRED COMPONENTS program_options thread filesystem chrono atomic date_time container)
```

killerwife:
image
image

We have a different problem. We likely want to remove STD::REGEX in favor of the boost one. The standard one is cursed.

insunaa:
What is cursed about `std::regex`?

killerwife:
Im just gonna link the first comment i find for you on google:

https://www.reddit.com/r/cpp/comments/1iwqsd0/what_are_the_gory_details_of_why_stdregex_being/

And there is hundreds of threads like this, and it will never be fixed. Meanwhile boost regex has none of these issues.

killerwife:
Merging this, but the bigger picture of using boost regex instead of std one remains.

PR #813, opened 2026-05-07
Fix issue with The Vengeful Harbinger quest not being completable

Velite12:
Fixes the issue with The Vengeful Harbinger quest where two gameobjects which should spawn at the end do not spawn, one of which is needed to finish the quest.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/terokkar_forest.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/terokkar_forest.cpp
@@ -1366,9 +1366,11 @@ enum
     POINT_ORB_WAYPOINT = 1,
 
     NPC_ORB_WAYPOINT_1 = 21443,
+    NPC_PORTAL_TRIGGER = 21463,
     NPC_DRAENEI_TOMB_GUARDIAN = 22285,
 
     DBSCRIPT_VENGEFUL_HARBINGER_FAKE_DEATH = 10063,
+    DBSCRIPT_RESPAWN_ASCENDANT_FLOORPIECE = 10067,
 
     EVENT_RESET_TIMER = 120000,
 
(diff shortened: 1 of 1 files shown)


killerwife:
What is this script:

DBSCRIPT_RESPAWN_ASCENDANT_FLOORPIECE = 10067,

Isnt it launched already from waypoints of a mob in DB?

Because we wouldnt have an unused relay script lying around i think

Velite12:
> What is this script:
> 
> DBSCRIPT_RESPAWN_ASCENDANT_FLOORPIECE = 10067,
> 
> Isnt it launched already from waypoints of a mob in DB?
> 
> Because we wouldnt have an unused relay script lying around i think

It's not, or I wouldn't have to make this commit.

Velite12:
It's possible that the commented out block of code above may at one point have triggered the GO to spawn for quest completion.

Velite12:
Also this is the particular mob with the waypoints, the harbinger. It's once they reach the end of the waypoints, where they get zapped, that's the time from which it should spawn. And if you look at the video you'll see that the way I have it is exactly when it should spawn because it coordinates with the floating purple orbs.

https://www.youtube.com/watch?v=jLhPHKOnbjM

CaptainJaySolo:
Tested this on my local core and it works well from what I've tested.
I had the GO spawned all the time as a work around, but I've replaced it with this method. Thanks.

killerwife:
I finally got around to this. Nothing inherently with the PR but the bigger issue is this.

The relay script is meant to be launched from here:
`select * from creature_movement_template where ScriptId=2163801;`
And its set to launch from same point as C++ so question is why is that broken?

Either you get to it first or i will, just letting you know.

Velite12:
Yeah you can get to it I moved on past this a while ago.

YU-Scuti:
The issue arises because "SCRIPT COMMAND SPAWN GAMEOBJECT" unconditionally creates std::vector<WorldObject*> sources; and std::vector<WorldObject*> targets;, while the database despawns those sources and targets after two seconds.
스크린샷 2026-06-28 163334

To resolve the issue, modify "SCRIPT COMMAND SPAWN GAMEOBJECT" so that it ignores sources and targets when executing.

Alternatively, configure the sources and targets in the database.
스크린샷 2026-06-28 171438

PR #817, opened 2026-05-17
Fix Ring of Blood quests fail condition incorrect behavior

Velite12:
Fix issue with Nagrand Ring of Blood quests where the quest would fail if you died. The quest should fail if the summoned creature is not killed aka despawns whether through evade by leashing or the party wipes.

This way, even if you've died, you will still receive credit for the kill if the rest of the group can defeat the creature.

Evidence:

https://www.youtube.com/watch?v=nSMuu8PZOi4
https://www.youtube.com/watch?v=BbPOSBOH29U

The video from retail tbc actually shows it is possible to receive kill credit even if the kill is not valid for you (greyed out npc), don't know how to do that.

The video from classic shows that you still receive completion credit even when you are dead. Since it's not possible to complete a failed quest, then the quest doesn't require you to stay alive to complete.

https://www.youtube.com/watch?v=4u2cLXNjmz0&t=3s

In this video you can see at the end that the quest is eventually failed later, after all of the combatants are killed, not immediately when you die like current behavior.

This change needs to be accompanied with the following DBC change to remove QUEST_FLAGS_STAY_ALIVE in Quest_flags or it will still fail when a player dies.

UPDATE `tbcmangos`.`quest_template` SET `QuestFlags` = '134' WHERE (`entry` = '9962');
UPDATE `tbcmangos`.`quest_template` SET `QuestFlags` = '134' WHERE (`entry` = '9967');
UPDATE `tbcmangos`.`quest_template` SET `QuestFlags` = '134' WHERE (`entry` = '9970');
UPDATE `tbcmangos`.`quest_template` SET `QuestFlags` = '134' WHERE (`entry` = '9972');
UPDATE `tbcmangos`.`quest_template` SET `QuestFlags` = '134' WHERE (`entry` = '9973');
UPDATE `tbcmangos`.`quest_template` SET `QuestFlags` = '134' WHERE (`entry` = '9977');


Velite12:
Just want to add that the most important thing is the DB change as currently the quest series is much more difficult than on classic or original tbc as it requires you to stay alive, in addition to killing the creature. If you die then the group will have to either redo the fight or move on without you. You may even have to repeat the same quest many times until everyone has completion...

src/game/AI/ScriptDevAI/scripts/outland/world_outland.cpp
```diff
@@ -291,6 +291,8 @@ struct world_map_outland : public ScriptedMap, public TimerManager
     std::vector<BashirCombatSpawn*> m_bashirPhaseSpawnDataUsed;
     // Shade of the Horseman village attack event
     ShadeOfTheHorsemanData m_shadeData;
+    // Ring of blood
+    ObjectGuid m_playerGuid2;
```
killerwife: Its used only for ring of blood, name it as such. Otherwise PR is ok.

killerwife:
Pls make pr to tbc-db after this

PR #819, opened 2026-05-25
fix castspell server crash

### LINKED ISSUE #4079: 🐛 [Bug Report] Class with pet out causes server crash when charmed
### Bug Details

When a player has a pet out and is charmed, the server crashes

### Steps to Reproduce

1. Be class with pet
2. Take pet out
3. Get chamed by npc spell (for example, time lost controllers in sethekk halls drop totem which charms you)
4. server crash

### Expected behavior

Server should not crash

### Suggested Workaround

_No response_

### Crash Log

_No response_

### Core SHA1 Commit Hash

latest

### Database SHA1 Commit Hash

latest

### Operating System

windows

### Client Version

2.4.3 (The Burning Crusade)

Velite12:
Fix bug where a unit with a pet when charmed by castspell would crash the server because victim is nullptr and we're trying to access functions of nullptr.

Fixes https://github.com/cmangos/issues/issues/4079

~~TODO: Pets should stop attacking their current target and instead attack the target that the charmed unit is attacking.~~

Pets are working for the most part. It seems to be random whether pet continues to attack same target or attacks targets hostile to the charmed unit. I'll say it's close enough to correct.

### DIFF
--- a/src/game/AI/BaseAI/PetAI.cpp
+++ b/src/game/AI/BaseAI/PetAI.cpp
@@ -292,7 +292,7 @@ std::vector<std::tuple<SpellEntry const*, Unit*, bool>> PetAI::PickSpellWithTarg
         }
     }
 
-    uint32 creatureEntry = m_pet->getPetType() == HUNTER_PET ? 1 : m_creature->GetCreatureInfo()->Entry;
+    uint32 creatureEntry = (m_pet != nullptr && m_pet->getPetType() == HUNTER_PET) ? 1 : m_creature->GetCreatureInfo()->Entry;
     // Cast a spell from autocast selection
     for (uint8 i = 0; i < m_creature->GetPetAutoSpellSize(); ++i)
     {
--- a/src/game/AI/BaseAI/PetAI.cpp
+++ b/src/game/AI/BaseAI/PetAI.cpp
@@ -292,7 +292,7 @@ std::vector<std::tuple<SpellEntry const*, Unit*, bool>> PetAI::PickSpellWithTarg
         }
     }
 
-    uint32 creatureEntry = (m_pet != nullptr && m_pet->getPetType() == HUNTER_PET) ? 1 : m_creature->GetCreatureInfo()->Entry;
+    uint32 creatureEntry = (m_pet && m_pet->getPetType() == HUNTER_PET) ? 1 : m_creature->GetCreatureInfo()->Entry;
     // Cast a spell from autocast selection
     for (uint8 i = 0; i < m_creature->GetPetAutoSpellSize(); ++i)
     {
--- a/src/game/AI/PlayerAI/PlayerAI.cpp
+++ b/src/game/AI/PlayerAI/PlayerAI.cpp
@@ -48,7 +48,7 @@ void PlayerAI::AddPlayerSpellAction(uint32 spellId, std::function<Unit*()> selec
             if (IsNextMeleeSwingSpell(spellInfo))
                 selector = [&]()->Unit* { return m_player->GetVictim(); };
             else if (HasSpellTarget(spellInfo, TARGET_UNIT_ENEMY) || (spellInfo->Targets & TARGET_FLAG_DEST_LOCATION)) // always random
-                selector = [&, spellId = spellInfo->Id]()->Unit* { return m_player->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0, spellId, SELECT_FLAG_PLAYER); };
+                selector = [&, spellId = spellInfo->Id]()->Unit* { return m_player->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0, spellId, SELECT_FLAG_IN_LOS); };
             if (HasSpellTarget(spellInfo, TARGET_UNIT_FRIEND) || HasSpellTarget(spellInfo, TARGET_UNIT_FRIEND_CHAIN_HEAL)) // heals only target self
                 selector = [&]()->Unit* { return m_player; };
         }
--- a/src/game/AI/PlayerAI/PlayerAI.cpp
+++ b/src/game/AI/PlayerAI/PlayerAI.cpp
@@ -48,7 +48,7 @@ void PlayerAI::AddPlayerSpellAction(uint32 spellId, std::function<Unit*()> selec
             if (IsNextMeleeSwingSpell(spellInfo))
                 selector = [&]()->Unit* { return m_player->GetVictim(); };
             else if (HasSpellTarget(spellInfo, TARGET_UNIT_ENEMY) || (spellInfo->Targets & TARGET_FLAG_DEST_LOCATION)) // always random
-                selector = [&, spellId = spellInfo->Id]()->Unit* { return m_player->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0, spellId, SELECT_FLAG_IN_LOS); };
+                selector = [&, spellId = spellInfo->Id]()->Unit* { return m_player->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0, spellId, SELECT_FLAG_PLAYER); };
             if (HasSpellTarget(spellInfo, TARGET_UNIT_FRIEND) || HasSpellTarget(spellInfo, TARGET_UNIT_FRIEND_CHAIN_HEAL)) // heals only target self
                 selector = [&]()->Unit* { return m_player; };
         }
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -1565,8 +1565,13 @@ SpellCastResult Unit::CastSpell(Unit* Victim, SpellEntry const* spellInfo, uint3
     SpellCastTargets targets;
     targets.setUnitTarget(Victim);
 
-    if ((spellInfo->Targets & TARGET_FLAG_DEST_LOCATION) && Victim)
-        targets.setDestination(Victim->GetPositionX(), Victim->GetPositionY(), Victim->GetPositionZ());
+    if ((spellInfo->Targets & TARGET_FLAG_DEST_LOCATION))
+    {
+        if (victim)
+            targets.setDestination(Victim->GetPositionX(), Victim->GetPositionY(), Victim->GetPositionZ());
+        else
+            sLog.outError("CastSpell: victim was nullptr but tried to get position: caster %s, spellId %i", GetGuidStr().c_str(), spellInfo->Id);
+    }
     if (spellInfo->Targets & TARGET_FLAG_SOURCE_LOCATION)
         if (WorldObject* caster = spell->GetCastingObject())
             targets.setSource(caster->GetPositionX(), caster->GetPositionY(), caster->GetPositionZ());
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -1567,7 +1567,7 @@ SpellCastResult Unit::CastSpell(Unit* Victim, SpellEntry const* spellInfo, uint3
 
     if ((spellInfo->Targets & TARGET_FLAG_DEST_LOCATION))
     {
-        if (victim)
+        if (Victim)
             targets.setDestination(Victim->GetPositionX(), Victim->GetPositionY(), Victim->GetPositionZ());
         else
             sLog.outError("CastSpell: victim was nullptr but tried to get position: caster %s, spellId %i", GetGuidStr().c_str(), spellInfo->Id);
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -1565,7 +1565,7 @@ SpellCastResult Unit::CastSpell(Unit* Victim, SpellEntry const* spellInfo, uint3
     SpellCastTargets targets;
     targets.setUnitTarget(Victim);
 
-    if (spellInfo->Targets & TARGET_FLAG_DEST_LOCATION)
+    if ((spellInfo->Targets & TARGET_FLAG_DEST_LOCATION) && Victim)
         targets.setDestination(Victim->GetPositionX(), Victim->GetPositionY(), Victim->GetPositionZ());
     if (spellInfo->Targets & TARGET_FLAG_SOURCE_LOCATION)
         if (WorldObject* caster = spell->GetCastingObject())
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -1569,7 +1569,10 @@ SpellCastResult Unit::CastSpell(Unit* Victim, SpellEntry const* spellInfo, uint3
     {
         // This shouldn't happen, but we should return gracefully if it does...
         if (!Victim)
+        {
+            sLog.outError("CastSpell: victim was nullptr but tried to get position: caster %s, spellId %i", GetGuidStr().c_str(), spellInfo->Id);
             return SPELL_FAILED_BAD_TARGETS;
+        }    
         targets.setDestination(Victim->GetPositionX(), Victim->GetPositionY(), Victim->GetPositionZ()); 
     }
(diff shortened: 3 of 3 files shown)


killerwife:
This needs to be made into a check and log and error out situation. The spell shouldnt even launch when it has implicit dest that isnt supplied

Velite12:
`targets.setUnitTarget(Victim);` handles Nullptr check in that function. Why exactly charms are giving a nullptr victim when player has a pet out escapes me. I can't really check in the case of the one in Sethekk halls because that's a dbscript.

Velite12:
image
There are several of these in boss_blackheart_the_inciter.cpp

Velite12:
image
HandleModCharm is problematic if the target is nullptr too...

Velite12:
I think that should help get to the root of the issue in the case of the charm bug.

killerwife:
Aura GetTarget can never be nullptr in any handler or any script handler. Thats a guarantee.

killerwife:
Also laugh periodic on inciter doesnt have dest location implicit target flag on spell. So thats correct.

src/game/AI/BaseAI/PetAI.cpp
```diff
@@ -292,7 +292,7 @@ std::vector<std::tuple<SpellEntry const*, Unit*, bool>> PetAI::PickSpellWithTarg
         }
     }
 
-    uint32 creatureEntry = m_pet->getPetType() == HUNTER_PET ? 1 : m_creature->GetCreatureInfo()->Entry;
+    uint32 creatureEntry = (m_pet != nullptr && m_pet->getPetType() == HUNTER_PET) ? 1 : m_creature->GetCreatureInfo()->Entry;
```
killerwife: PetAI should never have m_pet nullptr. For which npc did this occur.

Velite12: > PetAI should never have m_pet nullptr. For which npc did this occur.

image


Yes, it can be nullptr.

This is for any charm spell that NPCs do. I tested Sethekk Halls npc charm totem and Blackheart the Inciter, both caused the same crash.

Velite12: My guess is that there is something in the charm logic which unlinks the pet to the hunter or warlock. Sometimes, the pets will be mind controlled and sometimes they won't be but that is consistent with classic.

Velite12:
Sometimes when fighting Blackheart the Inciter, one of the pets will continue to attack him and not the other players, which would indicate to me that the faction assignment is not consistent, but again, this is consistent with classic, sometimes the pets will continue to attack him.

In the case of the charming totem, the Hunter/Warlock will be charmed but their pet won't always be charmed since the other mobs in the pack will continue to fight with the pet, and the pet will fight them.

Velite12:
image

another suggestion is perhaps the IsPet() fails because the ObjectGuid version is being called?

Velite12:
image
The caster of the spell now that I  checked with the commit before this one is the Owner of the Pet (so in the case of the test I did, I was alone on hunter character and I had a pet out. I got hit by charming totem). It looks like Htest (me, the hunter) is trying to cast at a nullptr victim a certain spell, which I would assume that victim is supposed to be my pet, which causes the crash.

Velite12:
https://www.wowhead.com/tbc/spell=27022/volley It it's volley an aoe spell. So I think the game is trying to find the xyz in order to cast the spell, which is an aoe type spell, for the targeting?

Velite12:
At some point between AttackClosestEnemy in PlayerAI::JustGotCharmed(Unit* charmer) and Unit::CastSpell, the victim goes from a Unit which is actually a Pet (which I assume is my pet) to nullptr. The addresses between me and the victim were different, different GUID too.

Velite12:
image
Here is the guilty party. Select attacking target is returning nullptr. It appears the targets are chosen from threat list.

Then that null target is passed to DoCastSpellIfCan where the crash happens.

Velite12:
MeetsSelectAttackingRequirement check is returning false, which means suitableunits remains 0, even though the target used as input is not nullptr.

Velite12:
image

When the HunterAI CharmAI casts volley, the target (who is a pet) is not a player, but the selectflag being passed is for player. So `(selectFlags & SELECT_FLAG_PLAYER) && (!target->IsPlayer() || !target->IsPlayerControlled())` will be valid, which causes this function to return false, which causes no target to be added, which then causes the nullptr victim.

Velite12:
attacking random enemy may not always be a player so I think using SELECT_FLAG_IN_LOS instead of SELECT_FLAG_PLAYER to be passed in `PlayerAI::AddPlayerSpellAction` when the CharmAI adds spells to the list should fix this.

Velite12:
image
And Voila! Volley is now being cast by hunters in mind control! And on the pet which wasn't working before and causing the crash.

And no more nullptr victim anymore.

As a side effect it looks like this may have fixed much of the spell failed cast spam that happens during player charms. Not happening as frequently, the Hunter is using raptor strike too.

Velite12:
image
Can see as well that the mc hunter is using explosive trap on pet.

Velite12:
We can consider this solved now with this commit. The CharmedAI spamming abilities when they're on cooldown is a separate topic.

src/game/AI/BaseAI/PetAI.cpp
insunaa: `(m_pet && m_pet->getPetType() == HUNTER_PET)` does the same thing (`nullptr` always evaluates to boolean `false` so just evaluating the pointer directly is the same thing as checking for not nullptr

src/game/AI/PlayerAI/PlayerAI.cpp
```diff
@@ -48,7 +48,7 @@ void PlayerAI::AddPlayerSpellAction(uint32 spellId, std::function<Unit*()> selec
             if (IsNextMeleeSwingSpell(spellInfo))
                 selector = [&]()->Unit* { return m_player->GetVictim(); };
             else if (HasSpellTarget(spellInfo, TARGET_UNIT_ENEMY) || (spellInfo->Targets & TARGET_FLAG_DEST_LOCATION)) // always random
-                selector = [&, spellId = spellInfo->Id]()->Unit* { return m_player->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0, spellId, SELECT_FLAG_PLAYER); };
+                selector = [&, spellId = spellInfo->Id]()->Unit* { return m_player->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0, spellId, SELECT_FLAG_IN_LOS); };
```
insunaa: What is the purpose of this change?

Velite12: > image
> 
> When the HunterAI CharmAI casts volley, the target (who is a pet) is not a player, but the selectflag being passed is for player. So `(selectFlags & SELECT_FLAG_PLAYER) && (!target->IsPlayer() || !target->IsPlayerControlled())` will be valid, which causes this function to return false, which causes no target to be added, which then causes the nullptr victim.

Velite12: > attacking random enemy may not always be a player so I think using SELECT_FLAG_IN_LOS instead of SELECT_FLAG_PLAYER to be passed in `PlayerAI::AddPlayerSpellAction` when the CharmAI adds spells to the list should fix this.

insunaa: Yes I saw that, but what is the issue with a spell having a nullptr victim? This should just cause the spell to fail with bad targets

Velite12: So therefore we have to change this to account for the pet not being a player when we try to attack it randomly, which prevents the nullptr causing the server crash when we try to request the xyz of a nullptr.

insunaa: Well again: a spell with invalid targets should just fail with "bad targets".
Do we have a confirmation that mind controlled players actually do attack pets or NPCs?

Velite12: > Yes I saw that, but what is the issue with a spell having a nullptr victim? This should just cause the spell to fail with bad targets

If you look at the original post, the crash is happening because we are passing a victim then getting the xyz of the victim to cast a spell in a particular area (like volley, hurricane, etc), and we assume the victim is not nullptr. So when we try to get the x of a nullptr that is an access violation crash.

insunaa: > If you look at the original post, the crash is happening because we are passing a victim then getting the xyz of the victim to cast a spell in a particular area (like volley, hurricane, etc), and we assume the victim is not nullptr. So when we try to get the x of a nullptr that is an access violation crash.

Sure, but that's an issue with the spell logic, nothing to do with the target selection in the charmed player logic

Velite12: This is part of the logic for charmed ai players casting attacks which follows different logic. It is only very late in the chain of code that the victim becomes a nullptr, up before that point we have the pet as a valid victim.

Velite12: > > Yes I saw that, but what is the issue with a spell having a nullptr victim? This should just cause the spell to fail with bad targets
> 
> If you look at the original post, the crash is happening because we are passing a victim then getting the xyz of the victim to cast a spell in a particular area (like volley, hurricane, etc), and we assume the victim is not nullptr. So when we try to get the x of a nullptr that is an access violation crash.

To expand on this point, when a player is charmed, they get a specific list of spells added to a spell list, and then they cast these spells randomly at random targets. And these random casts assume that another player is a target. I think before we get the cast result we get the access violation from the nullptr because we cast this on a pet assuming it's a player. So the fix remedies that assumption.

insunaa: > To expand on this point, when a player is charmed, they get a specific list of spells added to a spell list, and then they cast these spells randomly at random targets. And these random casts assume the player is a target. I think before we get the cast result we get the access violation from the nullptr.

Sounds like a separate issue to me.

Velite12: So in reality, the spell wouldn't have a nullptr victim, it actually does have the pet as victim up until just before we are getting the location to cast the spell, then it gets set to nullptr.

Velite12: > > To expand on this point, when a player is charmed, they get a specific list of spells added to a spell list, and then they cast these spells randomly at random targets. And these random casts assume the player is a target. I think before we get the cast result we get the access violation from the nullptr.
> 
> Sounds like a separate issue to me.

Well as it turns out, this IS the issue that causes the server to crash. So this is the real problem, I just didn't realize the solution until today. This is the ultimate source of the crash and so we can now use the pet that is targeted as the victim for getting the xyz for the volley cast, which you see in the screenshots.

insunaa: They're separate issues. The nullptr check in Unit.cpp is fine and the nullptr check in PetAI.cpp is also fine. But the change from this comment thread in my opinion needs to be a separate pull request, because it does not cause a crash. It just selects targets in a way that you don't think is accurate to the source material.

But I haven't seen evidence that player pets are supposed to be targeted with area of effect spells when players are mind controlled. So that should be in a separate PR where it can be evaluated on its own merit and not tacked on to a crashfix PR

Velite12: Because what I did 2 weeks ago, all that did was just require the victim to be not nullptr before getting the xyz. While that stops the crash, the charmed ai players could still not cast their spells on the pets until I made this change. So that is the reason for this change. Now you don't see the error log spam when a player is charmed and someone in the party or raid has a pet out.

insunaa: image
Also I'd do it like this instead. No real need to let the spell get any further in processing or log info about it.

Velite12: > image Also I'd do it like this instead. No real need to let the spell get any further in processing or log info about it.

Let me get back to you on being able to target pets during charm first.

insunaa: @Velite12 Even if pets can be targeted by area of effect spells during charm, that should still be in a separate PR. Maybe it has merit, maybe it doesn't, it just shouldn't be in the same PR that fixes the nullptr accesses.

Velite12: https://www.youtube.com/watch?v=moK-ZVRNfeg

You can see that the pet gets faerie fire'd right near the beginning, this is in actual tbc.

So pets should be valid targets of attacks during Charms. The issue isn't isolated to the aoe casts, because if the pet target becomes nullptr then you wouldn't be able to cast faerie fire in our core.

Velite12: image

And you can tell that they're fighting eachother so definitely valid for directly casted things.

insunaa: @Velite12 Again the issue I'm having here isn't whether or not it's valid, the issue is that you're bundling it with a PR that is on a technical level unrelated.

Velite12: Other thing is that I assumed that @killerwife  didn't merge my previous commits because he was expecting an actual solution since he said that a nullptr shouldn't be passed to begin with, and I agree, that solves his question.

insunaa: More likely is that killerwife just didn't get around to look at it properly yet. He's pretty busy

Velite12: > @Velite12 Again the issue I'm having here isn't whether or not it's valid, the issue is that you're bundling it with a PR that is on a technical level unrelated.

Okay how about this. The victim being nullptr is downstream of the actual problem. So we'll leave this here since these nullptr checks are better safe than sorry. And I'll put the upstream solution as a separate pr.

insunaa: The victim being nullptr is unequivocally a core error that needs to be fixed. So it's good that this PR exists. The nullptr check for `m_pet` is also unequivocally an error that needs to be fixed. These 2 I am in full agreement with remaining in this PR.

The target selection from first glance at your evidence *also* seems to be a valid problem, and I am in favor of fixing that in a separate PR. Thank you for finding these things :+1:

Velite12: > More likely is that killerwife just didn't get around to look at it properly yet. He's pretty busy

On this point he did respond later after I posted the previous commits, so I assumed he took a look at them before commenting, especially when he inspected one of them.

insunaa: > On this point he did respond later after I posted the previous commits, so I assumed he took a look at them before commenting, especially when he inspected one of them.

Not wrong, but there's like 5 PRs I have open in various repos here that have existed for months if not years and still haven't even been looked at. In CMaNGOS-verse getting a PR merged is unfortunately just pretty slow :(

src/game/Entities/Unit.cpp
```diff
@@ -1565,8 +1565,13 @@ SpellCastResult Unit::CastSpell(Unit* Victim, SpellEntry const* spellInfo, uint3
     SpellCastTargets targets;
     targets.setUnitTarget(Victim);
 
-    if (spellInfo->Targets & TARGET_FLAG_DEST_LOCATION)
-        targets.setDestination(Victim->GetPositionX(), Victim->GetPositionY(), Victim->GetPositionZ());
+    if ((spellInfo->Targets & TARGET_FLAG_DEST_LOCATION))
+    {
+        // This shouldn't happen, but we should return gracefully if it does...
+        if (!Victim)
```
killerwife, line 1571: keep the log here as well. Otherwise ur silently ignoring smth that shouldnt happen. Also add a tracelog to the error.

killerwife (changes requested):

Velite12:
Done.

PR #820, opened 2026-05-27
fix nullptr in movementhandler

Velite12:
fix nullptr possibility when character lands from flight path.

### DIFF
--- a/src/game/MotionGenerators/MovementHandler.cpp
+++ b/src/game/MotionGenerators/MovementHandler.cpp
@@ -44,6 +44,9 @@ void WorldSession::HandleMoveWorldportAckOpcode(WorldPacket& /*recv_data*/)
 
 void WorldSession::HandleMoveWorldportAckOpcode()
 {
+    if (!GetPlayer())
+        return;
+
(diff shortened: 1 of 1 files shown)


src/game/MotionGenerators/MovementHandler.cpp
```diff
@@ -44,220 +44,224 @@ void WorldSession::HandleMoveWorldportAckOpcode(WorldPacket& /*recv_data*/)
...
-    // get start teleport coordinates (will used later in fail case)
-    WorldLocation old_loc;
-    GetPlayer()->GetPosition(old_loc);
-
-    // get the teleport destination
-    WorldLocation& loc = GetPlayer()->GetTeleportDest();
-
-    // possible errors in the coordinate validity check (only cheating case possible)
-    if (!MapManager::IsValidMapCoord(loc.mapid, loc.coord_x, loc.coord_y, loc.coord_z, loc.orientation))
+    if (GetPlayer())
```
killerwife: Do an early return. Otherwise you rip history.

PR #822, opened 2026-06-08
Change when Blackheart resets threat

Velite12:
He is supposed to reset threat on the following:

Charge random player far away
Incite chaos end

He is not supposed to drop threat during war stomp or start of incite chaos

See following videos:

TBC
https://www.youtube.com/watch?v=kTQefSWeFHY

TBC
https://www.youtube.com/watch?v=1XulHGYZ5ME

TBC Classic
https://www.youtube.com/watch?v=fPZ70Si7wZc&t=1186s


Velite12:
First and Third video is more clear that the threat is preserved on war stomp, or else the tank would lose control immediately to ranged dps.

PR #823, opened 2026-06-09
Update PetAI.cpp

Velite12:
Addresses https://github.com/cmangos/issues/issues/4088

nullptr target self casted pet spells, rather than the pet unit, since blood pact is being cast on self after phase shift, but phase shift prevents the unit from being targeted. So the autocast fails to cast, even though the manual cast works fine.

Later on in `Spell::CheckCast()`, if there is a target (even self!), it will not be self casted as `bool selfTargeting` will remain `false`. So by keeping the target nullptr, then it will resolve the target as the caster and set `selfTargeting = true;`. But seeing as this code is downstream from this particular issue with a pet autocast, I chose to change this specifically in the autocast code, and then later on the target will default to self.


insunaa:
LGTM

killerwife:
This actually has a simple as balls fix, and exactly why I took so long to merge the Pet thing in the first place. 

https://github.com/cmangos/tbc-db/blob/master/Updates/0702_pet_spell_lists.sql

Use this as template, add SQL for all imp spells. By default use target 0 (none) for imp blood pact, and unless I am mistaken, it will work with no changes.

killerwife:
It has no wiki page yet, cos didnt get to it, but columns behave same way as creature_spell_list and are implemented using same functions in core

Velite12:
> This actually has a simple as balls fix, and exactly why I took so long to merge the Pet thing in the first place.
> 
> https://github.com/cmangos/tbc-db/blob/master/Updates/0702_pet_spell_lists.sql
> 
> Use this as template, add SQL for all imp spells. By default use target 0 (none) for imp blood pact, and unless I am mistaken, it will work with no changes.

so I did:

`DELETE FROM pet_autocast_spell_list WHERE CreatureEntry=416;
INSERT INTO pet_autocast_spell_list(CreatureEntry, SpellId, CombatCondition, TargetId, Comments) VALUES
(416, 6307, -1, 0, 'Imp - Blood Pact Rank 1'),
(416, 7804, -1, 0, 'Imp - Blood Pact Rank 2'),
(416, 7805, -1, 0, 'Imp - Blood Pact Rank 3'),
(416, 11766, -1, 0, 'Imp - Blood Pact Rank 4'),
(416, 11767, -1, 0, 'Imp - Blood Pact Rank 5'),
(416, 27268, -1, 0, 'Imp - Blood Pact Rank 6');`

And when I went into the game, yes, the imp did autocast blood pact and phase shift at same time... But then it proceeded to spam blood pact nonstop.

The noise was pretty loud. Is there a way to make them cast it only if they don't have the aura? Blood Pact has no cooldown.

killerwife:
I will fix that usecase. Gotta iron out the kinks on new tech.

killerwife:
Linked a commit that should solve that issue.

Velite12:
image

Read access violation when I summon the imp with latest commits.

Velite12:
This is before the autocast bar is even loaded, so the imp is out (I logged in) but the autocast bar is not visible.

Velite12:
If it helps it seems the values for the effect indices are the spell ids?

Velite12:
Target is nullptr, we need to check if target before we access HasAura.

Velite12:
I fixed the nullptr check so the server doesn't crash but the sound and cast spam is still happening 😞

Velite12:
image

Blood pact is considered an AREA_AURA_PARTY, but it's failing this check here. 

So the HasAura check which was crashing the server was actually for the one previous, but that statement is ultimately not true so the inside code doesn't happen. I think it is failing here because the pet is targeting itself.

GetOwnerGuid returned = {m_guid=14235 } (caster->GetMasterGuid() call)
GetObjectGuid returned = {m_guid=14235 } (target->GetObjectGuid() call)

The guid's match so the code inside doesn't reach. Maybe I have to change the targetId for the entry in the db. I've been using 0 for the targetId.

Velite12:
> This actually has a simple as balls fix, and exactly why I took so long to merge the Pet thing in the first place.
> 
> https://github.com/cmangos/tbc-db/blob/master/Updates/0702_pet_spell_lists.sql
> 
> Use this as template, add SQL for all imp spells. By default use target 0 (none) for imp blood pact, and unless I am mistaken, it will work with no changes.

It turns out the simple fix is not so simple, and it needs changes 😎 

For the record my initial commit fixed the original bug and didn't cause these issues, as I have been testing your proposal in a branch that does not have the commit I made. I think we should go with my commit until you fully implement the PetAI solution because right now it's broken.

killerwife:
Absolutely, if the status quo was to keep the old one working, I would agree. But I am attempting here to DELETE the old. And generally in open source the way things work after a time is if you dont rip the bandaid off and merge it to master, people are generally happy with the current status quo. (because they dont know there is an issue to begin with)

Working on a fix, because it is not as simple as it might be on first glance.

killerwife:
How is the situation of this after https://github.com/cmangos/mangos-tbc/commit/2ecf735a51818e76ae050461a5344244368c1c46 ?

Velite12:
Haven't got the chance to test it yet but I will.

Velite12:
> How is the situation of this after [2ecf735](https://github.com/cmangos/mangos-tbc/commit/2ecf735a51818e76ae050461a5344244368c1c46) ?

Tested again, he's still spamming blood pact.

```sql
DELETE FROM pet_autocast_spell_list WHERE CreatureEntry=416;
INSERT INTO pet_autocast_spell_list(CreatureEntry, SpellId, CombatCondition, TargetId, Comments) VALUES
(416, 6307, -1, 0, 'Imp - Blood Pact Rank 1'),
(416, 7804, -1, 0, 'Imp - Blood Pact Rank 2'),
(416, 7805, -1, 0, 'Imp - Blood Pact Rank 3'),
(416, 11766, -1, 0, 'Imp - Blood Pact Rank 4'),
(416, 11767, -1, 0, 'Imp - Blood Pact Rank 5'),
(416, 27268, -1, 0, 'Imp - Blood Pact Rank 6');

killerwife:
During merge i made sure this is now working. Closing, will add the SQL to DB in your name

PR #826, opened 2026-06-13
Include pets for SelectAttackingTarget ATTACKING_TARGET_RANDOM

Velite12:
As a continuation of https://github.com/cmangos/mangos-tbc/pull/819, here we fix the bug that Pets are being set to nullptr when they are to be attacked by a charmed player at random.

Here is some evidence that pets should be targets for attacks while players are charmed:

Actual TBC:
https://www.youtube.com/watch?v=moK-ZVRNfeg

Classic:
https://youtu.be/RCkHCuQu1VI?t=1983

You can see in both examples that Pet should be considered targets for random attacks. You can see the druids casting mangle/faerie fire on the hunter pet in the first video. In the second video you can see the warrior using cleave and autos on the balance druid summoned treants. So we can't assume that only players should be targeted for random attacking.

Because the select flag is for player, it then causes a function later to return a nullptr for the victim even though the victim was already the pet, which is passed later on and causes the failure to cast/server crash.


killerwife:
I added a safety against this. In an ideal world, this code here will be all deleted and use basically same system as im prototyping in PetAI (and i have been thinking about it for a long time), just for players with autoscaling ranks. However the flag change is legit still even after my safety addition.

PR #827, opened 2026-06-14
Disable SetFacingTo to prevent animation interruption during channel

Velite12:
Blackheart the incter's laugh_periodic during the charm phase was considered NYI. During the charm phase he will stare at people but he won't do his laugh channel like he's supposed to.

It turns out the code as is already implements the laugh, but you don't see his laugh animation because SetFacingTo during the channel is interrupting the Animation. If you cast this spell as a player, this function isn't called and the spell works (try .cast 33722), since players are client controlled so this code never reaches. Removing this function call fixes the implementation.


killerwife:
Remove the commenting out and delete it outright

PR #828, opened 2026-06-15
Fix IsImmuneToSpellEffect

Velite12:
Fix IsImmuneToSpellEffect check which was previously failing for stun type spells by using a more robust getter instead of accessing a field.

### DIFF
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -2066,7 +2066,7 @@ bool Creature::IsImmuneToDamage(SpellSchoolMask meleeSchoolMask)
 
 bool Creature::IsImmuneToSpellEffect(SpellEntry const* spellInfo, SpellEffectIndex index, bool castOnSelf) const
 {
-    if (!castOnSelf && GetCreatureInfo()->MechanicImmuneMask & convertEnumToFlag(spellInfo->EffectMechanic[index]))
+    if (!castOnSelf && GetCreatureInfo()->MechanicImmuneMask & convertEnumToFlag(GetEffectMechanic(spellInfo, index)))
         return true;
 
     return Unit::IsImmuneToSpellEffect(spellInfo, index, castOnSelf);


killerwife:
Please detail the usecase of what this fixes, spell ids etc

Velite12:
Casting spell ID 10308 https://www.wowhead.com/wotlk/spell=10308/hammer-of-justice. If target had immunity mask for stun, it would say that the target is not immune to stun. Hammer of Justice effect 0 is a stun, the field was not returning correctly for the bitwise &. The getter method is a more robust search for the effect mechanic.

Velite12:
That being said, IsImmuneToSpell works as is to achieve the same effect, but I noticed that the core uses a mix of both of these, sometimes both at the same time in the case of Kalecgos.

killerwife:
The way this is designed to work, is that a IsImmuneToSpellEffect is meant to be called after IsImmuneToSpell logically, cos first immunity for a spell globally is meant to be checked, and then only locally. So where in the core did global immunity check let it go through but fail on the per effect check.

Anyway, explaining my reasoning here, I will obviously check before merging this.

Velite12:
Not the core but it was being called in the playerbot code and when I traced the breakpoints I found it was returning false for immunity.

killerwife:
Ok but we dont alter core due to wrong playerbot code(which has not been as closely curated as the core is). I understand that some of these functions are a bit obtuse, but these immunity func generally try to closely resemble client. Some of them can be found in wotlk client (not tbc, tbc has had fewer checks clientside).

So unless there is an issue in the core without any addons, then this a non-issue.

Velite12:
Well if it doesn't work in the core then we have a problem. I suppose I will come back doing a core test. But I don't see why calling the same function with the same parameters should produce a different result.

On a technical level the change I made does the same thing as what it has previously assuming the field exists

image

It would still resolve to a value from the enum. But if the effectmechanic fails then it will get the mechanic field. This is also used here in the core:

image

Just for the record this isn't going to be used in playerbots anymore since this function doesn't work, so I'm changing the usage to IsImmuneToSpell, but I thought that from my findings it would be worth a look.

killerwife:
What I meant is that blizzlike code, first evaluates Spell level, where IsImmuneToSpell applies. IsImmuneToSpellEffect is the next level granular check later down the gutter in the innards.

One produces failed spell, second one just immunes out effect. I am guessing playerbot code wanted to do former, not latter. (or maybe even both, and in that case, you are meant to call BOTH functions)

Velite12:
Oh I see, all 3 effect mechanics are 0. That's why this check failed. It was only looking at effectmechanic.
image

But mechanic field is 12 which is stun.

So basically you have to check both. If either one pops up immune then there's an immunity.

PR #829, opened 2026-06-17
Nazan should use Top Aggro, not victim for ground fireballs

Velite12:
Evidence has come to light that supports the idea that Nazan's Fireballs should consider aggro, not necessarily victim.

https://www.youtube.com/watch?v=xbbHr28kvTo

At 2:29:
You can see in this video, only the tank and the hunter taking the video ever get the fireballs. If you look, the first fireball the hunter gets, the boss doesn't move and switches back to the tank. Then the hunter gets the next fireball, then after he feigns death, the tank gets the rest. None of the other party members are hit by a fireball, the only HP changes for them are commanding shout fading and lifetaps. In any case, the tank took the majority of the fireballs in this fight, the hunter took some, which would support the idea that the boss will target fireballs at whoever has the highest threat value, which can be ranged dps or healers, even if they aren't high enough to overtake on threat.

Threat overtaking is harder for ranged than it is for melee, so it's possible to have more threat but not be the current "tank".

image
https://warcraft.wiki.gg/wiki/Threat


Velite12:
https://www.youtube.com/watch?v=jNncb45K3Ck

This video is another good example. Once Nazan is on the ground, It's only the tank and later on the hunter who takes them back to back.

PR #833, opened 2026-06-26
[SD][Spell] Readd Target Change Logic for Raging Flames 20481

AnonXS:
Removed with https://github.com/cmangos/mangos-tbc/commit/329ad4f81d8576625d8dec6d710eebe66b5bef96
Originally added with https://github.com/cmangos/mangos-tbc/commit/ea618a7c4eec3f8d9dee94b1b62966f21f91d3b8

Can also Focus Tank on Aggro, should be even more unpredictable when they change their focus, needs to be figured out or timed. (they only have 35281 aura passive)

See: https://www.youtube.com/watch?v=KuRH5PzZ3DA first 10 seconds how the add turns around without inferno used

https://www.youtube.com/watch?v=f3P3-odMKeA - watch with sound on

Also seem to cast invis and stealth detect within the first few seconds of a fight - might be a classictbc thing, confirmed with 7.3.5 sniff

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/the_mechanar/boss_nethermancer_sepethrea.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/the_mechanar/boss_nethermancer_sepethrea.cpp
@@ -102,6 +102,7 @@ struct boss_nethermancer_sepethreaAI : public CombatAI
 
 enum
 {
+    SPELL_INVIS_AND_STEALTH_DETECT = 18950,
     SPELL_RAGING_FLAMES = 35281,
     SPELL_INFERNO       = 35268,
     SPELL_INFERNO_H     = 39346,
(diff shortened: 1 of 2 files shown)


src/game/AI/ScriptDevAI/scripts/outland/tempest_keep/the_mechanar/boss_nethermancer_sepethrea.cpp
```diff
@@ -137,11 +139,23 @@ struct npc_raging_flamesAI : public CombatAI
...
 
+// 35268,39346 - Inferno
+struct RagingFlamesInferno : public AuraScript
+{
+    void OnApply(Aura* aura, bool apply) const override
+    {
+        if (!apply)
+            if (Unit* caster = aura->GetCaster())
+                if (Unit* newTarget = caster->SelectAttackingTarget(ATTACKING_TARGET_RANDOM, 0, nullptr, SELECT_FLAG_PLAYER))
+                    caster->AddThreat(newTarget, 10000000.0f);
```
killerwife: 1000000 in tbc
10000000.0 - is legion value

killerwife:
CAn confirm it uses the stealth detect even in tbc prime

miraco:
You can drop aggro with jumping on TBC classic/anni servers maybe this is the case here too? 

For example Leotheras in SSC
The way his Whirlwind works is line this:

When he casts whirlwind he locks his current threatlist and just goes from 1 to 2nd to 3rd and so on.
Now the tactic to parse on this boss was to keep him in melee range all the time and overheal melees 

The way we achieved this on tbc classic and Anni servers is to drop threat before whirlwind (feign death for huters or soul shatter for locks as example) or to jump if he is in whirlwind and targets you (there is a weakaura for this) this way he will ignore you and go to next target 

And iirc this was the same on original 2.4.3 (I think there was a blue post from blizz about this)

PR #834, opened 2026-06-28
Fixes for Cipher of Damnation final quest 10588.

Velite12:
Fixes for Cipher of Damnation final quest 10588. This is a companion PR with DB changes: https://github.com/cmangos/tbc-db/pull/1293

Both PRs need to be done together.

Video of quest:
https://www.youtube.com/watch?v=e5kLAoug2Ms
https://youtu.be/4wwSlUF7QV8
https://youtu.be/4wwSlUF7QV8

Summary:

Herein Oronok Torn-Heart is referred to as the leader npc.

Rework the intro with the npc fly in by using spawn group with formation instead of summons with follows, which ensures that the 2 npcs following the leader do not get stuck when the leader flies down and dismounts. They were lagging behind and got stuck on a rock previously which caused incorrect pathing for them.

Address leftover npcs in case the leader dies, since their despawn code was assuming he reaches certain waypoints/points, which he can't if he's dead. Also address the npc corpses being leftover if the summoned boss were to evade if everyone was dead, since again the leader isn't alive to do rp to force despawn them.

Address inaccuracies with the fight positioning. You can see in all the videos that the npcs move further up the hill near that earth elemental mob (they may even sometimes fight with it), and the boss moves out of the lava to engage with them. These locations xyz were not available in our core/db so I added them. While these are not 100% accurate, they are necessary since they fix a separate issue which is caused by the current points:

- When the boss presently engages with the npcs at the current fight point (elementals move loc), they will attack at the edge of the lava and this can cause pathing issues at the end of the fight if they are not moved away (one or both the followers may be stuck, leader may not do rp). Seeing as how they're supposed to be fighting much further inland as in the video, this fixes the problem. Elementals Move Loc isn't the final point where the fight starts anyway.

Move the leader npc's abilities out of the core and to the db. I also added targeting in the db for the healing wave spell, as you can see in some of the videos he does not only cast it on himself.

Allow Grom'tor's sunder spell to stack with warrior sunders as you can see in the videos that there are multiple sunders active (the npc's and the warrior player's). Right now when he casts the spell it starts to fill up the server log in mangosd.exe with messages that the spell was bounced as it probably does not expect this spell to be bouncing.

Change the point at which the mark of kael'thas is visible, roughly speaking in the videos it is at some point when the Fire elemental starts speaking. Currently the mark is not showing at all as either the event id is invalid or it is never called, so you never see the mark of kael'thas go.

Fix the positioning for the 2 follower npcs during the end rp, as they are supposed to turn with the leader in formation. SetFacing/SetFacingToObject does not do this automatically. I found the easiest way was to use Compact(false), as this calls a function which would be perfect to call when an npc formation leader turns in place, but this function is private, so I use compact as a wrapper.

Fix the leader npc not giving his weapon at the end of the quest. It turns out that the text say emote he does places his weapon on his back, which means that when you set the virtual item, it just causes him to be empty handed but you would still see the staff on his back. So I delayed when their weapons disappear to after he finishes this emote, that way the weapon can be back on his hands in time to make it disappear. Although, I think it should be possible to remove the item when it is on his back, but this is out of scope of these changes.

I added a reference id for the spawngroup in the db, it's unused but it's helpful when looking for the db changes.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/shadowmoon_valley.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/shadowmoon_valley.cpp
@@ -1940,6 +1940,7 @@ enum
     NPC_ORONOK_TORN_HEART           = 21685,
     NPC_GROMTOR_SON_OF_ORONOK       = 21687,
     NPC_BORAK_SON_OF_ORONOK         = 21686,
+    SPAWN_GROUP_ORONOK              = 26004,
     NPC_CYRUKH_THE_FIRELORD         = 21181,
     // NPC_EARTH_SPIRIT              = 21050,
     NPC_REDEEMED_SPIRIT_OF_EARTH    = 21739,
@@ -1954,7 +1955,10 @@ enum
 
     POINT_ID_ATTACK_READY           = 1,
     POINT_ID_ELEMENTS               = 2,
-    POINT_ID_EPILOGUE               = 3,
+    POINT_ID_START_FIGHT            = 3,
+    POINT_ID_CYRUKH_APPROACH        = 4,
+    POINT_ID_EPILOGUE               = 5,
+    WAYPOINT_ID_START               = 1,
     WAYPOINT_ID_DESPAWN             = 5,
     WAYPOINT_ID_INTRO_FINAL         = 31,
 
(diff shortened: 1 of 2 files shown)


src/game/AI/ScriptDevAI/scripts/outland/shadowmoon_valley.cpp
```diff
@@ -2170,18 +2171,77 @@ struct npc_spawned_oronok_tornheartAI : public ScriptedAI, private DialogueHelpe
...
+                case POINT_MOTION_TYPE:
+                    if (pointId == POINT_ID_CYRUKH_APPROACH)
+                    {
+                        // Cyrukh starts to attack
+                        m_creature->setFaction(FACTION_ORONOK_COMBAT);
+                        if (Creature* borak = GetSpeakerByEntry(NPC_BORAK_SON_OF_ORONOK))
+                            borak->setFaction(FACTION_ORONOK_COMBAT);
+                        if (Creature* gromtor = GetSpeakerByEntry(NPC_GROMTOR_SON_OF_ORONOK))
+                            gromtor->setFaction(FACTION_ORONOK_COMBAT);
+                        if (Creature* pCyrukh = GetSpeakerByEntry(NPC_CYRUKH_THE_FIRELORD))
```
killerwife: hungarian notation

killerwife:
Just gonna point out here that its pretty visible that this script predates all the tech. I wrote this back in the day when none of the systems you used existed. Still cool it held out this long.

PR #837, opened 2026-07-08
Drop earth shield charges on proc

Velite12:
Fix earth shield not dropping charges on proc.


killerwife:
Implemented in a different way, reason being DropAuraCharge is part of the common code for procs and if not done in that one specific location, you can drop charge and rip rest of effects. Has to be done after whole proc holder execution due to said side effects. My script was however wrongly safeguarded.

Velite12:
Because the dummy aura is the first effect (effect index 0) but there is another spell effect for the other aura (the interruption resistance), it was messing up because we didn't handle the result of that effect index 1 properly? Am I understanding this right, for posterity sake.

killerwife:
Yes, script was written for effect 0 but was executing twice, causing the drop stack issue.

PR #839, opened 2026-07-09
Fix creature assistance on pull not working as intended

Velite12:
address https://github.com/cmangos/issues/issues/4099

Creature assistance on pull was not working as intended: On pull, the call for help radius should determine what mobs aggro alongside the mob that you just pulled, assuming that they are capable of aggroing eachother (linking outside of a group).

Here's a comparison of what it looks like in classic tbc anniversary:

https://www.youtube.com/watch?v=KzAp7Gw623s

You can see in this instance that when the mob is pulled, the nearby mobs within it's callforhelp distance (which we will assume here is 10 as the 5 yards default doesn't reach) will pull at the exact same time.

If you try to replicate this pull on cmangos in those exact positions, neither of those mobs will pull from that angle, because it won't send out the request for assistance until about 3 seconds have passed.

As it is working currently, two delays are being applied to creatures on pull: first the delay for checking for help, and then another delay to actually do the linking. As a result, once a mob is pulled, it won't aggro a nearby mob, even if it should because it is within the callforhelp radius, because the aggro event won't happen for a couple of seconds, at which point the creature would be out of range of those mobs.

Also, the radius used to check for help is different from the one used to call for help. Callforhelp is something that's set in creature_template (0 would use default 5 yd in server config), while checkforhelp was using a radius that comes from the config and is by default 5. I made the checkforhelp radius use the callforhelp radius if the value is > 0 (which would mean it is not the default 5 yards), such that the call and check radius match each other for creatures who do not use the standard callforhelp value.

For the purposes of testing I found that the callforhelp of this particular creature (tempest-forge patroller) should be 10 (it currently has 0 which defaults to 5 yards which is too short to reach the 3 pack on the side when you do this pull, but it does reach the patroller behind it which is impossible in current code). 

Here's what the pulls look like with the changes I made

https://www.youtube.com/watch?v=SqSuqze5xT0

The behavior in this video is not applicable in current code when setting callforhelp for creature 19166 to 10.

As soon as I pull the mob, it sends out the event for assistance so the other mobs within its callforhelp will pull. Currently we didn't have a delay assigned to CallAssistanceOnPull reaction event (delay for when the mobs actually decide to assist after receiving the event to do so) now they wait a bit (based on server config) and then assist. We had the delay set to sending out the event, which is wrong since it will cause those mobs nearby in the original position to not assist as the pulled mob would be too far away from them by that time.

TL;DR: Creatures who are not in the same group are not linking together accurately when one of them is pulled and all are stationary as the event is sent out too late, once the pulled mob has moved away, for the others to be in range to receive it. Also, the radius to check for help was not accounting for different call for help radius in the creature_template. By fixing both of these, creature linking outside of groups works as close to blizzlike as possible.

### DIFF
--- a/src/game/AI/BaseAI/UnitAI.cpp
+++ b/src/game/AI/BaseAI/UnitAI.cpp
@@ -81,7 +81,14 @@ void UnitAI::MoveInLineOfSight(Unit* who)
         return;
 
     if (who->GetObjectGuid().IsCreature() && who->IsInCombat())
-        CheckForHelp(who, m_unit, sWorld.getConfig(CONFIG_FLOAT_CREATURE_CHECK_FOR_HELP_RADIUS));
+    {
+        float radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_CHECK_FOR_HELP_RADIUS);
+        if (((Creature*)who)->GetCreatureInfo()->CallForHelp > 0)
+            radius = ((Creature*)who)->GetCreatureInfo()->CallForHelp;
+
+        CheckForHelp(who, m_unit, radius);
+    }
+        
 
     if (!HasReactState(REACT_AGGRESSIVE)) // mobs who are aggressive can still assist
         return;
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -2234,26 +2234,18 @@ void Creature::CallAssistance(Unit* enemy)
     }
 }
 
-std::pair<bool, CreatureList> Creature::MarkCallAssistanceOnPull(Unit* enemy)
+bool Creature::MarkCallAssistanceOnPull()
 {
     bool stored = m_AlreadyCallAssistance;
     SetNoCallAssistance(true);
 
     if (!CanCallForAssistance())
-        return {false, CreatureList()};
+        return false;
 
-    float radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS);
-    if (GetCreatureInfo()->CallForHelp > 0)
-        radius = GetCreatureInfo()->CallForHelp;
-
-    CreatureList receiverList;
-    MaNGOS::AnyAssistCreatureInRangeCheck u_check(this, enemy, radius);
-    MaNGOS::CreatureListSearcher<MaNGOS::AnyAssistCreatureInRangeCheck> searcher(receiverList, u_check);
-    Cell::VisitAllObjects(this, searcher, radius);
-    return {stored, receiverList};
+    return stored;
 }
 
-void Creature::CallAssistanceOnPull(Unit* enemy, CreatureList const& receiverList)
+void Creature::CallAssistanceOnPull(Unit* enemy)
 {
     if (enemy && !HasCharmer())
     {
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -2254,7 +2254,7 @@ void Creature::CallAssistanceOnPull(Unit* enemy)
         float radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS);
         if (GetCreatureInfo()->CallForHelp > 0)
             radius = GetCreatureInfo()->CallForHelp;
-        AI()->SendAIEventAround(AI_EVENT_CALL_ASSISTANCE, enemy, 0, radius);
+        AI()->SendAIEventAround(AI_EVENT_CALL_ASSISTANCE, enemy, sWorld.getConfig(CONFIG_UINT32_CREATURE_FAMILY_ASSISTANCE_DELAY), radius);
     }
 }
 
@@ -2264,7 +2264,14 @@ void Creature::CallForHelp(float radius)
         return;
 
     if (radius <= 0.0f)
-        radius = GetCreatureInfo()->CallForHelp;
+    {
+        // call for help can be 0
+        if (GetCreatureInfo()->CallForHelp > 0)
+            radius = GetCreatureInfo()->CallForHelp;
+        else
+            radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS); 
+    }   
+
 
     MaNGOS::CallOfHelpCreatureInRangeDo u_do(this, GetVictim(), radius);
     MaNGOS::CreatureWorker<MaNGOS::CallOfHelpCreatureInRangeDo> worker(this, u_do);
--- a/src/game/Entities/Creature.h
+++ b/src/game/Entities/Creature.h
@@ -788,8 +788,8 @@ class Creature : public Unit
         void CallForHelp(float radius);
         void CallAssistance();
         void CallAssistance(Unit* enemy);
-        std::pair<bool, CreatureList> MarkCallAssistanceOnPull(Unit* enemy);
-        void CallAssistanceOnPull(Unit* enemy, CreatureList const& receiverList);
+        bool MarkCallAssistanceOnPull();
+        void CallAssistanceOnPull(Unit* enemy);
         void SetNoCallAssistance(bool val) { m_AlreadyCallAssistance = val; }
         bool CanAssistTo(const Unit* u, const Unit* enemy, bool checkfaction = true) const;
         bool CanInitiateAttack() const;
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -580,7 +580,7 @@ void Unit::TriggerAggroLinkingEvent(Unit* enemy)
     if (callAssistance)
         static_cast<Creature*>(this)->CallAssistanceOnPull(enemy);
 
-    m_events.AddEvent(new UnitLambdaEvent(*this, [enemyGuid = enemy->GetObjectGuid(), creatureGroup = static_cast<Creature*>(this)->GetCreatureGroup(), callAssistance, receiverList](Unit& unit)
+    m_events.AddEvent(new UnitLambdaEvent(*this, [enemyGuid = enemy->GetObjectGuid(), creatureGroup = static_cast<Creature*>(this)->GetCreatureGroup(), callAssistance](Unit& unit)
     {
         Unit* enemy = unit.GetMap()->GetUnit(enemyGuid);
         if (!enemy || !unit.IsInCombat())
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -577,6 +577,8 @@ void Unit::TriggerAggroLinkingEvent(Unit* enemy)
         return;
 
     bool callAssistance = static_cast<Creature*>(this)->MarkCallAssistanceOnPull();
+    if (callAssistance)
+        static_cast<Creature*>(this)->CallAssistanceOnPull(enemy);
 
     m_events.AddEvent(new UnitLambdaEvent(*this, [enemyGuid = enemy->GetObjectGuid(), creatureGroup = static_cast<Creature*>(this)->GetCreatureGroup(), callAssistance](Unit& unit)
     {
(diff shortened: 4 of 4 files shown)


killerwife:
"As it is working currently, two delays are being applied to creatures on pull: first the delay for checking for help, and then another delay to actually do the linking."

I changed it a week ago to be one timer

Next:

```
if (radius <= 0.0f)
    {
        // call for help can be 0
        if (GetCreatureInfo()->CallForHelp > 0)
            radius = GetCreatureInfo()->CallForHelp;
        else
            radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS); 
    }   
```

This is wrong, beeeecause, default is when 0 is that call for help is entirely turned off. There are npcs like that, although it is rare.

Lastly the AI CheckForHelp in MoveInLineOfSight is not meant to engage during pull, it is disabled during the initial pull for that reason using the SetCanCheckForHelp - false

Next:

`As a result, once a mob is pulled, it won't aggro a nearby mob, even if it should because it is within the callforhelp radius, because the aggro event won't happen for a couple of seconds, at which point the creature would be out of range of those mobs.`

The mobs are cached on pull, i.e. the list, i.e. i broke this last week when making it one timer. This is a real issue I created that I did not realize at that time. What needs to be done in TriggerAggroLinkingEvent is the list is made and then CallAssistanceOnPull uses that list. It used to be done implicitly in SendAIEventAround when waittime is not 0, but that meant we had those two timers under the hood.

Velite12:
> ```
> if (radius <= 0.0f)
>     {
>         // call for help can be 0
>         if (GetCreatureInfo()->CallForHelp > 0)
>             radius = GetCreatureInfo()->CallForHelp;
>         else
>             radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS); 
>     }   
> ```
> 
> This is wrong, beeeecause, default is when 0 is that call for help is entirely turned off. There are npcs like that, although it is rare.
> 

image

So is the issues wiki wrong? It says that when call for help is set to 0, the default will be 5. In the core you can see a default is set to 5 as well.

setConfigPos(CONFIG_FLOAT_CREATURE_CHECK_FOR_HELP_RADIUS,         "CreatureCheckForHelpRadius",     5.0f);

Velite12:
> "As it is working currently, two delays are being applied to creatures on pull: first the delay for checking for help, and then another delay to actually do the linking."
> 
> I changed it a week ago to be one timer
> 
2 weeks ago and yes I can see the changes made. But this linking issue has been a thing (I've been doing mechanar before 2 weeks ago) at least in this dungeon the pulls were less strict than classic. You're able to pull side packs without getting the nearby mobs.

killerwife:
This is more of a "i realized this recently by playing a lot of classic", so technically yes wiki is wrong, but i havent known this for a long time.

Velite12:
> This is more of a "i realized this recently by playing a lot of classic", so technically yes wiki is wrong, but i havent known this for a long time.

Okay, but we still make this assumption in the code too. We do

```c++
void Creature::CallAssistanceOnPull(Unit* enemy)
{
    if (enemy && !HasCharmer())
    {
        MANGOS_ASSERT(AI());

        float radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS);
        if (GetCreatureInfo()->CallForHelp > 0)
            radius = GetCreatureInfo()->CallForHelp;
        AI()->SendAIEventAround(AI_EVENT_CALL_ASSISTANCE, enemy, 0, radius);
    }
}
```

So if 0 then the default of 5 would get used.

killerwife:
I pushed a commit, which in tandem with:

UPDATE creature_template SET CallForHelp=10 WHERE entry=19166;

sort of works. The patrols dont have correct points at a glance (i did mechanar two days ago on tbc classic) so there is now a delta in terms of the points / spawns, but it works like you wanted it to.

In regards to the code, yes, that is still there, I am always slow to change things for call to help because we dont actually have any official information on how it works. So I am always hesitant to change defaults. Especially since its a per creature setting and we dont have any leaks in regards to it. Not saying its ideal, just saying you stumbled upon a situation that isnt fully figured out yet.

Velite12:
I'll test it tomorrow and see.

miraco:
I partly reworked mechanar already, and the waypoints of those npcs should all be based on sniffs

only the formation npcs (middle is leader) could be off but thats part of formation movement code then.

killerwife:
@miraco you need to pause formation movement at the end and make them move to custom points. Afaik they never move after the first time?

miraco:
Yeah the pause formation is there
Prob have to delete formation on last waypoint and then let the outer npcs run to their final points

currently its only the middle npc having waypoint, and movement change to 0 on last waypoint

killerwife:
Yep, but considering i did the dung on classic tbc, thats not how the final point looks even on their end. (none of that customary turning etc)

miraco:
Yes they spread much more then we currently have (we have FormationSpread 2 atm its prob 3) resulting in other positions.

But i am not sure we need to add formation delete and custom waypoints here, i think changing spread to 3 is already enough, also on blizz server it doesnt look like they get own movement, just adjusting to leaders position

https://youtu.be/Ea32pA5P1eA

Core change looks working fine tho

Velite12:
Tested latest, effectively same result with the pulls as the changes I did, so I'll close.

Although the way I did it doesn't involve the caching so this is less memory efficient.

Velite12:
Reopening because the latest commit made is causing server crash after about 30 seconds. I ran and crashed the server around 5 times today, consistently within a minute of just idling, then I just got around to running it in relwithdebug to see what was happening.

image
image

Tempspawned creatures at Stair of Destiny are causing access violation because they are not in game anymore but are being referenced in the receiverlist  of the new CallForAssistanceOnPull

With the commits in this PR I wasn't encountering this problem. I played through the entirety of Heroic Mechanar and then some, total 1.5 hours and no crash. I think it is easier if we don't cache but that's me. 

We'd have to prune the bad references in the list when creatures are removed from the game. Which would mean that the receiverlist has to be accessed outside of this scope to be modified. We can't really check the object anymore since it would point to bad/garbage memory, so it's not like we can just nullptr guard.

Or, we should only access them when we're sure that they still exist ingame... which is what the PR does. Then it doesn't matter if they are confirmed to pull but delay and then despawn since there is only 1 moment where we need to reference them, and they have to exist at that moment to even go in the receiverlist.

I also checked and the PR calls the function SendAIEventAround which already uses a ReceiverList and there's no issues with the references in that case. I don't see the need for maintaining a separate ReceiverList and reinventing SendAIEventAround.

Unless this is the way (delayed event send to cached references) it's supposed to work in Classic, but it seems like a more complicated way to do what I did which effectively has the same result in gameplay without the access violation potential.

killerwife:
Yes there is a benefit, optimization and having it in one place. Combat is such a frequent operation that it is actually worth it. Anyway, fixed crash.

Velite12:
Tested, no crash after 10 mins, and pulls still work as intended so closing.

Velite12:
Btw, why don't we use GUID lists instead of creature lists in general if GUID lists are bad-memory-proof? We use creaturelists in general for events, esp with delay.

```c++
class AiDelayEventAround : public BasicEvent
{
    public:
        AiDelayEventAround(AIEventType eventType, ObjectGuid invokerGuid, Unit& owner, CreatureList const& receivers, uint32 miscValue) :
            BasicEvent(),
            m_eventType(eventType),
            m_invokerGuid(invokerGuid),
            m_owner(owner),
            m_miscValue(miscValue)
        {
            // Pushing guids because in delay can happen some creature gets despawned => invalid pointer
            m_receiverGuids.reserve(receivers.size());
            for (auto receiver : receivers)
                m_receiverGuids.push_back(receiver->GetObjectGuid());
        }

        bool Execute(uint64 /*e_time*/, uint32 /*p_time*/) override
        {
            Unit* pInvoker = m_owner.GetMap()->GetUnit(m_invokerGuid);
            if (!pInvoker)
                return true;

            for (ObjectGuid guid : m_receiverGuids)
            {
                if (Creature* pReceiver = m_owner.GetMap()->GetAnyTypeCreature(guid))
                {
                    pReceiver->AI()->ReceiveAIEvent(m_eventType, &m_owner, pInvoker, m_miscValue);
                    // Special case for type 0 (call-assistance)
                    if (m_eventType == AI_EVENT_CALL_ASSISTANCE)
                        pReceiver->AI()->HandleAssistanceCall(&m_owner, pInvoker);
                }
            }
            m_receiverGuids.clear();

            return true;
        }
 ```
        
actually it turns out that the aidelayeventaround already accounts for this? So the creaturelist just get turned into guid lists when searching through it. I missed this comment before and this is what was being used here before:

```c++
            // Pushing guids because in delay can happen some creature gets despawned => invalid pointer
            m_receiverGuids.reserve(receivers.size());
            for (auto receiver : receivers)
                m_receiverGuids.push_back(receiver->GetObjectGuid());
```

AiDelayEventAround is used by SendAIEventAround, which was being used before for the pull assistance. The difference is that because (prior to any commits) the delay was initially 0, it wouldn't use this function. But if a delay is added to the pull reaction then it will do this.

```c++
            if (delay)
            {
                AiDelayEventAround* e = new AiDelayEventAround(eventType, invoker ? invoker->GetObjectGuid() : ObjectGuid(), *m_unit, receiverList, miscValue);
                m_unit->m_events.AddEvent(e, m_unit->m_events.CalculateTime(delay));
            }
            else
            {
                for (Creature* receiver : receiverList)
                {
                    receiver->AI()->ReceiveAIEvent(eventType, m_unit, invoker, miscValue);
                    // Special case for type 0 (call-assistance)
                    if (eventType == AI_EVENT_CALL_ASSISTANCE)
                        receiver->AI()->HandleAssistanceCall(m_unit, invoker);
                }
            }
```

```c++
AI()->SendAIEventAround(AI_EVENT_CALL_ASSISTANCE, enemy, sWorld.getConfig(CONFIG_UINT32_CREATURE_FAMILY_ASSISTANCE_DELAY), radius);
```

So maybe we should be using this? Or is this a special case where it needs its own code which does something similar. 

I guess what I'm saying is why is the NEW MarkCallAssistanceOnPull needed if SendAIEventAround already handles searching nearby to check assist creatures and the event to those is later sent out with a delay if one is specified.

```c++
        // This code is very similar to new MarkCallAssistanceOnPull + CallAssistanceOnPull and already exists

        // Allow sending custom AI events to all units in range
        if ((eventType >= AI_EVENT_CUSTOM_EVENTAI_A && eventType <= AI_EVENT_CUSTOM_EVENTAI_F && eventType != AI_EVENT_GOT_CCED) || eventType > AI_EVENT_START_ESCORT)
        {
            ...
        }
        else // TODO: Expand functionality in future if needed
        {
            // This would be reached for AI_EVENT_CALL_ASSISTANCE
            MaNGOS::AnyAssistCreatureInRangeCheck u_check(m_unit, invoker, radius);
            MaNGOS::CreatureListSearcher<MaNGOS::AnyAssistCreatureInRangeCheck> searcher(receiverList, u_check);
            Cell::VisitAllObjects(m_unit, searcher, radius);
        }

        if (!receiverList.empty())
        {
            // If a delay is specified then sending out the pull event to the cached CreatureList would be sent on a delay...
            if (delay)
            {
                AiDelayEventAround* e = new AiDelayEventAround(eventType, invoker ? invoker->GetObjectGuid() : ObjectGuid(), *m_unit, receiverList, miscValue);
                m_unit->m_events.AddEvent(e, m_unit->m_events.CalculateTime(delay));
            }
```

Velite12:
If I didn't know any better, after reading this I'd say that my PR actually does the same thing that is being done by the latest commit. In both cases, MarkCallAssistanceOnPull is being called before the first delay.

The difference is where in the code CreatureList of nearby creatures who can assist is being created and then being used as GUID list. With my PR, it happens when the SendAIEventAround happens, which is happening instantly because it's being called before the m_events.CalculateTime in void Unit::TriggerAggroLinkingEvent(Unit* enemy) and right after MarkCallAssistanceOnPull. The delay on acting on this list (which gets converted to guids because it has a delay) is then defined by the SendAIEventAround parameter.

With the latest commit, it happens within MarkCallAssistanceOnPull which happens before the first delay and is effectively instantly. Then with this creaturelist created we're manually sending an event with the delay given by the m_events.CalculateTime.

So in both cases it's doing the same thing but with my PR we're using existing functionality.

killerwife:
Original usecase is for scripts when u use list immediately. I couldve made a searcher that aggregates guid to vector immediately i was just lazy this one time.

Velite12:
> Original usecase is for scripts when u use list immediately. I couldve made a searcher that aggregates guid to vector immediately i was just lazy this one time.

Well that's just it, the list is used with delay if a delay is specified. It does both immediate and delay. It's just that we didn't put a delay for the pull before (it was 0 in original code before we started this all). So in my PR i used a delay which switched the operations which is why my PR didn't have server crashes because it was using creaturelist as a guid list instead.

So in other words, in both this PR and latest, the creatures who are going to help are stored at the moment of the pull, and they will only assist after a delay, and both lists are safe to reference if creatures have despawned.

PR #840, opened 2026-07-09
[SD] Add Root for Blade Dance Target 20709

AnonXS:
Movement Flags: 1536 (DisableGravity, Root)

Other option is to remove all the script stuff for this npc, because staticflags handle this i think and manually add SESSILE to it too.


killerwife:
Ur meant to use SetImmobilizedState not SetCombatMovement cos thats not the flags u wrote.

SetLevitate false, SetImmobilizedState true (passive can stay)

PR #841, opened 2026-07-11
Refactor calling for help on pull 

Velite12:
Continuation of https://github.com/cmangos/mangos-tbc/pull/839 but I had to open separate PR because I force pushed on the branch.

Refactor assist on pull to use less code. This achieves the same thing as latest but without having to make a receiverlist like SendAIEventAround does in MarkCallAssistanceOnPull when a delay is specified. This also turns the receiverlist into a guid list so that it's safe to check the references if mobs are despawned by using code that already exists.

In other words, this aims to do what my original PR did but without the peripheral changes I made. I tested as well and it works.

### DIFF
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -2264,14 +2264,7 @@ void Creature::CallForHelp(float radius)
         return;
 
     if (radius <= 0.0f)
-    {
-        // call for help can be 0
-        if (GetCreatureInfo()->CallForHelp > 0)
-            radius = GetCreatureInfo()->CallForHelp;
-        else
-            radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS); 
-    }   
-
+        radius = GetCreatureInfo()->CallForHelp;
 
     MaNGOS::CallOfHelpCreatureInRangeDo u_do(this, GetVictim(), radius);
     MaNGOS::CreatureWorker<MaNGOS::CallOfHelpCreatureInRangeDo> worker(this, u_do);
--- a/src/game/Entities/Creature.cpp
+++ b/src/game/Entities/Creature.cpp
@@ -2254,7 +2254,7 @@ void Creature::CallAssistanceOnPull(Unit* enemy)
         float radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS);
         if (GetCreatureInfo()->CallForHelp > 0)
             radius = GetCreatureInfo()->CallForHelp;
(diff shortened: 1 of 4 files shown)


src/game/Entities/Creature.cpp
```diff
@@ -2234,42 +2234,27 @@ void Creature::CallAssistance(Unit* enemy)
...
-            if (Creature* receiver = GetMap()->GetAnyTypeCreature(receiverGuid))
-            {
-                receiver->AI()->ReceiveAIEvent(AI_EVENT_CALL_ASSISTANCE, this, enemy, 0);
-                receiver->AI()->HandleAssistanceCall(this, enemy); // Special case for type 0 (call-assistance)
-            }
-        }
+        float radius = sWorld.getConfig(CONFIG_FLOAT_CREATURE_FAMILY_ASSISTANCE_RADIUS);
+        if (GetCreatureInfo()->CallForHelp > 0)
+            radius = GetCreatureInfo()->CallForHelp;
+        AI()->SendAIEventAround(AI_EVENT_CALL_ASSISTANCE, enemy, sWorld.getConfig(CONFIG_UINT32_CREATURE_FAMILY_ASSISTANCE_DELAY), radius);
```
killerwife, line 2257: This reverts exactly why I refactored it. The intention was ONE timer, ONE timer event, in specific order.

For the record, this: m_events.AddEvent(new UnitLambdaEvent

Velite12:
I have given it some looking into and I see the difference because events are being pushed onto the "stack", meaning the pullassistance event being called before the other event, even though it has the same time delay, means it doesn't necessarily resolve at the same time. By batching the calls together in 1 event "bus" you can do that. Refactor is only possible if we didn't also have to deal with the linking which should happen at the same time as non-group pull assistance (as far as the event order goes, not time order). Therefore you have to dis-aggregate the event delay code this way because we need the creaturelist to be formed, acted on a delay, but ensured that it is acted on at the same time as other linking types. 

The original issue was that this list wasn't being formed until after the delay at which point the creature may be out of range to call those who would otherwise be in range to assist. And while both approaches solve this issue, we have to act on the list at the same time as the other linking types, so all of this has to be received together at the same time and event call otherwise it can happen out of order (pull assist before group assist). I see that now that there's a very slight time difference that can be exacerbated the more players on the server because of the event order (pull aggro was happening quicker with this pr). And I don't want to make the game more difficult than it has to be.

Therefore this pr is not possible if the intent is to preserve event batching and game difficulty.

killerwife:
There is even simpler thing at play. Guess what the most expensive perf line of code is in all of this.

Velite12:
> There is even simpler thing at play. Guess what the most expensive perf line of code is in all of this.

If I had to guess I would say it's the object search. But I haven't profiled the code thoroughly.

killerwife:
Nah. "new TimerEvent". By far.

Velite12:
Where is that? can't see in visual studio code.

image

PR #842, opened 2026-07-15
Fix Hungarfen mushroom spell bug and behavior

Velite12:
Fix the mushroom spell on Hungarfen not working properly.

The spell should leave a cloud that does damage and also applies a DoT debuff to players standing nearby. It works very similarly to Rotten Mushrooms in SSC, which works correctly. Currently, it's leaving a cloud but you then get an aura that causes all enemies nearby to receive the DoT, and since each DoT stacks per caster, the boss would receive 5 versions of this DoT and die quickly.

Also, we shrink the mushrooms when they spawn to start small (correct) but we forgot to shrink them once we reach the shrink timer after they cast their Spore cloud spell (we removed grow but forgot to actually cast shrink).

So I looked at how the Spore Cloud spell in SSC works for Rotten Mushrooms and I emulated the same requirements.

Proof:

https://youtu.be/bGLJ4zhHH7M?t=193

You can see that players are supposed to receive the debuff, not the mushrooms or the boss.

Harder to see, but when mushrooms explode, they should shrink after some time, which it looks like the code intended but they weren't actually shrinking.

TODO: Now that players are receiving the dot correctly, and this debuff stacks to 5, we need to determine if this is unique per each mushroom that casts the spore cloud or if they all share it and each application just adds the same stack count to 5. At the moment, every spore cloud has a unique debuff, so it will stack up to max debuff count given enough mushrooms... quickly killing players especially in normal mode.

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/underbog/boss_hungarfen.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/underbog/boss_hungarfen.cpp
@@ -69,6 +69,12 @@ struct boss_hungarfenAI : public CombatAI
         DoCastSpellIfCan(nullptr, SPELL_DESPAWN_MUSHROOMS, CAST_TRIGGERED);
     }
 
+    void EnterEvadeMode() override
+    {
+        DoCastSpellIfCan(nullptr, SPELL_DESPAWN_MUSHROOMS, CAST_TRIGGERED);
+        CombatAI::EnterEvadeMode();
+    }
+
     void ExecuteAction(uint32 action) override
     {
         switch (action)
@@ -179,6 +185,16 @@ struct DespawnMushrooms : public SpellScript
     }
 };
 
+// 34168 - Spore Cloud (Underbog)   
+struct SporeCloud : public AuraScript
+{
+    void OnPeriodicTrigger(Aura* aura, PeriodicTriggerData& data) const override
+    {
+        data.caster = aura->GetCaster();
+        data.target = aura->GetTarget();
+    }
+};
+
 void AddSC_boss_hungarfen()
 {
     Script* pNewScript = new Script;
@@ -192,4 +208,5 @@ void AddSC_boss_hungarfen()
     pNewScript->RegisterSelf();
 
     RegisterSpellScript<DespawnMushrooms>("spell_despawn_underbog_mushrooms");
+    RegisterSpellScript<SporeCloud>("spell_spore_cloud_underbog");
 }
--- a/src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/underbog/boss_hungarfen.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/underbog/boss_hungarfen.cpp
@@ -132,7 +132,7 @@ struct mob_underbog_mushroomAI : public ScriptedAI
         {
             if (m_uiSporeTimer <= uiDiff)
             {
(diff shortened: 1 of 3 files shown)


src/game/Spells/SpellAuras.cpp
```diff
@@ -1933,7 +1933,8 @@ void Aura::TriggerSpell()
                 casterGUID = target->GetObjectGuid();
                 break;
             }
-            case 38652:                                     // Spore Cloud
+            case 34168:                                     // Spore Cloud (Underbog)
```
killerwife: Throw the script into his file so its localized.
OnPeriodicTrigger is the hook for that. If you can be arsed also move the ssc one to ssc.

killerwife (changes requested):

src/game/Spells/SpellAuras.cpp
Velite12: Example on how to do this? Not clear to me.

Velite12: We have to update the db to do this too?

```sql
INSERT INTO `spell_scripts` (`Id`, `ScriptName`) VALUES ('34168', 'spell_spore_cloud_underbog');
```

Velite12:
```c++
// 34168 - Spore Cloud (Underbog)   
struct SporeCloud : public AuraScript
{
    void OnPeriodicTrigger(Aura* aura, PeriodicTriggerData& data) const override
    {
        data.caster = aura->GetCaster();
        data.target = aura->GetTarget();
    }
};
```

And this should be it no?

killerwife: Pretty much, just gotta also Register it below and add it to here: 

https://github.com/cmangos/mangos-tbc/blob/master/sql/scriptdev2/spell.sql#L359

Velite12: When does this sql actually get added to the database? Seems it wasn't on server launch.

Velite12: Well there you go, I made the changes requested but for Underbog only since I haven't got around to SSC yet and this PR is focused on this boss in Underbog.

I also went and added a despawn mushroom call at the moment he evades. Since the spell cast is only 100yds, and he was only casting it from his home point, if you were to drag him our further than that, say near the entrance, those mushrooms would just hang around forever since they're not supposed to despawn themselves.  The boss dying/reset does that.

Velite12: And this video is proof that he definitely tells them to despawn near him at or around the time he evades and he's far from home. From actual TBC heroic.

https://www.youtube.com/watch?v=KN_MG7fahW8

The only unknown here is if it is possible for separate instances of the spore cloud debuff to exist. The debuff itself stacks to 5 but so far I've not seen a video where someone has 2 of the same debuff on them.

Current behavior is that every mushroom applies its own copy of the debuff so you can actually hit the debuff cap pretty quickly if you manage to stay alive long enough. If this is not correct then additional changes are needed.

src/game/AI/ScriptDevAI/scripts/outland/coilfang_reservoir/underbog/boss_hungarfen.cpp
```diff
@@ -178,6 +185,16 @@ struct DespawnMushrooms : public SpellScript
     }
 };
 
+// 34168 - Spore Cloud (Underbog)   
+struct SporeCloud : public AuraScript
```
killerwife, line 189: Just a heads up, due to c++ mangling, class names have to be unique, else on gcc it borks. So if another spore cloud is added in the future it would become an issue. But thats not a problem rn.

killerwife (approved):

PR #843, opened 2026-07-22
[Karazhan] Fix Midnight not respawning after wiping in P3

miraco:
Midnight gets despawned when P3 starts, so its not in storage anymore and check fails

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


killerwife:
Ik im a bit late to this party, but pls add string id and use that in core instead. I would like the only hardcode string ids and not actual IDs in case we refactor db in some way at some point.

SpawnGroup* spawnGroup = player->GetMap()->GetSpawnManager().GetSpawnGroup("SILITHUS_HIVE_ASHI_TRAP_AREATRIGGER");

like i did here

miraco:
Done

PR #848, opened 2026-08-05
[Karazhan] Fix chess event

Velite12:
Address various issues and inaccuracies with Karazhan Chess event, in descending order of severity:

1. When you control a piece, prevent it from running off the board to your teleport spot. This is because they are charmed npcs, and as you'll read, there are problems with using default charm behavior for this encounter. This is one of them.

2. Make the pieces you control actually move when you use movement abilities. This is because by default, when we access the AI of the units, we were not returning the chess_piece AI but instead the default unit AI as they are charmed units. This resulted in them not receiving the custom events that are sent out which are necessary to use the moving spells like Medivh's pieces do.

3. Fixed problems with the Orc unit abilities. This is paired with a DB PR. Basically, the Orc abilities were targeting their own units instead of enemy, and the way the usage was implemented in the script meant that, under AI control, the units would not use the spells. In practical terms, this meant that under AI control, the Orc Warlock and Necrolyte would not use their Poison Cloud/Shadow Mend spells respectively, making the encounter easier for Alliance/harder for Horde players.

4. Fix the Rain of Fire/Poison Cloud spells missing the second tick of damage. This is because the dynamic object that has the persistent aura has a 10 second lifetime but the units don't get the aura debuff immediately, meaning it despawns about 100 or so ms before the second tick comes out (as the spells do 1 tick every 5 seconds). So we just keep it hanging around for a little bit extra to make sure they get the final tick.

5. Fix the AI pieces moving more squares than they were supposed to. And fix the pieces (yours and medivh's) attacking larger pieces from 2 squares away with the default melee ability.

6. Fix the AI pieces not respecting the movement spell GCD, this is due to the technicality of how those pieces move being different than the player's.

7. Approximate 1-2 units moved at a time by Medivh, currently we had it at about 2-3, in some cases 4-5 units moving at a time regularly. 1-2 is much closer to videos, we currently roll a chance to have them move so I lowered it to correlate with this average, but 3+ units moving at a time is still possible.

8. Fix the player controlled king moving by himself due to a technicality with him being the starting piece for the game, player pieces do not move by themselves...

9. Fix the player controlled pieces not doing their spells automatically like before you controlled them, after you stopped controlling them. Essentially when the player is not controlling a piece on their side, they will proceed to still use spells, just not move. Currently, before you controlled the pieces, they would be doing this, but not after you stopped controlling them. Requires a new aura script to recognize when the charm ends.

10. Fix Medivh's fire spell not damaging pieces that players were controlling directly. This is because by default, charmed units use player faction temporarily, and only chess faction units are damaged by the spell. So we just give them their original faction back.

11. The code was assuming that the AI controlled pieces alternated their abilities, but video evidence shows that the usage is random (they can use same ability 4+ times), so fixed that.

12. The chess pieces have no armor, all abilities do unmitigated damage. As part of this, need to tweak code as the rounding function will never return 0. So if the armor multiplier is actually 0, return 0. In addition this comes with DB PR where their armor and agility multipliers are set to 0.

13. When the server shutdown/crashes, and the instance is reloaded, respawn all the pieces. If pieces had died before but the game hadn't ended, then they wouldn't respawn, in particular this can softlock progression as if your king doesn't come back, you can't start the event.

14. The default pet commands (stay/follow/passive/defensive/aggressive etc.) are not visible by patch 2.4.3 for these pieces. Some of these early videos may show them but by 2008 they shouldn't be visible on the action bar.

15. Before some of the pieces move, they do an emote, so I added those emotes.

16. When the game starts, all pieces should have a light beam appear and they all do an attacking emote, so fixed that.

If I forgot to mention anything here, it's probably commented.

Proof (pure TBC sources since I don't trust classic implementation):
https://www.youtube.com/watch?v=eX_4OCouXIw
https://www.youtube.com/watch?v=5932kY_P55A
https://www.youtube.com/watch?v=Fz61qKEI_Lw


src/game/Entities/PetHandler.cpp
```diff
@@ -344,7 +356,18 @@ void WorldSession::HandlePetAction(WorldPacket& recv_data)
             const SpellRangeEntry* sRange = sSpellRangeStore.LookupEntry(spellInfo->rangeIndex);
 
             if (!IsSpellRequireTarget(spellInfo))
-                unit_target = nullptr;
+            {
+                switch (spellInfo->Id)
+                {
+                    // Not sure if TARGET_UNIT_SCRIPT_NEAR_CASTER in general should require target, but these spells
```
killerwife, line 350: TARGET_FLAG_DEST_LOCATION requires target, not TARGET_UNIT_SCRIPT_NEAR_CASTER

src/game/Entities/Creature.h
```diff
@@ -688,7 +688,8 @@ class Creature : public Unit
 
         virtual UnitAI* AI() override
         {
-            if (m_charmInfo && m_charmInfo->GetAI()) return m_charmInfo->GetAI();
+            // Check if not a controlled chess piece
+            if (m_charmInfo && m_charmInfo->GetAI() && !HasAura(30019)) return m_charmInfo->GetAI();
```
insunaa: idk this seems like massive overreach. this function is called constantly for many different reasons all across the entire codebase. checking for a single hardcoded aura doesn't seem like the right solution to me

src/game/Entities/PetHandler.cpp
```diff
@@ -312,6 +312,18 @@ void WorldSession::HandlePetAction(WorldPacket& recv_data)
 
             Unit* unit_target = targetGuid ? _player->GetMap()->GetUnit(targetGuid) : nullptr;
 
+            // Karazhan Chess Pieces: these spells can target any unit but they shouldn't be hitting same faction
```
killerwife: Add this as spell script with OnCheckCast hook.

src/game/Entities/DynamicObject.cpp
```diff
@@ -103,6 +103,8 @@ bool DynamicObject::Create(uint32 guidlow, Unit* caster, uint32 spellId, SpellEf
 
     if (spellId == 44007) // Akilzon - Electrical storm - always takes 500 ms more than spell data - confirmed in sniff
         duration += 500;
+    else if (spellId == 37465 || spellId == 37469) // Karazhan Chess Aoe spells have delay before units get aura so the dynobject would delete early before last tick
```
killerwife: This needs to be solved on a technical level, because I can already imagine a case where 200 wont do the trick. Also need to investigate in sniff, this will not have a solution out in the open by watching vids im afraid.

src/game/Entities/Creature.h
```diff
@@ -688,7 +688,8 @@ class Creature : public Unit
 
         virtual UnitAI* AI() override
         {
-            if (m_charmInfo && m_charmInfo->GetAI()) return m_charmInfo->GetAI();
+            // Check if not a controlled chess piece
```
killerwife, line 691: This is not correct solution. Please check ACTION_TRIGGERS_WHILE_CHARMED and its implementation. To my knowledge chess pieces should use that.

src/game/Entities/Creature.cpp
```diff
@@ -1358,8 +1358,10 @@ void Creature::SelectLevel(uint32 forcedLevel /*= USE_DEFAULT_DATABASE_LEVEL*/)
             powerMultiplier = cinfo->PowerMultiplier;
 
         // armor
-        if (cinfo->ArmorMultiplier >= 0)
+        if (cinfo->ArmorMultiplier > 0)
             armor = std::round(cCLS->BaseArmor * cinfo->ArmorMultiplier);
+        else if (cinfo->ArmorMultiplier == 0) // round doesn't allow 0, this is for certain npcs like Kara chess pieces
```
killerwife: Whats the evidence for 0 armor and what is the problem. Based on https://en.cppreference.com/cpp/numeric/math/round

it says it handles 0, "If num is ±0, it is returned, unmodified."

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/karazhan.cpp
```diff
@@ -376,10 +376,14 @@ void instance_karazhan::Load(const char* chrIn)
             i = NOT_STARTED;
     }
 
-    if (m_auiEncounter[8] == DONE) // if chess event is done, enable friendly games
+    if (m_auiEncounter[8] >= DONE) // if chess event is done, enable friendly games
     {
         m_bFriendlyGame = true;
     }
+    else // reset the event in case server was shutdown to prevent progress softlock
```
killerwife, line 383: Make this explicit for checking specific state, I am guessing, IN_PROGRESS.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
```diff
@@ -551,11 +571,18 @@ struct npc_chess_piece_genericAI : public Scripted_NoMovementAI
                         m_fCurrentOrientation = m_creature->GetOrientation();
                         // workaround which provides specific move target
                         if (Unit* spellTarget = GetMovementSquare())
-                            spellTarget->CastSpell(m_creature, SPELL_MOVE_TO_SQUARE, TRIGGERED_NONE);
+                        {
+                            CanCastResult result = spellTarget->AI()->DoCastSpellIfCan(m_creature, SPELL_MOVE_TO_SQUARE, TRIGGERED_NONE);
```
killerwife, line 563: I believe this should use spell id 32306

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
```diff
@@ -408,7 +409,19 @@ struct npc_chess_piece_genericAI : public Scripted_NoMovementAI
     {
         // do a soft reset when the piece is controlled
         if (pCaster->GetTypeId() == TYPEID_PLAYER && pSpell->Id == SPELL_CONTROL_PIECE)
+        {
             Reset();
+            // Restore original faction so medivh's cheat spell can damage us, since normally charm unit faction = player unit faction
+            m_creature->RestoreOriginalFaction();
+            CharmInfo* charmInfo = m_creature->GetCharmInfo();
+            MotionMaster* mm = m_unit->GetMotionMaster();
```
killerwife: Cant do this, because if you start controlling mid move, it will break. In fact motion master must not be touched outside of the move action and facing action and all else has to be disabled.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
```diff
@@ -132,12 +132,12 @@ enum
     SPELL_ELEMENTAL_BLAST           = 37462,                    // human queen
     SPELL_RAIN_OF_FIRE              = 37465,
     SPELL_FIREBALL                  = 37463,                    // orc queen
-    // SPELL_POISON_CLOUD           = 37469,
-    SPELL_POISON_CLOUD_ACTION       = 37775,                    // triggers 37469 - acts as a target selector spell for orc queen
+    SPELL_POISON_CLOUD              = 37469,
+    // SPELL_POISON_CLOUD_ACTION    = 37775,                    // triggers 37469 - acts as a target selector spell for orc queen (npc can't use this)
```
killerwife: Why not?

src/game/Spells/SpellEffects.cpp
```diff
@@ -6252,9 +6254,18 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
                     return;
                 }
                 case 39341:                                 // Karazhan - Chess, Medivh CHEAT: Fury of Medivh, Target Horde
```
killerwife, line 6255: Port these to the file as well.

src/game/Spells/SpellEffects.cpp
```diff
@@ -6062,7 +6063,8 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
                 case 37345:                                 // Karazhan - Chess NPC Action: Melee Attack: Orc Warlock
                 case 37348:                                 // Karazhan - Chess NPC Action: Melee Attack: Warchief Blackhand
                 {
-                    if (!unitTarget)
+                    // check if we are within range and facing the target to attack; prevents ai pieces from attacking past 2 tiles
```
killerwife: This check should ideally be either in OnCheckTarget or in OnCheckCast

src/game/Spells/SpellEffects.cpp
```diff
@@ -6252,9 +6254,18 @@ void Spell::EffectScriptEffect(SpellEffectIndex eff_idx)
...
+                        return;
+
+                    m_caster->CastSpell(unitTarget, m_spellInfo->CalculateSimpleValue(eff_idx), TRIGGERED_OLD_TRIGGERED);
+                    return;
+                }
                 case 39344:                                 // Karazhan - Chess, Medivh CHEAT: Fury of Medivh, Target Alliance
                 {
-                    if (!unitTarget)
+                    // Prevent targeting Medivh's pieces
+                    if (!unitTarget || unitTarget->GetFaction() != 1690)
```
killerwife: Same, OnCheckTarget

src/game/Spells/Spell.cpp
```diff
@@ -3073,11 +3086,31 @@ SpellCastResult Spell::CheckScriptTargeting(SpellEffectIndex effIndex, uint32 ch
...
                 {
-                    MaNGOS::AllCreatureEntriesWithLiveStateInObjectRangeCheck u_check(*caster, entriesToUse, type != SPELL_TARGET_TYPE_DEAD, radius, type == SPELL_TARGET_TYPE_CREATURE_GUID, false, true);
-                    MaNGOS::CreatureListSearcher<MaNGOS::AllCreatureEntriesWithLiveStateInObjectRangeCheck> searcher(foundScriptCreatureTargets, u_check);
-                    Cell::VisitAllObjects(caster, searcher, radius); // Visit all, need to find also Pet* objects
-                    if (u_check.FoundOutOfRange())
-                        foundButOutOfRange = true;
+                    switch (m_spellInfo->Id)
+                    {
+                        // Karazhan Chess Poison Cloud needs to check range, not radius from caster
+                        case 37469:
```
killerwife, line 3092: This stuff is the direct result of the spell you wrote above that cannot be used, and I can say for 100% certain this is wrong. That being said, I am not saying there is some other mystical spell for this. These targeting thingies generally cause most bugs and break most often over the years. (code duplicity essentially)

src/game/Spells/Spell.cpp
```diff
@@ -3056,10 +3056,23 @@ SpellCastResult Spell::CheckScriptTargeting(SpellEffectIndex effIndex, uint32 ch
...
                         (type == SPELL_TARGET_TYPE_DEAD && ((Creature*)target)->IsCorpse()))
                     {
-                        if (target->IsWithinDistInMap(caster, radius) && OnCheckTarget(target, effIndex))
-                            foundScriptCreatureTargets.push_back((Creature*)target);
-                        else
-                            foundButOutOfRange = true;
+                        switch (m_spellInfo->Id)
+                        {
+                            // Karazhan chess Rain of Fire casts directly on target but needs to check within radius around target, not caster
+                            case 37465:
```
killerwife, line 3062: I believe the asumption here is wrong.

The reason being is that you already are picking targets AROUND caster:

TARGET_UNIT_SCRIPT_NEAR_CASTER - the NEAR_CASTER part

Also it isnt cast on target, but on DEST: TARGET_FLAG_DEST_LOCATION (the spell doesnt accept a unittarget, but xyz)

That being said, you can still do shenanigans in OnCheckTarget, its not like LOS matters on chess board anyway.

The target that does what you thought this one did is: TARGET_ENUM_UNITS_SCRIPT_AOE_AT_DYNOBJ_LOC

But thats clearly not used here (often because early tbc didnt have it)

killerwife (commented):
Also check these out:

Image

I sort of never tried to improve chess, just always made it work, which i guess broke due to the static flag i linked and it should be patchable in some spots.

src/game/Entities/Creature.cpp
Velite12: Look at the videos and read the description. All the spells must do by default the damage listed on the tooltip. Because of the nonzero armor and agility given to the units that is not the case.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
Velite12: So what do you suggest because this is solving an encounter breaking issue? Without this you can't beat the encounter. It's literally unplayable.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
Velite12: Because they can't. The way the scripts are set up, the orc necrolytes wouldn't use their heal. But the human clerics use their heal... so something is wrong. I instead opted to have them use the heal spell directly instead of going through this action and it works.

src/game/Entities/Creature.h
Velite12: This is the getter for m_ai. We don't have a way to get it directly even if the unit is charmed. We have to bypass the charmed checks so that the unit can use the moving spells as coded in the chess pieces file.

src/game/Entities/DynamicObject.cpp
Velite12: So how do we get persistent aura to apply immediately as soon as the dynobject is created? I have to look at how other spells handle this situation, or we're just missing final ticks in general?

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/karazhan.cpp
Velite12: This needs to happen if you failed the event too. That's precisely the edge case this is meant to handle, if the king dies but the board is not yet reset and the server shuts down.

src/game/Spells/Spell.cpp
Velite12: No, you're confusing two different things.

First of all, this code came after I had fixed the reason why the orc units couldn't use 2 of the spells. Not the other way around. This code isn't the reason why they couldn't use those spells.

Read carefully my comments. Notice what's missing here? The shadow mend spell. But how did I fix the npcs using the shadow mend spell? Because in the chess event script it was trying to use it like the human counterpart units. But those units just cast the spell directly, so that's what I made the orc units do. So this code here is not necessary for the npcs to use the ability...

What this is addressing is the fact that YOUR unit as controlled as a pet is unable to cast this spell at the appropriate range like the NPCs can due to pet technicalities. This and the Rain of Fire spell keep saying "your pet is out of range" from 3 squares away, even though that should be a valid distance, because the distance checking is using the wrong value. But because poison cloud uses a targeting reticle and rain of fire uses a selected unit, they had to be solved in different places.

In other words, this code does not in any way affect the comment I made in the other file, unlike what you're saying. These address separate issues.

src/game/Spells/Spell.cpp
Velite12: If you check the videos, you cast the spell on the unit you have selected. The spell_script_target file lists the eligible units it can be cast on. This spell doesn't have a targeting reticle which is the source of this issue.

As mentioned in another reply, this is solving the problem of why your chess piece when controlled by YOU can't cast the spell at the correct range. I went through code using breakpoints. It correctly determines that you are in range using this function for range around caster, but then the code later runs this function again, and tries to check RADIUS around the caster... but the radius is much shorter than the spell range, so it says "your pet is out of range" error message, and the unit is stuck in a casting state and doesn't cast the spell. It needs to check if the unit you are targeting is within the radius of the spell cast. Or, just not run this pass again and only check if we are within range. You can't explicitly cast this spell on a specific location.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
killerwife: Go check ACTION_TRIGGERS_WHILE_CHARMED

Velite12: Yes as I wrote later I realize that this would solve the problem and several others.

image

In our code we only check for the flag. How do we give it? DB?

src/game/Entities/Creature.cpp
killerwife: But I asked a C++ question, why is there a problem with std::round when documentation says otherwise.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
killerwife: Did you check combat reach for bad data in db?

src/game/Spells/Spell.cpp
Velite12: And by the way, this is also after making the DB changes necessary so that the orc units are not hit by the poison cloud spell, as we had mistakenly assigned the orc unit ids as the targets for the poison cloud in spell_script_target and not the human unit ids. So it may look like you can cast the spell before these changes but that's only because the game sees that the caster (your controlled piece) is within the spell radius (not range which is much larger) of the units the spell should target (your fellow orcs, but that's wrong). Now that it's targeting the humans, it fails to cast because the humans start on the other side of the board.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
Velite12: > Did you check combat reach for bad data in db?

The combat reach for the spell? It's not the unit, they use the melee spell (the one where they attack unit within 1 tile) fine...

src/game/Spells/Spell.cpp
killerwife: I am merely stating how targeting behaves in the current data. I am not doubting the fact the symptom of the problem is in this location. I am saying how the current spell data is meant to behave. But I am not saying its the correct spell data, that is a very important distinction. Client is checking distance from caster to target. That much I know 100% and I have no problem to fight 1v1 on that one anymore. Script targeting is purely serverside, hence there is some wiggle room. 

Have you considered that the small npcs in the room that you target with movement etc might be in wrong position for example?

Velite12: You mean for the moving spell? As a gm I can see they are in the center of the squares which is what the targeting reticle for the moving spell is checking for when it "hits". But this isn't a moving spell so forgive me if I'm not seeing your point. In addition this is a problem even when you haven't moved the piece.

The units that this can target (that being the orc ids) are the ones out of the radius which is why this fails the second distance check. I would assume they are where they're supposed to be? These are the npc controlled pieces and the other code which checks for where they are in order to move is correct. There was an issue before where it was possible for a unit to move (and thus reserve the square so someone else can't move to it), but that was only for your side.

src/game/Entities/Creature.cpp
Velite12: Ok I concede that the code here can indeed account for 0 multiplier. Then the leftover must have been the agility.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/karazhan.cpp
Velite12: Image

Here you can see that we set the instance data to fail immediately once the king is dead. But he does not respawn until 35 seconds later. In this window you can be softlocked. So it needs to be more than IN_PROGRESS.

src/game/Entities/Creature.h
Velite12: Resolved by DB edit

src/game/Entities/Creature.cpp
killerwife: Well, agility is a fairly new addition as are all npc stats. Need to come up with an idiotic test case, for example like I found Damage Multi applies even to AP buffs/debuffs, could easily be same case with agi armor.

Velite12: Yes because even setting the armor multi to 0, they only take 3906 from the cleave/sweep spell the king does when it should be 4000, without any buffs or debuffs. There is code somewhere, I think in creature, where it adds the armor contribution from the agility. So setting the agility to 0 causes the spell to do the 4000 damage.

src/game/Entities/PetHandler.cpp
Velite12: This spell just seems to require a target when you cast with the pet actionbar, and the npc caster (with unmodified code) casts this with its center on a particular orc unit. The poison cloud doesn't because it uses a targeting reticle to cast on an xyz.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
Velite12: 32306 has a dummy and we don't have that spell implemented in the core.

Velite12: Second:
image
We use 30253 to move the piece, SPELL_MOVE_TO_SQUARE. This is for both player controlled and npc controlled. And it works fine as it is right now. This code here is just to add what the GCD would give for casting the spell to the move timer. As I noted in the description, while the spell assigns GCD to the piece (therefore you can't cast the move again for 12 seconds for your controlled piece), for npc the square moves the piece here, so it doesn't have the GCD.

So I don't really know how we'd implement this new spell (well, it's in the DB but it's unused), and the behavior here works already, just needs to set the move timer properly for the script.

You can't give the GCD to the square because by the time the move is executed you'll be selecting a different square to move with. The delay has to be specific to the piece.

Velite12: I honestly don't know how to implement what you're suggesting and it's not necessary. I tried using this ability and setting it up like how the player does a move but it doesn't work. Whichever combo I used it as (target->caster, caster-target etc) they did not move anymore. I think this is similar to the issue with the shadow mend action/ poison cloud action spell, the npcs cant use these spells or the code isn't implementing them properly when they try to use them. Either way, it's not necessary to change what spell is being used here. Even if we did, we still have to modify the timer for the AI here like I did.

Velite12:
This first commit addresses some of the concerns with how I overrode the default charm behavior. Now most of this code isn't needed because they have the proper flag set with the recent db commit.

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/chess_event.cpp
Velite12: For the record, these spells that I commented out were not the ones the player gets on the actionbar, it was just npc controlled using them. If I had to guess it's that these might have been leftover spells from previous patches (as there are quite a few of them).

The human cleric has a healing spell which never had nor has any corresponding action like the shadow mend does. And functionally speaking, shadow mend and the cleric heal work exactly the same, it is just a cosmetic difference. So the event works fine without using the shadow mend action spell.

I can understand maybe with the poison cloud since that spell has a targeting circle but the rain of fire doesn't, but the npcs seem to use it fine without the action spell. Like I said before, because these actions were being used instead of the direct spell, the npc controlled pieces weren't casting the direct spell, making the Horde side weaker than Alliance side. This change makes both sides' pieces balanced.

miraco: Well from Sniff for the case:  // SPELL_POISON_CLOUD_ACTION    = 37775,                    // triggers 37469 - acts as a target selector spell for orc queen (npc can't use this)

```
ServerToClient: SMSG_SPELL_START (0x630029) Length: 123 ConnIdx: 1 Time: 03/29/2026 21:24:16.729 Number: 506009
(Cast) CasterGUID: Full: 0x2064584280153D800377CE0000497142 Creature/0 R6422/S227278 Map: 532 Entry: 21750 Low: 4813122
(Cast) CasterUnit: Full: 0x2064584280153D800377CE0000497142 Creature/0 R6422/S227278 Map: 532 Entry: 21750 Low: 4813122
(Cast) CastID: Full: 0xBC6458428024E3C30377CE0000497C64 Cast/3 R6422/S227278 Map: 532 Entry: 37775 Low: 4815972
(Cast) OriginalCastID: Full: 0x0
(Cast) SpellID: 37775 (37775)
```

it gets used from Orc Warlock - and we are playing alliance so its not by player controle

Velite12: They can reach with the direct spell fine. But if you really wanted to use these spells (which you don't have to), then their respective scripts would have to work differently than their alliance counterparts which is needlessly complicating the problem.

Velite12: > it gets used from Orc Warlock - and we are playing alliance so its not by player controle

So if it must be the case that they use this ability, then how we have 37775 implemented has a problem which is preventing them from using 37469. Because they can use 37469 correctly. How exactly this spell should be used therefore is not understood by me. Is it supposed to replace the target selection code that all of the spells use? Such that these respective spells use this spell to select a target instead of the function? If I read the intent of the comment and name it has to do with getting them to select a target. But again, this target selection action spell doesn't exist for the alliance counterpart and we already have a function which returns a target, where this spell is issued as a parameter.

Velite12: And it's the same with the Orc Necrolyte, the alliance version doesn't have an action spell for the heal. So how would that spell work?

Velite12: Image
In theory the spell should be working already, it just functions as a wrapper for casting 37469. I'll try adjusting some things.

Image
The target selection is fine because this function (GetTargetByType) is what I was using for just casting the spell directly. So I don't see why it doesn't work already...

Velite12: So apparently this spell (37775) is a delayed spell. It has a speed value of 40. But the result returned from DoCastSpellIfCan(pTarget, SPELL_POISON_CLOUD_ACTION); is CAST_OK(0), but the followup doesn't get cast. However, a target is definitely chosen, I can see that they're looking at the target they want to cast on.

Velite12: Image
Okay I see the real problem, this doesn't hit. So it would mean that this case is placed in the wrong function. Because if the spell cast is ok, then it should be reaching here, no?

I think this code needs to be in dummy and not effectscripteffect. For starters the spell 37775 has only 1 effect which is a dummy. 37469 has a script effect.

Velite12: image
Yep that did the trick. So now they can use these actions and they will properly follow through to the direct spell which would actually be used. This therefore satisfies the requirement, even though I think it's a bit silly anyway. I don't get why it has to be different than how the alliance spells work but eh who am I to judge.

Velite12: But now we have a problem with the shadow mend action, where they will do it, but the resulting direct spell it will heal EVERYONE in an area even though it's only supposed to heal 1 target...

Velite12: I got it to heal only the orc units, turns out the spell (shadow mend action)  was missing script targets (the orc units) even though shadow mend has those targets. This would explain why the spell started to heal me and the squares. But the max affected targets in the DB is 0... Are we allowed to change this to 1 so that it's directed at 1 target like the poison cloud action? I noticed that we have a file which overrides the MaxTargets in the client spell data.

image

There's actually a couple of Karazhan spells in this list so it would not be unprecedented. Both hand of medivh spells there have a targets of 0 in the client but here it gets set to 1 (since only 1 unit should be targeted). If I were to revert them to 0, Hand of Medivh would be targeting everyone like this spell is doing... so yeah I'm going to add it to this list.

image

Spell 37824 has these values for implicit targets which is causing the directly casted spell (37479) to heal everyone in an area, but it's only supposed to heal 1 target like the human cleric spell. Hand of Medivh spells have these same targets (22 and 7 for A1 B1) so yeah.

Velite12: image
Yep that's the fix. This keeps the action being used as is required but it functions as if you had just casted the direct spell anyway.

Velite12:
Second commit addresses the requirement that the orc warlock and orc necrolyte must use the action spell for their poison cloud and shadow mend respectively.

src/game/Spells/SpellEffects.cpp
Velite12: This change is redundant

src/game/Spells/SpellEffects.cpp
Velite12: The change in my commit is redundant so no need to change this.

src/game/Spells/SpellEffects.cpp
Velite12: Done

Velite12:
Added spell scripts for all of the melee attacks with OnCheckTarget used to check range and angle

Velite12:
Turns out that by using the actions, the NPCs are casting the spells more often than their alliance counterparts would. This is because action spells have no recovery time, but they have a 5 second category recovery time, so the npc would think they have a 5 second cooldown when in reality it is much greater for the ability.

So we cast the action spell, but extract the recovery time from the direct spells (poison cloud and shadow mend) so the 2 sides (Horde and Alliance) are balanced. This keeps the requirement that the action spells must be casted.

src/game/Entities/PetHandler.cpp
Velite12: Done

Velite12:
The remaining issues with this PR now are:

Casting the aoe spells when you control the chess piece (warlock/conjurer) as pet is failing to cast. NPC casts it fine.
Dealing with how to preserve 2 ticks of damage for the aoe spells (Dynobject doesn't apply aura immediately on spawn)
~~When to reset (needs to be at least IN_PROGRESS and FAILED chess event).~~ Done

src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/karazhan/karazhan.cpp
Velite12: For good measure, let's just have it on IN_PROGRESS or FAILED

src/game/Entities/DynamicObject.cpp
Velite12: I think I have an idea. The spell has a speed value listed in client. Perhaps there is a "travel time" for the aoe to reach even though the green cloud is already there?

miraco:
Please stop using this as your personal thread.

You’re unnecessarily spamming the GitHub channel on Discord.

We don’t need to hear every little thought that crosses your mind, if it helps you do it in a local text file for your self

Velite12:
I have to respond to review comments, of which there are unfortunately many in this pr.

But if it pleases you, then I won't write unless I have to respond.

miraco:
I can btw know confirm that those chess pieces have armor when you controle them
WowClassic_lgQc6cKUpn

Velite12:
> I can btw know confirm that those chess pieces have armor when you controle them WowClassic_lgQc6cKUpn

Not sure what the best way forward is with solving the problem here then. Easiest explanation is 0 armor (even old value assumed that) but now that is not true.

We have to ignore or bypass the armor. I made the assumption that it is ignored by these npcs in the armor calculation, but if anyone has a better idea then reply. There is a flag that makes a spell ignore all resistances but these spells do not have it listed in client nor db. In fact, the wiki does not even list which one this is.

Seeing as how we can't set the armor in the chess_event script this is the only way I found.

PR #854, opened 2026-08-15
Fix OHF bugs

Velite12:
This needs to be merged with https://github.com/cmangos/tbc-db/pull/1301 or encounters will break

Fix Captain Skarloc boss state not going back to fail if you fail the fight and release before Thrall dies. This would cause the Durnholde Warden and Veteran npcs in the preceding waves to not spawn if you had attempted the boss at least once due to a logical oversight, as these creature entries are the same as the adds that accompany Skarloc.

Also, Thrall does not worship Hakkar. He should say the proper text 17609 with voice entry 10457, we had 10457 value listed as his text entry which seems like a copying error. You can see that the text 17609 uses his voice ingame when he says it.

Also, Epoch Hunter is supposed to engage with Thrall (the code even indicates this) after the last wave is dead, but he was not attacking because he was not doing SetInCombatWithZone().

Also, the waves of enemies when you first free Thrall do not attack Thrall immediately. As they are not in combat, you can sap/cc them ahead of time. 

This also addresses the TODO that says the npcs during Epoch Hunter should attack once Thrall reaches home, which in the videos it does appear like they do not attack immediately after the last group dies. Epoch Hunter himself attacks Thrall once he reaches home.

Also fixes Thrall not remounting the horse while running from keep->barn if he was in combat and him facing the wrong way when meeting Skarloc.

Also the barrel diversion counter should disappear at the same time as the quest credit, not when Lieutenant Drake enters combat. It's kinda hard to tell because it's cutoff here but at the top of the middle is the counter.
https://www.youtube.com/watch?v=x2kRsK17Lz0

Proof:
Classic:
This video is a good showcase for the dialogue we are missing, especially ones that don't involve Thrall:
https://www.youtube.com/watch?v=lVWiffqhFOY
TBC:
https://www.youtube.com/watch?v=dqh8grFq0sU

For Durnholde waves not going right away into combat see:
https://www.youtube.com/watch?v=afbf54etkcY

### DIFF
--- a/src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
@@ -434,7 +434,7 @@ struct npc_thrall_old_hillsbradAI : public npc_escortAI, private DialogueHelper
                 }
             }
         }
-        else if (m_pInstance && m_uiEpochWaveId == 3 && m_bEpochAttackTimer <= 0)
+        else if (m_pInstance && m_uiEpochWaveId == 3 && m_epochAttackTimer <= 0)
         {
             if (Creature* epoch = m_pInstance->GetSingleCreatureFromStorage(NPC_EPOCH))
             {
@@ -973,7 +973,7 @@ struct npc_thrall_old_hillsbradAI : public npc_escortAI, private DialogueHelper
     {
         UpdateEscortAI(uiDiff);
         if (m_pInstance && m_uiEpochWaveId == 3)
-            m_bEpochAttackTimer -= uiDiff;
+            m_epochAttackTimer -= uiDiff;
     }
 
     void UpdateEscortAI(const uint32 uiDiff) override
--- a/src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
@@ -1297,7 +1297,7 @@ struct npc_tarethaAI : public npc_escortAI, private DialogueHelper
                 m_creature->RemoveFlag(UNIT_FIELD_FLAGS, UNIT_FLAG_UNINTERACTIBLE);
                 m_creature->SetStandState(UNIT_STAND_STATE_STAND);
 
-                Start(true, nullptr, nullptr, false, false, THRALL_PATH_ID);
+                Start(true);
                 m_bHasStartedEpilogue = true;
             }
         }
--- a/src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
+++ b/src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
@@ -104,7 +104,7 @@ enum
     SAY_LOOKOUT_BARN_1              = 15958,
     SAY_PROTECTOR_BARN_2            = 15968,
     EMOTE_TH_CALM_HORSE             = 15970,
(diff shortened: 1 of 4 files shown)


al3xc1985:
https://github.com/cmangos/issues/issues/4150

I did some observations here also, maybe u didn't saw them

src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
```diff
@@ -750,13 +776,17 @@ struct npc_thrall_old_hillsbradAI : public npc_escortAI, private DialogueHelper
                 break;
             // *** Escort event - Part I - meet Skarloc ***
             case 35:
+            {
                 m_pInstance->SetData(TYPE_SKARLOC, IN_PROGRESS);
-                m_creature->SummonCreature(NPC_SKARLOC, 2000.201f, 277.9190f, 66.4911f, 6.11f, TEMPSPAWN_DEAD_DESPAWN, 0);
+                Creature* pSkarloc = m_creature->SummonCreature(NPC_SKARLOC, 2000.201f, 277.9190f, 66.4911f, 6.11f, TEMPSPAWN_DEAD_DESPAWN, 0);
```
killerwife: Name these new vars without the hungarian notation. I dont expect you to rename whole script like I do but new stuff should be compliant

src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
```diff
@@ -419,6 +419,36 @@ struct npc_thrall_old_hillsbradAI : public npc_escortAI, private DialogueHelper
         }
     }
 
+    void JustReachedHome() override 
+    {
+        // Have the Epoch Hunter and adds attack once Thrall reaches home, not on summon
```
killerwife: I would expect there to be a delay and not immediate hm

killerwife (changes requested):

src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
Velite12: https://www.youtube.com/watch?v=ZyNyfxsI574

This video is 2.4.3. There are some periods in between pulls where you are not in combat, which doesn't happen currently. Additionally there appears to be a hard delay before Epoch Hunter attacks Thrall. But once he reaches home they will start to attack, hence putting people in combat.

https://www.youtube.com/watch?v=r3gCO7mmGMA

this is classic anni, at 48:17 can see the same non-combat period, however there is a difference in that the 2nd pull doesn't attack Thrall immediately, nor does Thrall go to engage. In the 2.4.3 video they will go to attack Thrall. But they do spawn in right after the previous group was killed, but like this PR and TODO claims, they do not attack immediately else they would be running as they appear like current.

https://youtu.be/jdQ6gL5PwRc?t=406

This is an early TBC video and you can see that while they do attack on spawn, there is a delay between Thrall reaching home and when they actually spawn thus attack. Behavior was different through TBC and between TBC and Classic.

What would be the best way to add the delay to Epoch Hunter? uidiff and give him his own script? Or timer?

src/game/AI/ScriptDevAI/scripts/kalimdor/caverns_of_time/old_hillsbrad/old_hillsbradScripts.cpp
```diff
@@ -283,6 +282,21 @@ struct npc_thrall_old_hillsbradAI : public npc_escortAI, private DialogueHelper
...
         }
+        if (m_bHadMount)
+        {
+            m_creature->Mount(MODEL_SKARLOC_MOUNT);
+            m_bHadMount = false;
+        }
+            
+    }
+
+    void EnterCombat(Unit* enemy) override
```
killerwife, line 293: Add call of parent EnterCombat

PR #859, opened 2026-08-19
OHF things I missed

Velite12:
Oversight with counting the Epoch waves; accidentally made Epoch Hunter attack during last wave of adds instead of after, so fixed that.

Also, in fixing the mobs which transform to only have a specific set abilities, need to allow SCRIPT_COMMAND_UPDATE_TEMPLATE  to init the AI again so they will have the new list of abilities. This fixes the first Epoch Hunter wave still using the human abilities like serpent sting and cleanse/heal, instead of the dragonkin abilities like shadow shock and mortal strike.

This is in line with this function (Creature::UpdateEntry) updating the spell lists for the new entry. But these mobs in question in OHF do not use spell lists, they use creature_ai_scripts instead, so we must update these as well.


src/game/DBScripts/ScriptMgr.cpp
```diff
@@ -2940,7 +2941,16 @@ bool ScriptAction::ExecuteDbscriptCommand(WorldObject* pSource, WorldObject* pTa
             Creature* pCSource = static_cast<Creature*>(pSource);
 
             if (pCSource->GetEntry() != m_script->updateTemplate.newTemplate)
+            {
                 pCSource->UpdateEntry(m_script->updateTemplate.newTemplate);
+                // reset EventAI if new entry
+                if (pCSource->GetAIName() == "EventAI")
```
killerwife: This is actually wrong. B swaps spell lists on entry change, but not AI. So if u need the different combat spells, use spell lists and update record.

killerwife (changes requested):

src/game/DBScripts/ScriptMgr.cpp
Velite12: Not going to work because the different creatures have different say scripts on death/aggro. Spells lists just handles the spells part of it, we'd still need to change the scripts they point to for the say.

image

Also the dragonkin don't flee... so we'd have to not have that on their scripts as well when we do this.

I had considered just spell list reorganizing it but this is the hangup against doing that to begin with.

So we need some way to prevent these scripts from happening and instead use the proper scripts for the new entry... makes sense to do it here, no? Or would we have to create a new command to do that?
image

killerwife: Spell 33133 - once its spell based you can change AI. Do the UpdateEntry and AI change in dummy script and we are good. I just confirmed in sniff thats how this works.

src/game/Spells/SpellEffects.cpp
```diff
@@ -1668,6 +1669,37 @@ void Spell::EffectDummy(SpellEffectIndex eff_idx)
                     ((Creature*)unitTarget)->ForcedDespawn(1000);
                     return;
                 }
+                case 33133:                                 // Transform
```
killerwife: Add it as spell script, not like this. We never add to these switches anymore.

killerwife (changes requested):

Velite12:
Too soon, I just fixed the compile issue!

PR #861, opened 2026-08-20
Fix mobs treating new targets in combat as though they were first in combat

Velite12:
Currently in dungeons, when you are near mobs that are fighting someone else, and that someone else has 0 threat on them, when those mobs place combat status on you by virtue of being nearby, they will switch to treating you as the new victim and not the person who went in first, even if everyone has 0 threat. If a third person then went near these mobs and everyone still had 0 threat, that third person would then become the victim, and so on and so forth which is a behavior not shown in real tbc.

This is especially noticeable in the Old Hillsbrad Foothills dungeon where often Thrall will be attacked and placed in combat and thus by extension, you are too, and because you get placed after Thrall does, they go to attack you and not Thrall because of this line of code, they act as though there isn't already a current target to attack and combat just started every time.

Take for example this video from 2.4.3. https://youtu.be/6lOWgKhAMvU?t=257

At 4:17, when the next wave shows up, everyone gets placed in combat. There is no threat assigned to Thrall but the mobs are scripted to attack him first. Now you see in the video they do actually go up to Thrall and start attacking him, even though the player filming the video was placed in combat (the crossed swords above his level), but he has not done anything to merit threat so they do not attack him.

In our server in this situation, the mobs would target Thrall at first, then they would then quickly switch to attacking someone else who was just placed in combat instead of focusing on Thrall, even if everyone in the party and Thrall has 0 threat. That is because of this line not accounting for current victim already. If hypothetically the distance is determining the order of who gets assigned combat (so as the mobs go closer, people get added to combat), then the furthest player from the spawn would become the ultimate victim rather than the first thing they find or, in this case, the intended target of Thrall. Meaning it is very well possible that the mobs are going to go after your healer if they are in the back, rather than the tank up front, if nobody does anything and the healer is added last to combat. This is of course a problem if the mobs have instant attack abilities that they then dump on your healer immediately on spawn, rather than the target that should be attacked (Thrall or the nearest target to them).

Basically unless the mobs are scripted to choose random targets, in a dungeon, the person who first attracts the attention of mobs should be the target of their auto attacks (victim), until someone does something. It should not depend on new people being placed in combat to change this current victim. If there is random spell targeting or mob random targeting scripted, that is a different story, but I am referring to mobs who would have no such mechanics, which is the majority.

So to recap, if everyone has 0 threat and you join in combat after someone else, you should be placed in combat, you should get a threat value assigned to you (usually 0 by AddThreat), but if the mobs already have someone with 0 threat to attack, then they shouldn't switch to auto attacking you and following you around. Once someone has above 0 threat, this doesn't matter as threat rules start to apply.

### DIFF
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -8407,7 +8407,7 @@ void Unit::SetInCombatState(bool PvP, Unit* enemy)
                     if (PvP)
                         enemy->GetCombatManager().TriggerCombatTimer(controller);
                 }
-                else if (controller->CanHaveThreatList() && controller->getThreatManager() && !controller->getThreatManager().getCurrentVictim())
+                else if (controller->CanHaveThreatList() && controller->getThreatManager() && !controller->getThreatManager().isThreatListEmpty && !controller->getThreatManager().getCurrentVictim())
                 {
                     MANGOS_ASSERT(controller->AI()); // a player without UNIT_FLAG_PLAYER_CONTROLLED should always have AI
                     controller->AI()->AttackStart(enemy);
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -8407,7 +8407,7 @@ void Unit::SetInCombatState(bool PvP, Unit* enemy)
                     if (PvP)
                         enemy->GetCombatManager().TriggerCombatTimer(controller);
                 }
-                else if (controller->CanHaveThreatList() && controller->getThreatManager().isThreatListEmpty())
+                else if (!controller->GetVictim())
                 {
                     MANGOS_ASSERT(controller->AI()); // a player without UNIT_FLAG_PLAYER_CONTROLLED should always have AI
                     controller->AI()->AttackStart(enemy);
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -8407,7 +8407,7 @@ void Unit::SetInCombatState(bool PvP, Unit* enemy)
                     if (PvP)
                         enemy->GetCombatManager().TriggerCombatTimer(controller);
                 }
-                else
+                else if (!controller->GetVictim())
                 {
                     MANGOS_ASSERT(controller->AI()); // a player without UNIT_FLAG_PLAYER_CONTROLLED should always have AI
                     controller->AI()->AttackStart(enemy);
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -8407,7 +8407,7 @@ void Unit::SetInCombatState(bool PvP, Unit* enemy)
                     if (PvP)
                         enemy->GetCombatManager().TriggerCombatTimer(controller);
                 }
-                else if (controller->CanHaveThreatList() && controller->getThreatManager() && !controller->getThreatManager().isThreatListEmpty && !controller->getThreatManager().getCurrentVictim())
+                else if (controller->CanHaveThreatList() && controller->getThreatManager().isThreatListEmpty())
                 {
                     MANGOS_ASSERT(controller->AI()); // a player without UNIT_FLAG_PLAYER_CONTROLLED should always have AI
                     controller->AI()->AttackStart(enemy);
--- a/src/game/Entities/Unit.cpp
+++ b/src/game/Entities/Unit.cpp
@@ -8407,7 +8407,7 @@ void Unit::SetInCombatState(bool PvP, Unit* enemy)
                     if (PvP)
                         enemy->GetCombatManager().TriggerCombatTimer(controller);
                 }
-                else if (!controller->GetVictim())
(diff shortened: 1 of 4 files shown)


killerwife:
So this overall is NPC case. NPC has pet. Pet gets into combat and now NPC is called upon to enter as well. NPC should never have controller->GetVictim != nullptr and yet !controller->IsInCombat()

So your code in question essentially deletes the Else case for NPCs cos it should never be reached under normal circumstances after your change.

For player case your assumptions are correct, and will work like you describe, but i dont want to break NPC to NPC case.

I.e. This likely needs a layer in addition on top of this.

Velite12:
then shouldn't the common factor of the criteria just be "the threat table is empty"?

Since threat assignment happens before combat assignment. So in other words you don't have a target in threat slot 1, even if it isn't yet your victim. Victim can be nullptr in this instance but you have top threat already even if it's 0.

killerwife:
Player has no threat list when not charmed. Otherwise for npc yes.

Velite12:
well in that case, we just allow player in this check? or just check if has threat list, since the assert under this assumes we need to actually have an AI which a non charmed player wouldn't.

killerwife:
Throw an empty threat list check in there and we go from there. I just realized player case gets consumed by first IF

src/game/Entities/Unit.cpp
```diff
@@ -8407,7 +8407,7 @@ void Unit::SetInCombatState(bool PvP, Unit* enemy)
                     if (PvP)
                         enemy->GetCombatManager().TriggerCombatTimer(controller);
                 }
-                else if (!controller->GetVictim())
+                else if (controller->CanHaveThreatList() && controller->getThreatManager() && !controller->getThreatManager().getCurrentVictim())
```
killerwife: This is what GetVictim resolves to, you want to check if its empty

killerwife (changes requested):

src/game/Entities/Unit.cpp
Velite12: Idk if this is actually it but I updated. Have to test but tomorrow. I feel like this is wrong because if the threat list isn't empty then there should already be an assumed victim in terms of the threat manager, not necessarily the unit. This explains why breakpointing I found that the mobs would actually attack you before they even made you their target.

Velite12: image
The original commit I made actually handles this edge case, you can see even though the pet is far away, when attacked the owner gets in combat too. Any other mutations to the check and this does not happen. The owner has a threat list, it is not empty at this time, but they do not have a victim according to the threat list.

image
Hunter brings friends too

As an aside, this mob is bugged as after 5 seconds, they will both evade, probably because waypoints aren't paused is my guess.

So in other words this is good to go already.

killerwife: The code inside AttackStart puts a chase movegen on top of waypoint movement. Is it getting called for that guy too? I will look at this tomorrow morning.

Velite12: Actually it's two things, one is that and the other is when you use distract, the pet goes one way and the hunter goes the other way, even if you distracted both at the same time. This made it easier to prove the threat transfer when the pet enters combat (as it's practically immediate) because I wanted to ensure that the owner is attacking because of this and not for pull assistance, but I am pretty sure that's not right.

They aren't linked innately since when testing I did other checks and they didn't pull together even if this code would fail to bring the owner in.

image
image
Then at some point the wolf defaults to moving around a specific point in random motion like most of these other mobs, instead of continuing the waypoint.

Velite12:
image

So this is not solved yet, this is currently only solved for the situation with 1 mob. When there is more than 1 mob the issue arises again.

image

In below image I left GM and went into combat and the mob didn't attack me even though both players have 0 threat. But in the above picture, as soon as I left GM, both attacked me even though same situation.

image

CallStack of when this part of the bug happens. Appears pull code is being called, even though pull already happened? And just to confirm, the enemy being passed is also me, so it is coming from one of these two mobs. This is after the normal code happens (where if nearby they place you in combat). So I'm already in combat, already in the threat list, they are continuing to attack other player... then this happens.
image


Update: I think I solved it, let me know if this is good. CallAssistanceOnPull was not checking m_AlreadyCallAssistance was false, but CallAssistance was checking for m_AlreadyCallAssistance. Therefore they should not be calling for assistance on pull after the pull has already happened. They were calling for help as though adding that new target to their table had just started combat.

No, now this is breaking pull assist on actual pull... so somewhere around here definitely controls this bug.

UPDATE 2: I'm confident this is the solution for the multi-mob problem. We have to be able to check the m_AlreadyCallAssistance at the CreatureAI level before we try calling for help. So now, mobs will still call for help on pull (because it would be false), but after this they should not be calling for help on pull mid combat. And it works again if you reset the pull, they'll only call for pull assistance on pull now.

I think it makes sense to do this here because we keep doing NoCallAssistance  right before calling for help, but we never actually check if we've called for help before, so we just keep setting this value to true.

image

Low health fleeing (retreating) after pull still works, running through mobs and pulling them still works... so let me know if there's any edge cases but so far all looks good. Seconds later in above screenshot the other 2 mobs in the back came to pull to assist.

PR #867, opened 2026-09-01
Fix Shattered Halls missing/misplaced dialogue and positions

Velite12:
Fix Shattered Halls missing/misplaced dialogue and positions

Proof:
https://www.youtube.com/watch?v=PQIt63gvEOo
https://www.youtube.com/watch?v=7bPYV-iN8RI

For prisoner order hint (TBC Patch 2.4):
https://www.youtube.com/watch?v=S_ptBqUgQzE

This shows officer is not the first one dead. The locations of the prisoners are consistent with Classic TBC.

- Add missing intro execution dialogue for horde, and move the alliance one out of the officer death, placing both right after you initially get first debuff as shown in videos
- Fix Prisoner kill order. From video and broadcast entry order, it can thus be surmised: dwarf/tauren, officer, then human/orc, currently we had officer dying first but this is not so according to real TBC. In addition, add the correct texts for each of them when they die.
- Switch the locations of the front two prisoners so it matches TBC/Classic TBC
- Rename enums for NPC prisoners to be more accurate to who they are
- Rename and add missing broadcast text entries

TODO: There is a dialogue entry that is currently and will still be unused, but it will be added to the enums. I don't know and can't find where this actually gets used, so maybe someone else can figure out when, but I would imagine it would be after the intro but but before the first prisoner is killed.


killerwife:
@miraco review this pls since u touched SHH last. I got no code style comments

PR #870, opened 2026-09-15
[Core/Movement] Fix use-after-free crash during teleports

joeyc83:
This PR resolves a use-after-free crash in MovementHandler.cpp that can occur when a player is teleported while the movement generator is processing.


src/game/MotionGenerators/MovementHandler.cpp
```diff
@@ -165,97 +165,103 @@ void WorldSession::HandleMoveWorldportAckOpcode()
     }
     else
         GetPlayer()->Relocate(loc.coord_x, loc.coord_y, loc.coord_z, loc.orientation);
-    auto lambda = [this, loc, old_loc, mEntry, mInstance](Map* map)
+    auto lambda = [playerGuid = GetPlayer()->GetObjectGuid(), loc, old_loc, mEntry, mInstance](Map* map)
     {
-        if (GenericTransport* transport = map->GetTransport(GetPlayer()->m_teleportTransport))
+        Player* player = ObjectAccessor::FindPlayer(playerGuid, false);
```
killerwife: Use Map GetPlayer. This will matter in the future.

